PR #1, opened 2019-04-26
feat: adds file updating support for simple node repos

bcoe:
This PR implements the first step in a release process for simple Node.js based repos, updating:

- [x] CHANGELOG.md.
- [x] package.json.
- [ ] samples/package.json (if it exists).

Using tooling built around conventionalcommits.org.

Here's an example, you can run it against any repo you have write access to:

https://github.com/bcoe/examples-conventional-commits/pull/22/files

## Next Steps

* add support for tagging, so that we can tag with `autorelease: pending`, and let our other release tools take over.

CC: @ofrobots


package.json
```diff
@@ -5,11 +5,13 @@
   "main": "./build/src/index.js",
   "bin": "./build/src/bin/release-please.js",
   "scripts": {
-    "test": "c8 --reporter=html --reporter=text mocha --timeout=5000 build/test",
-    "test:integration": "mocha --timeout=5000 build/test-integration",
-    "test:all": "c8 --reporter=html --reporter=text mocha --teimout=5000 build/test-integration build/test",
+    "test": "cross-env ENVIRONMENT=test c8 --reporter=html --reporter=text mocha --recursive --timeout=5000 build/test",
+    "test:integration": "cross-env ENV=test mocha --timeout=5000 build/test-integration",
```
JustinBeckwith: So usually, we use `system-test` for the end to end.  Just for the sake of consistency, it would be nice to use the same here.

src/bin/release-please.ts
```diff
@@ -24,14 +24,17 @@ const yargs = require('yargs');
 
 yargs
```
JustinBeckwith: not sure about this `yargs` lib, is that something we should trust?  🤣

src/checkpoint.ts
```diff
@@ -0,0 +1,29 @@
...
+import * as figures from 'figures';
+
+export enum CheckpointType {
+  Success = 'success',
+  Failure = 'failure'
+}
+
+export function checkpoint(msg: string, type: CheckpointType) {
+  const prefix = type === CheckpointType.Success ?
+      chalk.green(`${figures.tick}`) :
```
JustinBeckwith: The string interpolation just feels like extra steps?

src/checkpoint.ts
```diff
@@ -0,0 +1,29 @@
...
+      chalk.red(`${figures.cross}`);
+  if (!process.env.ENVIRONMENT) console.info(`${prefix} ${msg}`);
```
JustinBeckwith: This is a preference, and we should have it added to the style guide - mind using curly braces and a few extra lines for readability here?

src/mint-release.ts
```diff
@@ -75,7 +75,7 @@ export class MintRelease {
     const updates: Update[] = [];
 
     updates.push(new Changelog({
-      path: 'CHANGELOG.md',
+      path: 'CHANGELerG.md',
```
JustinBeckwith: heh, wat?

bcoe: whoops, was testing files that don't yet exist, reverted shortly.

src/bin/release-please.ts
bcoe: ¯\_(ツ)_/¯ what's the worst thing that can happen?

PR #6, opened 2019-05-01
fix: support H2 and H3 CHANGELOG headings

bcoe:
We need to be able to support both `##` and `###` markdown in CHANGELOGs, we should also support our old-style CHANGELOG template which doesn't link the version # (so that the transition process is easier).


bcoe:
blocked by #7

PR #22, opened 2019-05-07
test: output from html report was overwriting lcov report

bcoe:


bcoe:
success, we have reports.

Issue #140, opened 2019-05-20
release notes improperly extracted from CHANGELOG.md


bcoe:
In a recent release of c8, the release notes were not extracted correctly from CHANGELOG.md, see:

<img width="748" alt="Screen Shot 2019-05-19 at 11 13 47 PM" src="https://user-images.githubusercontent.com/194609/57999862-c8d5ac00-7a8b-11e9-8295-b1c0409525c8.png">

PR #142, opened 2019-05-20
fix: extend release notes regex to support patches

bcoe:
fixes: #140


bcoe:
corresponding fix here: https://github.com/googleapis/releasetool/pull/193

PR #164, opened 2019-05-26
feat: adding support for yoshi-PHP mono-repo to release-pr

### LINKED ISSUE #154: make test coverage less abysmal
the test coverage on release-please is quite low, I'd like to put some work into increasing it (now that we're starting to roll out to a wider audience).

bcoe:
Digging into the GitHub V3 API, I could not find an elegant way to fetch commits in conjunction with the file paths they modified ... I don't think there's a way to do it without a ton of requests, and potentially know more context about the repo being interacted with than I would like to, e.g., knowing which folders we need to fetch commits for.

I've opted to instead use the GitHub V4 GraphQL API, this allows us to fetch commits and file paths in one (granted hideous) query.

## Known Limitations in Implementation

If commits are pushed directly to the default branch (and therefore have no associated PR), we do not currently collect the files changed.

## To Do

- [x] refactor commits into object, the `\n` message + sha has reached the end of its shelf life.
- [x] write helper that transforms graphql output into commit objects.
- [x] write helper that partitions list of commits by folder (we only care about the commits that touch a given folder in the mono repo when determining changes to a given library).
- [x] pull everything together.
- [x] switch to using graphql for fetching all commits.

BREAKING CHANGE: dropped support for candidate issues as part of this refactor (as we've moved towards the workflow of keeping an open PR).

fixes #154


jdpedrie:
I've used the [Compare API](https://developer.github.com/v3/repos/commits/#compare-two-commits) in the past. It includes a list of modified files in each commit. Is there a drawback which makes it unsuitable for this case?

PR #192, opened 2019-06-10
feat: avoid updating release PR if no additional changes

### LINKED ISSUE #158: Running at night
There are always gonna be a few releases we sit on for a bit.  It seems like something is happening in a nightly cron somehow that's triggering the tool to get run:
image

It would be awesome if it only ran after things landed into master :)

bcoe:
currently, whenever we land a PR we update a release PR, even if there have been no additional changes.

fixes #158


bcoe:
@ofrobots @JustinBeckwith I think this will address the annoyance of PRs being updated overnight and whenever another patch is landed.

PR #203, opened 2019-07-17
feat: adding support for google-auth-library-java

bcoe:
Adds support for the files that should be updated during a release of `google-auth-library-java` (_this represents a good prototype of what things will look like once the repo split happens for Java_).

New updaters added:

* `versions.txt`: for normal releases updates all entries to `version:version`, for snapshot releases updates entries to `previous-release:snapshot-release`.
* `pom.xml`: update to `<version>newVersion</version>`.
* `README.md`: update examples of Java, Scala, and Groovy imports.

@chingor13 please take a look and let me know if the logic looks right, tomorrow I'll actually integrate it with the CLI and we can play around.

### DIFF
--- a/src/updaters/java-auth-readme.ts
+++ b/src/updaters/java-auth-readme.ts
@@ -1,0 +1,51 @@
+/**
+ * Copyright 2019 Google LLC. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { checkpoint, CheckpointType } from '../util/checkpoint';
+import { Update, UpdateOptions } from './update';
+import { GitHubFileContents } from '../github';
+
+export class JavaAuthReadme implements Update {
+  path: string;
+  changelogEntry: string;
+  version: string;
+  versions?: { [key: string]: string };
+  packageName: string;
+  create: boolean;
+  contents?: GitHubFileContents;
+
+  constructor(options: UpdateOptions) {
+    this.create = false;
+    this.path = options.path;
+    this.changelogEntry = options.changelogEntry;
+    this.version = options.version;
+    this.packageName = options.packageName;
+  }
+  updateContent(content: string): string {
+    content = content.replace(
+      /<version>.*<\/version>/,
+      `<version>${this.version}</version>`
+    );
+    content = content.replace(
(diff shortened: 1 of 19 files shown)


src/updaters/java-auth-readme.ts
```diff
@@ -0,0 +1,51 @@
...
+    this.version = options.version;
+    this.packageName = options.packageName;
+  }
+  updateContent(content: string): string {
+    content = content.replace(
+      /<version>.*<\/version>/,
+      `<version>${this.version}</version>`
+    );
+    content = content.replace(
+      /'com.google.auth:google-auth-library-oauth2-http:[^']+'/,
```
chingor13: If you're looking to generalize this behavior, we do tag lines to search for versions:
* `{x-version-update:google-auth-library-bom:current}`: replace any Java semver pattern on that same line with the `current-version` of `google-auth-library-bom` found in the `versions.txt` file.
* Between `{x-version-update-start:google-auth-library-oauth2-http:released}` and `{x-version-update-end}`: replace any Java semver pattern with the `released-version` of `google-auth-library-oauth2-http` found in the `versions.txt` file.

In theory, with these two rules, you could run it on any file, rather than having per-file rules. In gradle setups, there are different build files that specify versions.

src/updaters/java-auth-versions.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  constructor(options: UpdateOptions) {
+    this.create = false;
+    this.path = options.path;
+    this.changelogEntry = options.changelogEntry;
+    this.version = options.version;
+    this.packageName = options.packageName;
+  }
+  updateContent(content: string): string {
+    if (this.version.includes('SNAPSHOT')) {
```
chingor13: Another thing to note is that a Java version can have a "quality" attribute as well. `0.1.1-alpha-SNAPSHOT` is a valid Java version string. In that example "alpha" could be any arbitrary alpha-numeric value.

src/updaters/java-auth-readme.ts
bcoe: 👍 might hold off and make the logic generic as we do the repo-split. I think we'll potentially  want to pass `groupId` into the updated? so that it can use this along with `packageName` to create a more generic replacement?

Are we going to have similar lines in the README for each split repo?

chingor13: Yes, the plan is to keep the README for each google-cloud-[service] library as the main README on the split repo.

test/updaters/fixtures/pom.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.google.auth</groupId>
+    <artifactId>google-auth-library-parent</artifactId>
+    <version>0.16.2</version><!-- {x-version-update:google-auth-library-parent:current} -->
```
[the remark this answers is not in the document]
bcoe: added a few more test cases in the fixture for `-alpha-SNAPSHOT`, `-SNAPSHOT`, etc.

PR #217, opened 2019-08-01
fix: don't print full error in --debug mode

bcoe:
oktokit includes auth headers and URL in the error object, this makes printing the error object a security concern.

### DIFF
--- a/src/updaters/samples-package-json.ts
+++ b/src/updaters/samples-package-json.ts
@@ -45,7 +45,9 @@ export class SamplesPackageJson implements Update {
       } to ^${this.version}`,
       CheckpointType.Success
     );
-    parsed.dependencies[this.packageName] = `^${this.version}`;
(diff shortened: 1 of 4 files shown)


src/updaters/samples-package-json.ts
```diff
@@ -45,7 +45,9 @@ export class SamplesPackageJson implements Update {
       } to ^${this.version}`,
       CheckpointType.Success
     );
-    parsed.dependencies[this.packageName] = `^${this.version}`;
+    if (parsed.dependencies[this.packageName]) {
```
chingor13: What's this change doing? It looks like it was previously setting/clobbering whatever was in `parsed.dependencies[this.packageName]`. Now it only clobbers what was there if it was set.

bcoe: was branched off of another PR that just landed; should no longer see this.

PR #222, opened 2019-08-08
feat: add support for yoshi-ruby mono repo

bcoe:
adding support for our yoshi-style ruby mono repos.

_I've used this as an excuse to refactor the releasers for each language into their own file, as the one file was getting quite large._


src/releasers/ruby-yoshi.ts
```diff
@@ -0,0 +1,137 @@
...
+  protected async _run() {
+    if (!this.lastPackageVersion) {
+      throw Error(
+        'last library version must be provided for ruby-yoshi release type'
+      );
+    }
+    const lastReleaseSha = await this.gh.getTagSha(
+      `${this.packageName}/v${this.lastPackageVersion}`
+    );
+    const commits: Commit[] = await this.commits(lastReleaseSha);
```
bcoe: how do we handle the very first release of a library?

TheRoyalTnetennba: If we go with the workflow we talked about earlier, unreleased packages will call release-please with `lastPackageVersion == '0.1.0'`

PR #229, opened 2019-08-16
feat: allow octokit to be dependency injected

bcoe:


src/github.ts
```diff
@@ -85,24 +86,65 @@ export class GitHub {
...
+      this.request = request.defaults(defaults);
+      this.graphql = graphql;
+    } else {
+      // for the benefit of probot applications, we allow a configured instance
+      // of octokit to be passed in as a parameter.
+      this.probotMode = true;
+      this.octokit = options.octokitInstance;
+      this.request = options.octokitInstance.request;
+      // @ts-ignore // probot adds a graphql parameter to its instance.
+      this.graphql = options.octokitInstance.graphql;
```
chingor13: This feels dirty - can we make `this.graphql` a interface defined by release-please and probot can wrap it's graphql implementation to provide it.

src/github.ts
```diff
@@ -79,30 +86,76 @@ interface GitHubPR {
...
+      this.graphql = options.octokitAPIs.graphql;
+    }
+  }
+
+  private async graphqlRequest(_opts: {
+    [key: string]: string | number | null | undefined;
+  }) {
+    let opts = Object.assign({}, _opts);
+    if (this.probotMode === undefined) {
+      opts = Object.assign(opts, {
```
chingor13: Can we wrap this in the default function assigned in the constructor and avoid needing probotMode?

src/github.ts
```diff
@@ -499,16 +531,15 @@ export class GitHub {
...
-          Authorization: `${this.proxyKey ? '' : 'token '}${this.token}`,
-        },
-      })) {
+      for await (const response of this.octokit.paginate.iterator(
+        this.decoratePaginateOpts({
+          method: 'GET',
+          url: `/repos/${this.owner}/${this.repo}/issues?labels=${labels.join(
+            ','
+          )}${this.proxyKey ? `&key=${this.proxyKey}` : ''}`,
+          per_pag: 100,
```
chingor13: Is this correct?

src/github.ts
```diff
@@ -555,12 +586,14 @@ export class GitHub {
             key: this.proxyKey,
           }
         );
+        console.info('we get this far, i.e., created a branch');
```
chingor13: Keep the console.log?

src/github.ts
```diff
@@ -555,12 +586,14 @@ export class GitHub {
...
       } catch (err) {
         if (err.status === 404) {
           // the most likely cause of a 404 during this step is actually
           // that the user does not have access to the repo:
           throw new AuthError();
         } else {
+          console.info('>>>>', err.stack);
```
chingor13: Keep the console.log?

src/github.ts
```diff
@@ -79,30 +86,76 @@ interface GitHubPR {
...
   octokit: Octokit;
+  request: Function;
+  graphql: Function;
   token: string | undefined;
   owner: string;
   repo: string;
   apiUrl: string;
   proxyKey?: string;
-  request: Function;
+  probotMode?: boolean;
```
chingor13: If we keep this instance variable, it looks like it's always defined.

src/github.ts
```diff
@@ -79,30 +86,76 @@ interface GitHubPR {
 
 export class GitHub {
   octokit: Octokit;
+  request: Function;
```
chingor13: Question: what's the default visibility for these variables?

src/github.ts
bcoe: Because of the `?` it will be defined, but `undefined` if not provided ... didn't love this, the alternative would be perhaps dynamically defining `graphqlRequest` and `decoratePaginateOpts`, depending on whether `octokitAPis` have been provided.

PR #234, opened 2019-08-21
fix: docs should not count as feature

bcoe:
* if only fixes have landed, docs will show up in fix.
* if features have landed, docs will show up in features.


bcoe:
see: https://github.com/googleapis/release-please/issues/235 tracking issue for adding tests.

PR #242, opened 2019-08-22
fix: address a couple more bugs with Ruby libraries

bcoe:
* if one library contained the name of another library we would potentially close the PR for the related library (fixed this by making sure there are spaces surrounding library names when looking for PRs to close).
* grafeas had a slightly different naming convention `lib/grafeas` not `lib/grafeas-client`, we should be mindful of any other libraries that are edge cases.


bcoe:
we might actually want to fix grafeas client's naming in the ruby mono repo.

PR #243, opened 2019-08-22
fix: java snapshot releases skip autorelease label

chingor13:


src/releasers/java-yoshi.ts
```diff
@@ -45,6 +45,13 @@ const CHANGELOG_SECTIONS = [
 ];
 
 export class JavaYoshi extends ReleasePR {
+  constructor(options: ReleasePROptions) {
+    if (!options.label && options.snapshot) {
+      options.label = 'type: process';
```
bcoe: this codebase is quickly becoming object oriented :smile: I'm glad I refactored these classes out.

PR #247, opened 2019-08-23
feat: add ReleasePRFactory

chingor13:


src/release-pr.ts
```diff
@@ -64,7 +67,6 @@ export class ReleasePR {
   token: string | undefined;
   packageName: string;
   releaseAs?: string;
-  releaseType: ReleaseType;
```
bcoe: I like that we get rid of this.

PR #260, opened 2019-09-10
fix: replace all version update tags

### LINKED ISSUE #261: Java: doesn't replace all versions in the pom files
Example: https://github.com/googleapis/java-core/pull/18/commits/8b87caaa8db047102959d542f6eaa69e7b174acd

This should update the version of this artifact and the version of the parent (both have the region tags set).

chingor13:
Fixes #261


chingor13:
@bcoe Added a test

PR #266, opened 2019-09-16
feat: add initial support for Python libraries

bcoe:
adds initial support for a Python repo.

### DIFF
--- a/src/updaters/python/setup-py.ts
+++ b/src/updaters/python/setup-py.ts
@@ -1,0 +1,42 @@
+/**
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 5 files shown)


src/updaters/python/setup-py.ts
```diff
@@ -0,0 +1,42 @@
...
+  constructor(options: UpdateOptions) {
+    this.create = false;
+    this.path = options.path;
+    this.changelogEntry = options.changelogEntry;
+    this.version = options.version;
+    this.packageName = options.packageName;
+  }
+  updateContent(content: string): string {
+    return content.replace(
+      /version ?= ?"[0-9]+\.[0-9]+\.[0-9](-\w+)?"/,
```
busunkim96: Could this regex also allow for single quotes around the version?

bcoe: see no reason why not, although it would switch them to `"` when generating the new PR, is that a problem?

busunkim96: It's fine if they get switched for our repositories. Most are using a code-formatter that will always use double-quotes, but we might have some stragglers.

PR #290, opened 2019-10-17
feat: populate packageName from package.json

### LINKED ISSUE #289: fetch packageName parameter from repository itself
could we fetch the packageName from the repository itself, we already pull in `package.json` to update it... perhaps we could load the `package.json` earlier in the process, and use this value to override the `packageName` set in configuration.

bcoe:
fixes #289.

### DIFF
--- a/src/releasers/node.ts
+++ b/src/releasers/node.ts
@@ -72,7 +72,7 @@ export class Node extends ReleasePR {
       `package.json`
     );
     const pkg = JSON.parse(contents.parsedContent);
-    if (pkg.name) this.packageName = pkg.name
+    if (pkg.name) this.packageName = pkg.name;
 
     updates.push(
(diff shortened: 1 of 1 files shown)


src/releasers/node.ts
```diff
@@ -81,6 +89,7 @@ export class Node extends ReleasePR {
         changelogEntry,
         version: candidate.version,
         packageName: this.packageName,
+        contents,
```
joeldodge79, line 92: @bcoe I just discovered a bug here while switching from `sandbox.stub` to using a mock: it caught an unexpected extra call to get the contents for package.json. The reason is that the `PackageJson` constructor was never updated to set `this.contents` so `GitHub.getChangeSet()` has to reach back out to to get it. PR coming up.

PR #329, opened 2019-12-21
feat!: add autorelease: tagged label when creating GitHub release

bcoe:
BREAKING CHANGE: made `addLabels` API consistent with `removeLabels`.

The automation tool we use in conjunction with release-please, `autorelease`, adds the tag `autorelease: tagged`, when a GitHub release has been created, but the package hasn't yet been published to a package registry.

release-please now does the same; this will allow Node.js to continue using the same release monitoring helpers (which raise an alarm if a release failed) as we start moving towards a probot application for publishing to npm.


src/github.ts
```diff
@@ -514,14 +514,14 @@ export class GitHub {
     return tags;
   }
 
-  async addLabels(pr: number, labels: string[]) {
+  async addLabels(labels: string[], pr: number) {
```
chingor13: Is this technically a breaking change? It seems like methods default to public in classes.

bcoe: valid point, I should have made methods private at the outset.

bcoe: I went ahead and made this breaking.

PR #358, opened 2020-02-12
fix: process label was causing PRs to be closed

bcoe:
let's just close open issues with `autorelease: pending` by default, `process` is used too commonly.


chingor13:
Should release-please use it's own reserved labels to identify it's open PRs?

bcoe:
@chingor13 this is how I was picturing `autorelease: pending`, which seems unlikely for someone to add if it's not an autorelease PR; at some point we just made the logic less restrictive and caught the `process` labels too.

PR #352, opened 2020-02-03
feat: include docs and chores in PHP releases and changelogs

jdpedrie:
Fixes #348.

Hello friends! This change aims to cause release-please to include `chore` and `docs` changes in our changelog, and therefore to cause those types of changes to be released, even if no other changes have been made to a given component.

I followed the relevant portions of #238, where @bcoe added a similar feature to the ruby releaser. I looked through the test suites and did not find anything which seemed to relate closely to this part of the application, but if I missed something, please let me know and I'll work to fix it!

Thanks for considering the change. :)

PR #359, opened 2020-02-13
fix: support changelog types for PHP components

jdpedrie:
Follows up on #352, adds custom changelog sections to component changeset.


bcoe:
@jdpedrie thanks for the patch :ok_hand:

PR #374, opened 2020-03-19
feat(python): Add support to change version in setup.cfg

crwilcox:
Python supports using a config file for settings as well as in setup.py.

### DIFF
--- a/test/updaters/fixtures/setup.cfg
+++ b/test/updaters/fixtures/setup.cfg
@@ -1,0 +1,43 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


test/updaters/fixtures/setup.cfg
```diff
@@ -0,0 +1,43 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+[metadata]
+name = google-crc32c
+description = A python wrapper of the C library 'Google CRC32C'
```
busunkim96: I don't see a version in this setup.cfg?

crwilcox: I suppose that explains why it passed.

PR #380, opened 2020-03-26
feat(deps): update dependencies & fix gts lint

xiaozhenliu-gg5:
1. upgrade Typescript to 3.8.x
2. upgrade gts and fix lint.

PR #381, opened 2020-03-27
refactor!: upgrade octokit/gts/typescript

bcoe:
@xiaozhenliu-gg5 I ended up pulling over your work from #380 into this PR, as I had to do a significant amount of refactoring of types to upgrade @octokit (I'd been putting off this update for weeks because I knew it would take so much playing around).

Sorry to have wasted a little bit of your time on #380.


xiaozhenliu-gg5:
It's ok, thanks for doing this! @bcoe @octokit bites me for linter a lot so I have to disable some no-var-requires check for it. You are doing this in a cleaner way! 
Next time I will let you know which API I'm working on so that our work won't be overlapped😀

package.json
```diff
@@ -35,23 +35,24 @@
...
+    "@octokit/types": "^2.5.1",
     "@types/chai": "^4.1.7",
     "@types/mocha": "^7.0.0",
     "@types/node": "^11.13.6",
     "@types/semver": "^7.0.0",
     "c8": "^7.0.0",
     "chai": "^4.2.0",
     "cross-env": "^7.0.0",
-    "gts": "^1.0.0",
+    "gts": "^2.0.0-alpha.8",
```
summer-ji-eng: I saw some of package use `next` tag here. Dose it matter here?

src/conventional-commits.ts
```diff
@@ -11,14 +11,15 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+/* eslint-disable @typescript-eslint/no-var-requires */
```
summer-ji-eng: `eslint-disable` affect entire file, is this what we want?

package.json
bcoe: I prefer being explicit with the dependency, as soon as we do a few GAPIC library updates, I think we can go ahead and migrate gts to 2.x 😄

PR #391, opened 2020-04-01
fix: delay snapshots temporarily

chingor13:

### DIFF
--- a/src/release-pr.ts
+++ b/src/release-pr.ts
@@ -103,7 +103,18 @@ export class ReleasePR {
     this.gh = this.gitHubInstance(options.octokitAPIs);
   }
 
+  async delay({ms = 3000}) {
+    new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


src/release-pr.ts
```diff
@@ -103,7 +103,21 @@ export class ReleasePR {
...
+      }, ms);
+    });
+  }
+
   async run() {
+    // TODO: this temporarily resolves a race condition between creating a release
+    // and updating tags on the release PR. This should be replaced by a queuing
+    // mechanism to delay/retry this request.
+    if (this.snapshot) {
+      await this.delay({ms: 10000});
```
bcoe, line 119: because this behavior is a bit weird, I would be tempted to add a `console.log` message before and after the `this.delay` so that we can catch any issues in the GCF logs.

chingor13: Done

PR #407, opened 2020-04-10
feat: add Java BOM versioning strategy

chingor13:
If a dependency update bumps a minor version, bump the minor, otherwise consider it a patch version bump.

* Extracts `BumpType` and `Version` into their own modules and test them separately
* Branch `JavaBom` from `JavaYoshi` and change how it determines the bump type from the dependency updates


src/releasers/java/bump_type.ts
```diff
@@ -0,0 +1,38 @@
...
+import * as semver from 'semver';
+
+export type BumpType = 'major' | 'minor' | 'patch' | 'snapshot';
+
+export function maxBumpType(bumpTypes: BumpType[]): BumpType {
+  if (bumpTypes.some(bumpType => bumpType === 'major')) {
+    return 'major';
+  }
+  if (bumpTypes.some(bumpType => bumpType === 'minor')) {
+    return 'minor';
```
bcoe, line 24: I like how you've decomposed this into a few helpers.

PR #415, opened 2020-04-24
chore: update npm scripts and synth.py

alexander-fenster:
Update npm scripts: add clean, prelint, prefix; make sure that lint and fix are set properly. Use post-process feature of synthtool.


bcoe:
release-please only uses synthtool for populating common templates, so I think we can get away with closing this.

PR #427, opened 2020-05-05
fix: don't include pre-release in latestTag calculation

bcoe:
pre-releases were being used in the logic for calculating the latest tag, this meant that as soon as we released a pre-release, it broke the CHANGELOG generation on the next run.

### DIFF
--- a/__snapshots__/java-bom.js
+++ b/__snapshots__/java-bom.js
@@ -836,8 +836,8 @@ This PR was generated with [Release Please](https://github.com/googleapis/releas
 `
 
 exports['labels-bom'] = {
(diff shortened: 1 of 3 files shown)


__snapshots__/java-bom.js
```diff
@@ -836,8 +836,8 @@ This PR was generated with [Release Please](https://github.com/googleapis/releas
 `
 
 exports['labels-bom'] = {
-  "labels": [
```
JustinBeckwith, line 839: Possible you need to ignore `__snapshots__` in your `.prettierignore`

Issue #428, opened 2020-05-07
Add support for releasing from non-master branches for legacy support


feywind:
The intent here is to add the follow capabilities to release-please and any related bots (though this issue is only about release-please).

* Projects can have a branch besides `master` that is designated as a release branch.
* When PRs are merged into those branches, release-please will notice and make a release PR just like it does with `master`.
  * Those PRs will correctly find the next appropriate semver within the branch, not `master`.
* When a maintainer merges the release PR, the regular machinery will push it along its way to a release in the relevant package manager.

The primary alternatives are:

* Don't release from branches; this is becoming increasingly unpalatable because of discussions about the possibility of supporting non-latest versions, as well as the need to e.g. backport fixes to a stable branch until `master` is ready to release for whatever reason.
* Manually handle the release process; having done this once, it is incredibly gross and error-prone. It also requires access to some potentially sensitive docs.

Context:

The motivating factor here is that the nodejs-pubsub library (`@google-cloud/pubsub`) has merged updates to `master` that are pointed at a 2.0 release; however, that 2.0 release requires a breaking TypeScript update, as well as breaking Node 8 support. The Node 8 deprecation has been held up for a while now due to various factors, so I have had the need to port important changes into what is currently the `latest` branch, and keep it releasing.

Note:

I've already completed a fair amount of the work here, I'm just working on getting it cleaned up and tested.

feywind:
Linking this PR also. https://github.com/googleapis/release-please/pull/429

PR #429, opened 2020-05-07
feat: add support for releasing from non-master branches for legacy support

feywind:
This PR is the ongoing work for implementing this issue:

https://github.com/googleapis/release-please/issues/428


bcoe:
@feywind I think we might be able to pluck the tests from this work into the work here:

https://github.com/googleapis/release-please/pull/468

I've made it so we now detect whether an alternate default branch has been enabled on GitHub.

feywind:
`sup` ... heh. I am still intending to get this merged in, I just need to finish up tests.

feywind:
I was told today that several more teams are interested in this, so I will bump it in my queue accordingly.

src/github.ts
```diff
@@ -452,17 +453,17 @@ export class GitHub {
     preRelease = false
   ): Promise<GitHubTag | undefined> {
     let regexpString = prefix || '';
-    if (!preRelease) regexpString = `${regexpString}[^-]*`;
+    if (!preRelease) regexpString = `^${regexpString}[^-]*$`;
```
bcoe, line 456: curious if this addresses a bug you bumped into, or if you're just being prudent by terminating the regex.

src/github.ts
```diff
@@ -226,53 +227,59 @@ export class GitHub {
...
+    const baseBranch = await this.getDefaultBranch();
+
     // The GitHub v3 API does not offer an elegant way to fetch commits
     // in conjucntion with the path that they modify. We lean on the graphql
     // API for this one task, fetching commits in descending chronological
     // order along with the file paths attached to them.
     try {
       const response = await this.graphqlRequest({
-        query: `query commitsWithFiles($cursor: String, $owner: String!, $repo: String!, $perPage: Int, $maxFilesChanged: Int, $path: String) {
+        query: `query commitsWithFiles($cursor: String, $owner: String!, $repo: String!, $baseBranch: String!, $perPage: Int, $maxFilesChanged: Int, $path: String) {
```
bcoe, line 238: I'm excited that we can support any branch with the grahql command now 😄

src/github.ts
```diff
@@ -789,17 +809,18 @@ export class GitHub {
 
   async getFileContentsWithDataAPI(
     path: string,
-    defaultBranch: string | undefined
+    branch: string
   ): Promise<GitHubFileContents> {
+    const options: any = {
```
bcoe, line 814: if we could figure out a type here might be nice to get rid of the `any`, I wonder if we can pluck something out of `@octokit`.

src/graphql-to-commits.ts
```diff
@@ -76,7 +90,7 @@ export async function graphqlToCommits(
   response: CommitHistoryGraphQLResponse
 ): Promise<CommitsResponse> {
   const commitHistory: CommitHistory =
-    response.repository.defaultBranchRef.target.history;
+    response.repository.refs.edges[0].node.target.history;
```
bcoe, line 93: I always end up feeling like I'm holding graphql wrong, if we gradually move more things to the v5 API, we might want to pull in an ORM.

src/github.ts
```diff
@@ -86,6 +86,7 @@ export interface GitHubReleasePR {
   number: number;
   version: string;
   sha: string;
+  normalizedVersion: string;
```
bcoe, line 89: not a request to change the approach, but curious why you opted to track both `version` and `normalizedVersion`.

bcoe (commented):
This looks good to me, although it's making me wish we'd already switched to sinon for our mocks. Lots of annoying changes.

Thanks for doing this work 👏

bcoe (approved):

src/github.ts
```diff
@@ -312,31 +320,36 @@ export class GitHub {
     maxLabels = 16,
     retries = 0
   ): Promise<CommitsResponse> {
+    const baseBranch = await this.getDefaultBranch();
```
chingor13, line 323: Should this be a method argument where the caller can provide the base branch name?

src/github.ts
```diff
@@ -226,53 +227,59 @@ export class GitHub {
     maxFilesChanged = 64,
     retries = 0
   ): Promise<CommitsResponse> {
+    const baseBranch = await this.getDefaultBranch();
```
chingor13, line 230: Should this be a method argument where the caller can provide the base branch name?

chingor13 (commented):
It looks like `GitHub#defaultBranch` is a misnomer now and should be `baseBranch` which defaults to the detected default branch.

It might simplify the code if we provide the base branch as a parameter to the relative methods - the same GitHub instance could be used for multiple base (releasable) branches.

Conceptually, do we expect separate release-please invocations for each base branch -or- iterating over a list of base branches (via configuration) in the same invocation?

src/github.ts
feywind: Yeah, I saw this could go either way... what we had before was that methods had to supply options.defaultBranch each time, and then it got cached, so the future invocations ignored the parameter. Which seems a bit deceptive.

I'm fine with going the other way, too, just make it have to be passed each time and caching the result value by key or something.

src/github.ts
```diff
@@ -56,7 +56,7 @@ import {
 } from './graphql-to-commits';
 import {Update} from './updaters/update';
 
-const VERSION_FROM_BRANCH_RE = /^.*:[^-]+-(.*)$/;
+const VERSION_FROM_BRANCH_RE = /^.*:release-(.*)$/;
```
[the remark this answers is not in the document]
feywind: I haven't specifically studied the Ruby releaser about this, but yeah, that might be a problem, and I am also mystified how it worked before. :S

src/github.ts
feywind: I'm pretty sure this is in response to a unit test that broke with the non-terminated one.

src/github.ts
feywind: Heh, yeah, I noticed that and was like... how did this pass the linter before? But I think it was introduced originally in the large file PR. I can take a look if you like.

src/graphql-to-commits.ts
feywind: Yeah, the way we're doing this honestly feels a bit hacky to me, but I'd have to look at how others are using the API for an example of how to do it right. 😄 I was under the impression that there hasn't been a huge uptake on the graphql API for GitHub. We can talk about this thought later though.

src/github.ts
feywind: Ah, good question! It looks like a few pieces in `github-release.ts` are still expecting tag names. But it's not used anywhere else, so it makes sense to me to just shift the normalized one back to just `version` and then make the few places add a `v`.

src/github.ts
feywind: Oh, I missed your last comment - my intent is that this first pass would just allow someone to run it from the command line on a onesies-basis, but I'm intending to follow it up with one that will read a list of branches from the `.release-please.yaml` file in a repo. So I'm good with assuming that's where things need to go, one way or another.

src/github.ts
feywind: Looks like a few tests were failing:
```
  1) GitHubRelease
       createRelease
         creates and labels release on GitHub:
     Uncaught Error: Different value of snapshot "GitHubRelease createRelease creates and labels release on GitHub"
{
  name: "foo v1.0.3" => "foo 1.0.3"
  tag_name: "v1.0.3" => "1.0.3"
}
```
But I think it should be pretty easy to just fix up the places in github-release.ts that need this. Then I can get rid of the normalized member and make `version` the normalized one.

src/github.ts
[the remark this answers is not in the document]
feywind: Oh, actually, I see it now. The old version just eats anything that's not a dash up to the first dash. The new one requires that the eaten text is "release". This change was just to ensure that it's actually a release PR that's being looked at. So I think we're okay here.

src/github.ts
feywind: It's good I took a look at that, there was a hidden chance for a deref'd `undefined`.

chingor13 (approved):

PR #439, opened 2020-05-19
refactor!: dynamically load releasers in ./releasers directory

bcoe:
This refactor pulls some additional logic to the `ReleasePR` class (it now has `releaserName` which is used rather than an `enum` for looking up a specific release type, and `lookupPackageName` method, which was a feature being done in an ad hoc way for only Node.js).

BREAKING CHANGE: this removers the `ReleaseType` `enum`, in favor of a dynamically generated list of loaders.


src/bin/release-please.ts
```diff
@@ -89,6 +89,21 @@ const argv = yargs
         .option('label', {
           default: 'autorelease: pending',
           describe: 'label to remove from release PR',
+        })
+        .option('release-type', {
+          describe: 'what type of repo is a release being created for?',
+          choices: [
+            'node',
```
alexander-fenster, line 96: I'm guessing this list should now include `simple`?

src/release-pr-factory.ts
```diff
@@ -34,26 +34,29 @@ export class ReleasePRFactory {
...
+        return TerraformModule;
       case ReleaseType.Python:
-        return new Python(releaseOptions);
+        return Python;
       case ReleaseType.Ruby:
-        return new Ruby(releaseOptions as RubyReleasePROptions);
+        return Ruby;
       case ReleaseType.RubyYoshi:
-        return new RubyYoshi(releaseOptions);
+        return RubyYoshi;
```
alexander-fenster, line 59: again, `ReleaseType.Simple` is missing here

src/bin/release-please.ts
bcoe: I think we missed this in the earlier PR yes, I'll correct.

src/releasers/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    !file.name.match(/index\.js/)
+  ) {
+    const obj = require(`./${file.name}`) as {[key: string]: typeof ReleasePR};
+    releasers.push(obj[Object.keys(obj)[0]]);
+  }
+}
+
+export function getReleaserNames(): string[] {
+  return releasers.map(releaser => {
+    console.info(releaser.releaserName);
```
alexander-fenster, line 38: do we still need this logging?

src/releasers/index.ts
```diff
@@ -0,0 +1,43 @@
...
+const root = dirname(require.resolve('./'));
+for (const file of readdirSync(root, {withFileTypes: true})) {
+  if (
+    file.isFile() &&
+    !file.name.match(/.*\.ts.*/) &&
+    !file.name.match(/.*\.map$/) &&
+    !file.name.match(/index\.js/)
+  ) {
+    const obj = require(`./${file.name}`) as {[key: string]: typeof ReleasePR};
+    releasers.push(obj[Object.keys(obj)[0]]);
```
alexander-fenster, line 32: nit, I would probably add them to an object instead, so that we could get one by accessing it by name directly and without a `for` loop. But I'm OK either way.

PR #443, opened 2020-05-20
feat: a flag to skip CI

alexander-fenster:
Instead of making assumptions such as "let's run / let's not run CI for the first / the last / whatever file", let's just make it configurable. I'm only using this flag in the simple releaser but any other releaser can be upgraded if needed. The default behavior is `skipCi: true` so nothing is changed for other releasers.

### DIFF
--- a/__snapshots__/java-yoshi.js
+++ b/__snapshots__/java-yoshi.js
@@ -973,3 +973,23 @@ public final class GoogleUtils {
   private GoogleUtils() {}
 }
 `
+
+exports['CHANGELOG-message'] = `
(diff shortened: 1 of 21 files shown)


alexander-fenster:
@bcoe OK, so now:
- if `skipCi` is unset for all updates (default behavior), we disable CI (set `skipCI = true`) for all files except the last one - this case covers all the existing releasers used by Yoshi;
- if `skipCi` is set at least for one update (to `true` or `false`), we think that this releaser cares about it, and skip CI for all updates where it is unset.

PR #448, opened 2020-05-21
docs: explain how to automate publication

bcoe:
Demonstrate how to compose GitHub actions such that you can automate publications with release-please.


README.md
```diff
@@ -88,6 +88,48 @@ The easiest way to run release please is as a GitHub action:
...
+on:
+  push:
+    branches:
+      - master
+name: release-please
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: bcoe/release-please-action@v1.2.1
```
JustinBeckwith, line 106: Can we use `GoogleCloudPlatform` now?

README.md
```diff
@@ -88,6 +88,48 @@ The easiest way to run release please is as a GitHub action:
...
+        id: release
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          release-type: node
+          package-name: test-release-please
```
JustinBeckwith, line 111: Generally, can't you get this out of `package.json`?

README.md
```diff
@@ -88,6 +88,48 @@ The easiest way to run release please is as a GitHub action:
...
+        # a new release is created:
+        if: ${{ steps.release.outputs.release_created }}
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+          registry-url: 'https://registry.npmjs.org'
+        if: ${{ steps.release.outputs.release_created }}
+      - run: npm ci
+        if: ${{ steps.release.outputs.release_created }}
+      - run: npm publish
```
JustinBeckwith, line 124: I'll admit to not understanding everything that's happening in this script.  I can say - I feel like this was far more straightforward _without_ a GitHub action with `semantic-release`:
https://github.com/JustinBeckwith/linkinator/blob/master/.github/workflows/ci.yaml#L53

Why is this so much harder?

README.md
bcoe: I was figuring I'd land everything against `bcoe/` then update to GoogleCloudPlatform once I've done a first release from that location.

README.md
bcoe: I find that code you linked to be pretty much identical, honestly maybe a little more confusing to me. The publish step is this:

```
      - run: npm publish
        env:
          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
        if: ${{ steps.release.outputs.release_created }}
```

☝️  all it's doing is saying that it should only publish if a release was created in the prior steps.

README.md
bcoe: Yeah, we do now detect this from `package.json` in the case of Node.js repos, so we could drop this line.

PR #452, opened 2020-06-06
fix: merge commits now handled

bcoe:
For merge commits, prEdge.node.mergeCommit.oid references the SHA of the
commit at the top of the list of commits, vs., its own SHA. We track the
SHAs observed, and if the commit references a SHA from earlier in the list
of commitHistory.edges being processed, we accept it as a valid commit.

### DIFF
--- a/test/fixtures/commits-with-labels.json
+++ b/test/fixtures/commits-with-labels.json
@@ -87,7 +87,7 @@
                     "edges": [
                       {
                         "node": {
-                          "mergeCommit": {"oid": "35abf13fa8acb3988aa086f3eb23f5ce1483cc5d"},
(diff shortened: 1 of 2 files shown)


test/fixtures/commits-with-labels.json
```diff
@@ -87,7 +87,7 @@
                     "edges": [
                       {
                         "node": {
-                          "mergeCommit": {"oid": "35abf13fa8acb3988aa086f3eb23f5ce1483cc5d"},
+                          "mergeCommit": {"oid": "fcd1c890dc1526f4d62ceedad561f498195c8939"},
```
bcoe, line 90: I switched the OID to an OID earlier in the list, so that we covered the edge-case addressed in this PR.

PR #457, opened 2020-06-10
feat(java-yoshi): handle build.gradle and dependencies.properties files

chingor13:
This will allow release-please to handle repos like gax-java and other gradle projects we manage


src/releasers/java-yoshi.ts
```diff
@@ -191,7 +192,13 @@ export class JavaYoshi extends ReleasePR {
       })
     );
 
-    const pomFiles = await this.gh.findFilesByFilename('pom.xml');
+    const pomFilesSearch = this.gh.findFilesByFilename('pom.xml');
+    const buildFilesSearch = this.gh.findFilesByFilename('build.gradle');
```
jef, line 196: what about `build.gradle.kts`?

PR #468, opened 2020-06-14
feat: detect default branch

bcoe:
Now programmatically detects default branch, making it easier to use an alternative.

This also should make it easier to support multiple release branches.

@feywind I expect this overlaps a bit with your work, but might make it a bit easier to see that work over the finish line 👌


src/graphql-to-commits.ts
```diff
@@ -124,7 +122,7 @@ async function graphqlToCommit(
   let prEdge: PREdge = commitEdge.node.associatedPullRequests.edges[0];
 
   // if the commit.sha and mergeCommit.oid do not match, assume that this
-  // was a push to master and drop the commit.
+  // was a push directl to the default branch.
```
JustinBeckwith, line 125: typo in the comment

PR #479, opened 2020-06-30
feat(snapshot): skip snapshot releases for releasers that lack this concept

bcoe:
releasers now need to indicate that they support nightly snapshot builds, before they will start running.

### DIFF
--- a/test/releasers/node.ts
+++ b/test/releasers/node.ts
@@ -190,5 +190,16 @@ describe('Node', () => {
       await releasePR.run();
       req.done();
     });
+    it('does not support snapshot releases', async () => {
+      const releasePR = new Node({
+        repoUrl: 'googleapis/node-test-repo',
+        releaseType: 'node',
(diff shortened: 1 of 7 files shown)


test/releasers/node.ts
```diff
@@ -190,5 +190,16 @@ describe('Node', () => {
...
+    it('does not support snapshot releases', async () => {
+      const releasePR = new Node({
+        repoUrl: 'googleapis/node-test-repo',
+        releaseType: 'node',
+        // not actually used by this type of repo.
+        packageName: 'node-test-repo',
+        apiUrl: 'https://api.github.com',
+        snapshot: true,
+      });
+      await releasePR.run();
```
chingor13, line 202: This test doesn't assert anything - does it only fail via nock disabling network access?

bcoe: Yes, we haven't mocked any of the routes, so it fails loudly with a ton of attempted network connections.

PR #504, opened 2020-07-28
fix(java-snapshot): handle no new commits since release

### LINKED ISSUE #500: Cannot read property 'sha' of undefined (Java)
We're seeing a bunch of these in the `repo-automation-bots` logs:

```
TypeError: Cannot read property 'sha' of undefined
at JavaYoshi._run (/workspace/node_modules/release-please/build/src/releasers/java-yoshi.js:84)
```

It looks like it's possible for the number of commits in this situation to be 0:

```js
if (this.snapshot) {
      const latestCommit = (
        await this.commits(latestTag ? latestTag.sha : undefined, 1, true)
      )[0];
      prSHA = latestCommit.sha;
    }
```

In this case, the size of the value returned from `this.commits` is 0.

bcoe:
Handles the case where there have been no new commits since the last release, and the snapshot cron executes.

_Note: @chingor13 I notice that [this snapshot](https://github.com/googleapis/java-storage/pull/432) was created, does this mean that our cron is working as long as the latest release isn't the most recent?_

Fixes #500

### DIFF
--- a/test/releasers/java-yoshi.ts
+++ b/test/releasers/java-yoshi.ts
@@ -368,6 +368,141 @@ describe('JavaYoshi', () => {
     await releasePR.run();
     req.done();
   });
+  it('creates a snapshot PR, when latest release sha is head', async () => {
+    const versionsContent = readFileSync(
+      resolve(fixturesPath, 'released-versions.txt'),
+      'utf8'
+    );
+    const pomContents = readFileSync(resolve(fixturesPath, 'pom.xml'), 'utf8');
+    const graphql = JSON.parse(
+      readFileSync(resolve(fixturesPath, 'empty-commits.json'), 'utf8')
+    );
+    const req = nock('https://api.github.com')
+      .get('/repos/googleapis/java-trace/pulls?state=closed&per_page=100')
+      .reply(200, undefined)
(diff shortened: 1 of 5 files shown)


test/releasers/java-yoshi.ts
```diff
@@ -368,6 +368,141 @@ describe('JavaYoshi', () => {
...
+  it('creates a snapshot PR, when latest release sha is head', async () => {
+    const versionsContent = readFileSync(
+      resolve(fixturesPath, 'released-versions.txt'),
+      'utf8'
+    );
+    const pomContents = readFileSync(resolve(fixturesPath, 'pom.xml'), 'utf8');
+    const graphql = JSON.parse(
+      readFileSync(resolve(fixturesPath, 'empty-commits.json'), 'utf8')
+    );
+    const req = nock('https://api.github.com')
```
JustinBeckwith, line 380: That mock chain is intense bro.

bcoe: Yes, this is just copied from the other snapshot PR that has commits. It might be better to start moving to sinon and mocking our github wrapper rather than the GitHub requests.

PR #514, opened 2020-08-05
feat(go): add support for releasing yoshi gapic go libraries

bcoe:
This adds support for releasing the Yoshi gapic go libraries.

TODO:

- [x] get the CHANGES.md closer to the current style.
- [x] properly build up a combined gapic PR, as we dedupe it.
- [ ] setup the configuration in the go repository, so this actually works.

Future Work:

* figure out the manual yoshi libraries.

CC: @codyoss

### DIFF
--- a/__snapshots__/yoshi-go.js
+++ b/__snapshots__/yoshi-go.js
@@ -1,0 +1,29 @@
+exports['CHANGELOG-go-yoshi-message'] = `
+chore: created CHANGELOG.md
+`
+
+exports['CHANGELOG-go-yoshi'] = `
+# Changelog
+
+## [0.124.0](https://www.github.com/googleapis/yoshi-go-test-repo/compare/v0.123.4...v0.124.0) 
+
+
+### Bug Fixes
+
+* **automl:** update dependency com.google.cloud:google-cloud-storage to v1.120.0 ([fcd1c89](https://www.github.com/googleapis/yoshi-go-test-repo/commit/fcd1c890dc1526f4d62ceedad561f498195c8939))
+* **deps:** update dependency com.google.cloud:google-cloud-spanner to v1.50.0 ([1f9663c](https://www.github.com/googleapis/yoshi-go-test-repo/commit/1f9663cf08ab1cf3b68d95dee4dc99b7c4aac373))
+
+`
+
+exports['PR body-go-yoshi'] = {
+  'title': 'chore: release 0.124.0',
+  'body': ':robot: I have created a release \\*beep\\* \\*boop\\* \n---\n## [0.124.0](https://www.github.com/googleapis/yoshi-go-test-repo/compare/v0.123.4...v0.124.0) \n\n\n### Bug Fixes\n\n* **automl:** update dependency com.google.cloud:google-cloud-storage to v1.120.0 ([fcd1c89](https://www.github.com/googleapis/yoshi-go-test-repo/commit/fcd1c890dc1526f4d62ceedad561f498195c8939))\n* **deps:** update dependency com.google.cloud:google-cloud-spanner to v1.50.0 ([1f9663c](https://www.github.com/googleapis/yoshi-go-test-repo/commit/1f9663cf08ab1cf3b68d95dee4dc99b7c4aac373))\n---\n\n\nThis PR was generated with [Release Please](https://github.com/googleapis/release-please).',
+  'head': 'release-v0.124.0',
+  'base': 'master'
+}
+
+exports['labels-go-yoshi'] = {
+  'labels': [
+    'autorelease: pending'
+  ]
+}
--- a/__snapshots__/yoshi-go.js
+++ b/__snapshots__/yoshi-go.js
@@ -8,22 +8,27 @@ exports['CHANGES-go-yoshi'] = `
 ## [0.124.0](https://www.github.com/googleapis/yoshi-go-test-repo/compare/v0.123.4...v0.124.0) 
 
 
+### Features
+
+* **all:** auto-regenerate gapics ([#1000](https://www.github.com/googleapis/yoshi-go-test-repo/issues/1000)) ([#1001](https://www.github.com/googleapis/yoshi-go-test-repo/issues/1001))
+* **asset:** added a really cool feature ([d7d1c89](https://www.github.com/googleapis/yoshi-go-test-repo/commit/d7d1c890dc1526f4d62ceedad581f498195c8939))
+
+
 ### Bug Fixes
 
-* **automl:** update dependency com.google.cloud:google-cloud-storage to v1.120.0 ([fcd1c89](https://www.github.com/googleapis/yoshi-go-test-repo/commit/fcd1c890dc1526f4d62ceedad561f498195c8939))
-* **deps:** update dependency com.google.cloud:google-cloud-spanner to v1.50.0 ([1f9663c](https://www.github.com/googleapis/yoshi-go-test-repo/commit/1f9663cf08ab1cf3b68d95dee4dc99b7c4aac373))
+* **automl:** fixed a really bad bug ([d7d1c89](https://www.github.com/googleapis/yoshi-go-test-repo/commit/d7d1c890dc1526f4d62ceedad581f498195c8939))
 
 `
 
 exports['PR body-go-yoshi'] = {
-  'title': 'chore: release 0.124.0',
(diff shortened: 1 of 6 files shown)


src/releasers/go-yoshi.ts
```diff
@@ -0,0 +1,151 @@
+// * we need a deny list of modules that are mono repos, which we keep track of
```
JustinBeckwith, line 1: Not sure what's going on with the comment block here.  At the very least, this should be below the copyright.

src/releasers/go-yoshi.ts
```diff
@@ -0,0 +1,151 @@
...
+
+import {ReleaseType} from 'semver';
+import * as semver from 'semver';
+
+// Generic
+import {Changelog} from '../updaters/changelog';
+
+// These repos should be ignored when generating a release for the
+// top level mono-repo:
+const IGNORED_PATHS = [
```
JustinBeckwith, line 38: I'm ok with this for now, but ...  I do wonder out loud what a more sustainable model for `release-please` looks like.  I like the idea of built-in support for monorepo setups, and for split repo setups across languages.  For cases where there is very customer configuration like this, I'd very much like to start thinking about a config format you could use so this code can remain more generalized.  This isn't blocking feedback for this PR, just a seed to plant.

src/releasers/go-yoshi.ts
```diff
@@ -0,0 +1,151 @@
...
+  'pubsub',
+  'spanner',
+  'storage',
+];
+const GAPIC_PR_REGEX = /.*auto-regenerate gapics.*/;
+
+export class GoYoshi extends ReleasePR {
+  static releaserName = 'go-yoshi';
+  protected async _run() {
+    const latestTag: GitHubTag | undefined = await this.gh.latestTag(
```
JustinBeckwith, line 53: No need to include a type here, it should be inferred from `this.gh.latestTag` I would guess.

src/releasers/go-yoshi.ts
```diff
@@ -0,0 +1,151 @@
...
+      this.monorepoTags ? `${this.packageName}-` : undefined
+    );
+    let gapicPR: Commit;
+    const commits: Commit[] = (
```
JustinBeckwith, line 57: Same feedback as above (and throughout) - no need to be overly specific with type definitions if the functions return good types.

src/releasers/go-yoshi.ts
```diff
@@ -0,0 +1,151 @@
...
+      await this.commits({
+        sha: latestTag ? latestTag.sha : undefined,
```
JustinBeckwith, line 59:
```js
sha: latestTag?.sha
```

src/releasers/go-yoshi.ts
```diff
@@ -0,0 +1,151 @@
...
+        path: this.path,
+      })
+    ).filter(commit => {
+      for (const ignoredPath of IGNORED_PATHS) {
```
JustinBeckwith, line 63: A few more comments on what's happening through here for future generations would be lovely :)

src/releasers/go-yoshi.ts
```diff
@@ -0,0 +1,151 @@
...
+        const re = new RegExp(`^\\w+\\(${ignoredPath}.*\\)`);
+        if (re.test(commit.message)) {
+          return false;
+        }
+      }
+      if (GAPIC_PR_REGEX.test(commit.message)) {
+        if (gapicPR) {
+          const issueRe = /.*(?<pr>\(.*\))$/;
+          const match = commit.message.match(issueRe);
+          if (match && match?.groups?.pr) {
```
JustinBeckwith, line 73: The `match?` should cover checking for `match` first.

test/releasers/yoshi-go.ts
```diff
@@ -0,0 +1,124 @@
...
+import {readFileSync} from 'fs';
+import {resolve} from 'path';
+import * as snapshot from 'snap-shot-it';
+
+const fixturesPath = './test/releasers/fixtures/yoshi-go';
+
+interface MochaThis {
+  [skip: string]: Function;
+}
+function requireNode10(this: MochaThis) {
```
JustinBeckwith, line 27: What is happening here?  Why are we specifically trying to add logic to avoid running tests on node 8 and below?

test/releasers/yoshi-go.ts
```diff
@@ -0,0 +1,124 @@
...
+}
+
+describe('YoshiGo', () => {
+  describe('run', () => {
+    before(requireNode10);
+    it('creates a release PR', async () => {
+      const graphql = JSON.parse(
+        readFileSync(resolve(fixturesPath, 'commits.json'), 'utf8')
+      );
+      const req = nock('https://api.github.com')
```
JustinBeckwith, line 41: Is it worth throwing a `nock.disableNetConnect()` in the `before`?

src/releasers/go-yoshi.ts
codyoss: I think we will eventually want this to be programatic, especially if we ever start split out into more and more sub-modules. But for now, this "works" for the top level module at least. I am still grokking what this code has context to. I have some ideas in the back of my head how we could make this better in the future. For now... 🤔

test/releasers/yoshi-go.ts
bcoe: copy pasta, we can get rid of this.

src/releasers/go-yoshi.ts
bcoe: so, we do have some support for mono-repos now:

https://github.com/googleapis/release-please/pull/501

But I was thinking this logic is a little yoshi  specific...

What I've been thinking  would be nice, would be to keep these yoshi  specific strategies  somewhat hidden, i.e.,  not calling them out  in documentation. And, when we add a strategy like yoshi-java, or yoshi-go, it would be nice to also add corresponding strategies that are closer to something the community could use.

codyoss:
I think this is good enough for the core parent module imo. I am not sure what is up with the failing header check, snapshots should be exempts according to the config...

@bcoe if you want to take a look again too, can't add you as a review since you opened the PR.

bcoe:
@codyoss we should add an exception for the `__snapshot__` folder to the header checker, it wants there to be copyrights on these auto generated fixtures (it can be ignored in this case).

__snapshots__/yoshi-go.js
```diff
@@ -0,0 +1,34 @@
...
+
+exports['CHANGES-go-yoshi'] = `
+# Changelog
+
+## [0.124.0](https://www.github.com/googleapis/yoshi-go-test-repo/compare/v0.123.4...v0.124.0) 
+
+
+### Features
+
+* **all:** auto-regenerate gapics ([#1000](https://www.github.com/googleapis/yoshi-go-test-repo/issues/1000)) ([#1001](https://www.github.com/googleapis/yoshi-go-test-repo/issues/1001))
```
bcoe, line 13: we're okay with having the module name in **bold** rather than the current formatting of your changelog?

src/releasers/go-yoshi.ts
```diff
@@ -0,0 +1,159 @@
...
+    );
+
+    let changelogEntry: string = await cc.generateChangelogEntry({
+      version: candidate.version,
+      currentTag: `v${candidate.version}`,
+      previousTag: candidate.previousTag,
+    });
+
+    // remove commit reference from auto-generate gapics
+    changelogEntry = changelogEntry.replace(/\s\(\[null\].*\n/, '\n');
```
bcoe, line 108: I find this logic a little bit weird, I'll take a look before I land and see if we can simplify.

__snapshots__/yoshi-go.js
codyoss: I am fine either way. Bolding is fine.

codyoss:
@bcoe It looks like snapshots should already be ignored if I am understanding this right.

https://github.com/googleapis/release-please/blob/a6c9e0bcf6a301f28ccbaa5db458ae1393ffd9ee/.bots/header-checker-lint.json#L2-L4

bcoe:
@codyoss weird, it's `__snapshots__/yoshi-go.js is missing a valid license header` that's being flagged.

Should that ignore rule be `**/__snapshots__/**`, I'm terrible at globs.

codyoss:
@bcoe I rule looked right to me, and it works for all the other files in that directory. I _think_ `**` is only need for recursive dirs.

PR #518, opened 2020-08-12
fix(java): detect need for snapshot unless a snapshot is explicitly requested

chingor13:
It seems that snapshot is being explicitly set to false. This is preventing snapshots being created except when snapshot is explicitly being set to true. (Snapshots are only being created by the nightly cron bot job).


chingor13:
For the test that's failing, I don't think that condition is actually valid. If the repo needs a snapshot, we should propose the snapshot, otherwise the repo is blocked until someone (cron) explicitly asks for a snapshot.

chingor13:
header-check is failing on the snapshot js files which are generated and will be stomped on

JustinBeckwith:
@chingor13 does the config for `header-check` support ignore files?

PR #534, opened 2020-09-01
fix(simple): create version.txt if it does not exist

bcoe:
there's no reason we can't just create the `version.txt` if it doesn't exist.


bcoe:
@JustinBeckwith if someone doesn't yet have a `version.txt` in their repository, for a simple release (_a release with no package.json that updates a `version.txt`_), it will create the file.

PR #538, opened 2020-09-03
feat:  pass an array of change log section options to the node releaser

### LINKED ISSUE #537: Pass in change log section Options
As discussed in this issue, https://github.com/cloudevents/sdk-javascript/issues/312#issuecomment-682308445, We are currently using standard release and we override the default changelog types that get output to the CHANGELOG.md

Currently, this library and the GH action that uses it, doesn't yet support modifying the sections that get output to the changelog

The feature i'm proposing is that a user can pass in an Array of type values,  like:

```
"types": [
      {
        "type": "feat",
        "section": "Features",
        "hidden": false
      },
      {
        "type": "docs",
        "section": "Documentation",
        "hidden": false
      },
      {
        "type": "chore",
        "section": "Miscellaneous",
        "hidden": false
      },
      {
        "type": "test",
        "section": "Tests",
        "hidden": false
      }
    ]
```

to modify what sections are output.

So an example using the `ReleaseFactory.buidStatic` method might look something like this:

```
const changelogSections = [{...}, {...}]

const release = ReleasePRFactory.buildStatic(releaseType, {
   ...
    changelogSections
  })
```

lholmquist:
Fixes #537

### DIFF
--- a/src/release-pr.ts
+++ b/src/release-pr.ts
@@ -55,6 +55,7 @@ export interface BuildOptions {
 
 export interface ReleasePROptions extends BuildOptions {
   releaseType: string;
+  changelogSections?: [];
 }
 
 export interface ReleaseCandidate {
@@ -96,6 +97,7 @@ export class ReleasePR {
   proxyKey?: string;
   snapshot?: boolean;
   lastPackageVersion?: string;
+  changelogSections?: [];
 
   constructor(options: ReleasePROptions) {
     this.bumpMinorPreMajor = options.bumpMinorPreMajor || false;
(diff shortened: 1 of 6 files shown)


lholmquist:
@googlebot I signed it!

bcoe (changes requested):
Mind adding this for the main languages we advertise in the action:

`simple`, `terraform-module`, `ruby`,  `python`, `node`?

I would just double check that none of them override the value already,  if they do we can just make the override the default.

Thank you for the  contribution 👏

bcoe (approved):

bcoe:
@lholmquist thank you for the contributions 👌

PR #553, opened 2020-09-14
chore: migrate to Trampoline V2

tmatsuo:
I think `pupulate-secrets.sh` is now using the service account attached
to the Kokoro VM, so I updated the comments in that file.

`.trampolinerc` will be sourced from `trampoline_v2.sh` in its early
stage. `.trampolinerc` sources `populate-secrets.sh`.

I confirmed that `populate-secrets.sh` works as expected in [this build](https://source.cloud.google.com/results/invocations/93aede25-b754-4399-b730-8f4d18fc825f/targets). Here is the log:

```
2020-09-15 15:11:37 Secret written to /tmpfs/src/gfile/secret_manager/npm_publish_token
```

To run the build locally, you can do something like:

```sh
$ docker pull gcr.io/cloud-devrel-kokoro-resources/node:12-user
$ TRAMPOLINE_IMAGE=gcr.io/cloud-devrel-kokoro-resources/node:12-user \
    TRAMPOLINE_BUILD_FILE=.kokoro/system-test.sh \
    SECRET_MANAGER_KEYS="npm_publish_token" \
    .kokoro/trampoline_v2.sh
```

### DIFF
--- a/.kokoro/trampoline_v2.sh
+++ b/.kokoro/trampoline_v2.sh
@@ -1,0 +1,487 @@
+#!/usr/bin/env bash
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# trampoline_v2.sh
+#
+# This script does 3 things.
+#
+# 1. Prepare the Docker image for the test
+# 2. Run the Docker with appropriate flags to run the test
+# 3. Upload the newly built Docker image
+#
+# in a way that is somewhat compatible with trampoline_v1.
+#
+# To run this script, first download few files from gcs to /dev/shm.
+# (/dev/shm is passed into the container as KOKORO_GFILE_DIR).
+#
+# gsutil cp gs://cloud-devrel-kokoro-resources/python-docs-samples/secrets_viewer_service_account.json /dev/shm
+# gsutil cp gs://cloud-devrel-kokoro-resources/python-docs-samples/automl_secrets.txt /dev/shm
+#
+# Then run the script.
+# .kokoro/trampoline_v2.sh
+#
+# These environment variables are required:
+# TRAMPOLINE_IMAGE: The docker image to use.
+# TRAMPOLINE_DOCKERFILE: The location of the Dockerfile.
+#
+# You can optionally change these environment variables:
+# TRAMPOLINE_IMAGE_UPLOAD:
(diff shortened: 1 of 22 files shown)


tmatsuo:
Samples test succeeded with weird message:

```
================================================================
2020-09-14T18:30:23Z: Running the tests in a Docker container.
================================================================
npx: installed 31 in 2.598s
Error: Wanted node version 10 (>=10.0.0 <11.0.0)
node: 12.18.3
To install node, run `nvm install 10` or see https://nodejs.org/
coverage is only reported for Node 10
================================================================
2020-09-14T18:30:32Z: Build finished with 0
================================================================
```

tmatsuo:
I think we should confirm that `populate-secrets.sh` successfully populate at least one secret.

Maybe we can experiment with one of the presubmit builds.

tmatsuo:
I confirmed that `populate-secrets.sh` works as expected in [this build](https://source.cloud.google.com/results/invocations/c2406c34-4436-4682-b279-ec6ff31c38aa/targets). Here is the log:

```
2020-09-14 14:09:45 Secret written to /tmpfs/src/gfile/secret_manager/npm_publish_token
```

I'll revert the change for this experiment.

.kokoro/trampoline_v2.sh
```diff
@@ -0,0 +1,487 @@
...
+# TRAMPOLINE_IMAGE_UPLOAD:
+#     (true|false): Whether to upload the Docker image after the
+#                   successful builds.
+# TRAMPOLINE_BUILD_FILE: The script to run in the docker container.
+# TRAMPOLINE_WORKSPACE: The workspace path in the docker container.
+#                       Defaults to /workspace.
+# Potentially there are some repo specific envvars in .trampolinerc in
+# the project root.
+
+
```
bcoe, line 49: This is equivalent to the `python3 "${KOKORO_GFILE_DIR}/trampoline_v1.py"` that was called by `trampoline.sh` prior to this?

tmatsuo: Yes, but there is some incompatibilities. For more details, read the [migration guide](https://github.com/GoogleCloudPlatform/docker-ci-helper/blob/master/MIGRATING_FROM_V1.md).

bcoe (commented):
Thanks for putting in this work, I'm going to test on another repo first, and then see about rolling this out more widely.

tmatsuo:
@bcoe It sounds good. If you have a specific repository in your mind? I'd be happy to create a similar PR there.

bcoe:
@tmatsuo actually, the best approach might be to make the change here:

https://github.com/googleapis/synthtool/tree/master/synthtool/gcp/templates/node_library/.kokoro

Then I can run `synthtool` from a branch on any of our repos, and we can see if it works; I was thinking I'd pick a library like `nodejs-phishing-protection`.

tmatsuo:
@bcoe Yeah, but in the first step of the migration, it's easier to work on the actual repo.

bcoe:
I think this would be a good choice:

https://github.com/googleapis/nodejs-phishing-protection

It was useful to see this in action, but I should have thought of the fact that we don't have sample tests on this repo (sorry about the extra work).

tmatsuo:
@bcoe No problem, I'll create a similar PR there.

sofisl (approved):

PR #587, opened 2020-10-09
fix: issues found by testing with the GCP ruby repo

feywind:
Related to https://github.com/googleapis/release-please/issues/428

Some of the monorepos were having troubles because of changes to the way tags are looked up. The new regex wasn't handling complex names like "foo-bar-baz_quux-v1.2-v1.1.0". This should fix that and add a regression test.


feywind:
Adding on to this a few other issues we found while testing with that monorepo. The graphql query was sometimes returning results out of order, and we were grabbing merged pull requests by update time rather than merge date.

PR #600, opened 2020-10-26
docs: Yarn how to comment

### LINKED ISSUE #599: Add info regarding relevant yarn command
In the NPM publishing example, there is a step that indicates the following:

```
- run: npm ci
if: ${{ steps.release.outputs.release_created }}
```

If you have your action set-up like this but you use Yarn, you will get the following error when the action is run:

```
cipm can only install packages with an existing package-lock.json or npm-shrinkwrap.json...
```

Add a comment to the example showing the alternative command to substitute if one uses Yarn.

schalkneethling:
Adds a comment to the NPM publishing example to indicate the command to use when using Yarn as opposed to NPM

Fixes #599 🦕


README.md
```diff
@@ -119,6 +119,7 @@ jobs:
           node-version: 12
           registry-url: 'https://registry.npmjs.org'
         if: ${{ steps.release.outputs.release_created }}
+      # if you are using Yarn, substitute the command below with `yarn install --frozen-lockfile`
       - run: npm ci
         if: ${{ steps.release.outputs.release_created }}
       - run: npm publish
```
bcoe, line 125: Do folks using yarn still use `npm` for publication?

schalkneethling: Because I only use Yarn, I do not have a `package.lock` nor a `shrinkwrap` file in my repo and that caused the above to fail with the error mentioned in the issue.

bcoe (approved):

bcoe:
@schalkneethling thank you for the contribution 👍

PR #604, opened 2020-10-27
fix(go): ignore tags with prefix, if no prefix provided

bcoe:
If no prefix is provided, only return branches that also have no prefix.


codyoss (approved):
LGTM, thanks Ben!

alexander-fenster (approved):

PR #606, opened 2020-10-28
fix(graphql): graphql query requested pagination at wrong level/did n…

bcoe:
When the `graphql` logic was reverted to use `ref`, pagination was left at the wrong level in the graphql query.

The variable being set for baseRef was also incorrect.


chingor13 (approved):

AVaksman (approved):

JustinBeckwith (approved):
You got any of them tests?

PR #610, opened 2020-10-28
feat(java): support special case of promotion to 1.0.0

chingor13:
Adds support for the special case of promoting a Java library to 1.0.0.

For our Java repos, we use `bumpMinorPreMajor` for breaking changes. We also manage multiple artifact versions in a `versions.txt` manifest.

For the special case of promoting to 1.0.0 (via `Release-As: 1.0.0` in a commit body metadata), we want to bump all "GA" artifacts to 1.0.0. For example, we release generated proto classes for the API endpoints. For the v1 service, these protos are considered stable so the proto-google-cloud-[service]-v1 artifact should be versioned as stable (1.0.0). For the v1beta service, these protos are considered unstable so we should leave them at 0.x.

Our current workaround is to manually update the version strings in the 1.0.0 release PR, which is error prone and extra work.


chingor13:
cc @stephaniewang526

test/releasers/java-yoshi.ts
```diff
@@ -580,4 +580,158 @@ describe('JavaYoshi', () => {
...
+        '/search/code?q=filename%3Adependencies.properties+repo%3Agoogleapis%2Fjava-trace'
+      )
+      .reply(200, {
+        total_count: 0,
+        items: [],
+      })
+      // check for CHANGELOG
+      .get(
+        '/repos/googleapis/java-trace/contents/CHANGELOG.md?ref=refs%2Fheads%2Fmaster'
+      )
```
bcoe, line 673: @chingor13 if it's too much work fo this PR, we don't have to do it, but how do you feel about this refactor to how we do testing:

https://github.com/googleapis/release-please/blob/master/test/releasers/simple.ts

I think it cuts down on the giant `nock` stubs, which I like.

chingor13: Let's do it in a follow-up PR

bcoe (approved):

PR #617, opened 2020-11-09
feat(go): add support for mono-repo google-cloud-go repo

bcoe:
Makes `github-release` able to handle mono-repos, when `monorepo-tags` is set to `true`.

### DIFF
--- a/src/github-release.ts
+++ b/src/github-release.ts
@@ -13,8 +13,6 @@
 // limitations under the License.
 
 import chalk = require('chalk');
-import {join} from 'path';
-
 import {checkpoint, CheckpointType} from './util/checkpoint';
 import {ReleasePRFactory} from './release-pr-factory';
 import {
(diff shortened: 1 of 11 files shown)


src/releasers/go-yoshi.ts
```diff
@@ -187,4 +204,26 @@ export class GoYoshi extends ReleasePR {
   protected defaultInitialVersion(): string {
     return '0.1.0';
   }
+
+  static tagSeparator(): string {
```
codyoss, line 208: This should be moved the the sub-module code I think?

src/releasers/go-yoshi-submodule.ts
```diff
@@ -0,0 +1,128 @@
+// Copyright 2019 Google LLC
```
codyoss, line 1: nit: 2020

src/github.ts
```diff
@@ -461,17 +462,24 @@ export class GitHub {
   // the branch we're configured for.
   async latestTag(
     prefix?: string,
-    preRelease = false
+    preRelease = false,
+    // Allow a branch prefix to differ from a tag prefix. This was required
+    // for google-cloud-go, which uses the tag prefix library/vx.y.z.
+    // this is a requirement in the go community.
+    branchPrefix?: string
```
codyoss, line 469: Is this still needed with the new separator method?

src/releasers/go-yoshi.ts
bcoe: I think we want it in the top level releaser, because this is what the step that creates GitHub tags looks at.

src/github.ts
bcoe: Unfortunately yes, because the very first release of each library will hit `latestTagFallback`, as we've never created a branch in the `release-[library]-version` format before.

codyoss: ah, gotcha.

src/github.ts
```diff
@@ -72,7 +72,7 @@ import {Update} from './updaters/update';
 // - make sure the branch name starts with "release"
 // - take everything else
 // This includes the tag to handle monorepos.
-const VERSION_FROM_BRANCH_RE = /^.*:release-(.*)+$/;
+const VERSION_FROM_BRANCH_RE = /^.*:release-?([\w-.]*)-(v[0-9].*)+$/;
```
alexander-fenster, line 75: Why do we need this `+` in the end?  The `v[0-9].*` part will consume everything that starts from `v` followed by a digit, up to the end of the line.

test/github-release.ts
```diff
@@ -41,7 +41,7 @@ describe('GitHubRelease', () => {
         .get('/repos/googleapis/foo')
         .reply(200, repoInfo)
         .get(
-          '/repos/googleapis/foo/pulls?state=closed&per_page=100&sort=merged_at&direction=desc'
+          '/repos/googleapis/foo/pulls?state=closed&per_page=25&sort=merged_at&direction=desc'
```
JustinBeckwith, line 44: The magic limit that appears to be here makes me nervous.  Should it?

JustinBeckwith (changes requested):

src/releasers/go-yoshi-submodule.ts
```diff
@@ -0,0 +1,109 @@
...
+import {Update} from '../updaters/update';
+
+// Generic
+import {Changelog} from '../updaters/changelog';
+
+const SCOPE_REGEX = /^\w+\((?<scope>.*)\):/;
+
+export class GoYoshiSubmodule extends ReleasePR {
+  static releaserName = 'go-yoshi-submodule';
+  protected async _run() {
```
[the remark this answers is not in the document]
bcoe: We can try to move away from the `_run`, today this is called by `run()` in `release-pr`, so we would need to refactor all of our releasers to move away from it.

src/github-release.ts
```diff
@@ -69,67 +72,99 @@ export class GitHubRelease {
...
+    if (this.monorepoTags) {
+      gitHubReleasePRs = await this.gh.findMergedReleasePRs(this.labels, 25);
     } else {
+      const releasePR = await this.gh.findMergedReleasePR(this.labels, 25);
+      if (releasePR) gitHubReleasePRs = [releasePR];
+    }
+    if (gitHubReleasePRs.length === 0) {
       checkpoint('no recent release PRs found', CheckpointType.Failure);
       return undefined;
+    } else {
```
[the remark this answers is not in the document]
fhinkel: Shouldn't there be a lint rule for that?

fhinkel (approved):
lgtm with Justin's comments addressed.

src/github-release.ts
```diff
@@ -69,67 +72,99 @@ export class GitHubRelease {
...
-      // has not yet been created.
-      await this.gh.removeLabels(this.labels, gitHubReleasePR.number);
-      return release;
+    let gitHubReleasePRs: GitHubReleasePR[] = [];
+    const rootPath = this.path;
+    const releases = [];
+    if (this.monorepoTags) {
+      gitHubReleasePRs = await this.gh.findMergedReleasePRs(this.labels, 25);
     } else {
+      const releasePR = await this.gh.findMergedReleasePR(this.labels, 25);
```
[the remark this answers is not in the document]
bcoe: We run the `github-release` step immediately after a PR is merged, so I figured that a page size of `100` should be far more than we need.

`25` leaves a bit of a buffer, in case something odd happens, e.g., we miss running release-please for a couple merged PRs, but in practice this could probably be a lower number.

test/github-release.ts
bcoe: Left a comment about why `25` was chosen.

src/github-release.ts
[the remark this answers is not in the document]
bcoe: left a comment.

src/github-release.ts
```diff
@@ -69,67 +72,99 @@ export class GitHubRelease {
...
+        checkpoint(
+          `found release notes: \n---\n${chalk.grey(
+            latestReleaseNotes
+          )}\n---\n`,
+          CheckpointType.Success
+        );
+
+        // Attempt to lookup the package name from a well known location, such
+        // as package.json, if none is provided:
+        if (this.packageName === undefined && this.releaseType) {
```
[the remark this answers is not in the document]
bcoe: `this.packageName` is passed in as config from a `yaml` file, I think you're right that we should just check for `!this.packageName`.

src/github-release.ts
```diff
@@ -69,67 +72,99 @@ export class GitHubRelease {
...
+        }
+        // Go uses '/' for a tag separator, rather than '-':
+        let tagSeparator = '-';
+        if (this.releaseType) {
+          tagSeparator = ReleasePRFactory.class(
+            this.releaseType
+          ).tagSeparator();
+        }
+        if (this.packageName === undefined) {
+          throw Error('could not determine package name for release');
```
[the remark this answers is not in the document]
bcoe: added repository to error.

src/github.ts
```diff
@@ -461,17 +462,24 @@ export class GitHub {
...
+    // Allow a branch prefix to differ from a tag prefix. This was required
+    // for google-cloud-go, which uses the tag prefix library/vx.y.z.
+    // this is a requirement in the go community.
+    branchPrefix?: string
   ): Promise<GitHubTag | undefined> {
-    const pull = await this.findMergedReleasePR([], 100, prefix, preRelease);
+    const pull = await this.findMergedReleasePR(
+      [],
+      100,
+      branchPrefix ?? prefix,
```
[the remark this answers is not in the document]
bcoe: this is intentional 👍, `go` is a special case where the branch format and tag format differ (so we broke this out into two flags). Other languages will not be setting the value.

src/github-release.ts
```diff
@@ -69,67 +72,99 @@ export class GitHubRelease {
...
+        // Add a label indicating that a release has been created on GitHub,
+        // but a publication has not yet occurred.
+        await this.gh.addLabels([GITHUB_RELEASE_LABEL], gitHubReleasePR.number);
+        // Remove 'autorelease: pending' which indicates a GitHub release
+        // has not yet been created.
+        await this.gh.removeLabels(this.labels, gitHubReleasePR.number);
+        releases.push(release);
+      }
     }
+    return releases[0];
```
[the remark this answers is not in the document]
bcoe: we initialized it as `[]` so if we didn't find a release, it returns `undefined`, which is the same behavior as before.

JustinBeckwith (approved):

PR #621, opened 2020-11-10
fix(go): we should not include - in the suffix

bcoe:
We were looking for a `-` suffix in the go releaser and should not be.

### DIFF
--- a/src/releasers/go-yoshi.ts
+++ b/src/releasers/go-yoshi.ts
@@ -176,7 +176,7 @@ export class GoYoshi extends ReleasePR {
       changelogEntry,
       updates,
       version: candidate.version,
-      includePackageName: this.monorepoTags,
+      includePackageName: false,
     });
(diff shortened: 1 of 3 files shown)


src/releasers/go-yoshi.ts
```diff
@@ -176,7 +176,7 @@ export class GoYoshi extends ReleasePR {
       changelogEntry,
       updates,
       version: candidate.version,
-      includePackageName: this.monorepoTags,
+      includePackageName: false,
```
bcoe, line 179: go is unique in that the top level release currently represents a "non-mono-repo" release, and submodules represent releases for individual libraries.

alexander-fenster, line 179: Just making sure this is expected to be here (this change is not listed in the description).

alexander-fenster (approved):

PR #627, opened 2020-11-16
fix: include multi-line breaking changes

### LINKED ISSUE #440: Only first line of `BREAKING CHANGE` is shown
#### Environment details

I'm running release-please from @bcoe's Github action https://github.com/bcoe/release-please-action v1.1.0 (running in ubuntu-latest).

#### Steps to reproduce

  1. Create a commit with a multi-line BREAKING CHANGE message.
```
feat: Implement something

BREAKING CHANGE: This commit breaks something which warrants
a description that spans multiple lines.
```
  2. Observe the generated change log, the BREAKING CHANGE only shows the first line.

#### Example

See https://github.com/cfware/babel-plugin-remove-ungap/pull/20 and the associated commit message at https://github.com/cfware/babel-plugin-remove-ungap/commit/a70cca32d5df2a00a609a6c6008a4d936e11cc29.

bcoe:
Include multiline breaking changes, unless separated by two lines.

Fixes #440

### DIFF
--- a/src/conventional-commits.ts
+++ b/src/conventional-commits.ts
@@ -93,7 +93,12 @@ class PostProcessCommits extends Transform {
     done: Function
   ) {
     chunk.notes.forEach(note => {
-      note.text = note.text.split(/\r?\n/)[0];
+      let text = '';
+      for (const chunk of note.text.split(/\r?\n/)) {
+        if (chunk === '') break;
+        else text += `${chunk} `;
+      }
+      note.text = text.trim();
     });
     this.push(JSON.stringify(chunk, null, 4) + '\n');
     done();
--- a/__snapshots__/conventional-commits.js
+++ b/__snapshots__/conventional-commits.js
@@ -15,3 +15,21 @@ exports['ConventionalCommits generateChangelogEntry does not include content two
 
 * upgrade to Node 7 ([abc345](https://www.github.com/bcoe/release-please/commit/abc345))
 `
+
+exports['ConventionalCommits generateChangelogEntry includes multi-line breaking changes 1'] = `
+## v1.0.0 (1665-10-10)
+
+
+### ⚠ BREAKING CHANGES
+
+* we were on Node 6 second line third line
+
+### Features
+
+* awesome feature ([abc678](https://www.github.com/bcoe/release-please/commit/abc678))
+
+
+### Miscellaneous Chores
+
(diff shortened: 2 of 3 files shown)


src/conventional-commits.ts
```diff
@@ -93,7 +93,12 @@ class PostProcessCommits extends Transform {
     done: Function
   ) {
     chunk.notes.forEach(note => {
-      note.text = note.text.split(/\r?\n/)[0];
+      let text = '';
+      for (const chunk of note.text.split(/\r?\n/)) {
+        if (chunk === '') break;
+        else text += `${chunk} `;
```
JustinBeckwith, line 99: Hmmm.  I've been known to write fairly lengthy upgrade guides for folks inside of the extended commit message, with full scale examples of old vs new.  I'm a little worried that removing newlines and replacing those with spaces is going to be a less than ideal experience.  

Can I float a different idea?  Typically in releases with Breaking Changes, I will include a list of commits (exactly what we do now) with the links to the corresponding PRs, but I will also have a dedicated "Breaking Changes" section at the very top of the changelog.  There, I like to go into detail on each breaking change, how to mitigate, etc.  Far more prose than we do these days.  What do you think of iterating over breaking changes, taking the extended commit messages, and doing something more prominent at the top of the log?   Example:
https://github.com/googleapis/google-auth-library-nodejs/releases/tag/v2.0.0

JustinBeckwith (approved):
Approving, since I turned this into a more philosophical discussion, and I noticed the bug in question was OOSLO :)

src/conventional-commits.ts
bcoe: I think this behavior is what you'd expect from the spec:

```
feat: I am a feature that's a breaking change

BREAKING CHANGE: description line one
description line two
description line three

additional content.
more additional content.
```

* the detailed information about the breaking change is in its own section of the CHANGELOG, under breaking changes (_the list of commits will have only the one line description._).
* `additional content` and `more additional content` will not be in the CHANGELOG.

Here's example output from the unit tests:

```md
## v1.0.0 (1665-10-10)

### ⚠ BREAKING CHANGES

* we were on Node 6 second line third line

### Features

* awesome feature ([abc678](https://www.github.com/bcoe/release-please/commit/abc678))
```

I think if we do decide we want to do anything differently, we should  capture it in the spec on conventionalcommits.org, so that all the parsers behave the same.

PR #629, opened 2020-11-16
feat: Add TF versions updater

### LINKED ISSUE #628: Updater for module version in versions.tf

Update versions.tf to the candidate version in a release PR.

Add an updater for replacing old versions if present in versions.tf.

bharathkkb:
Fixes #628 

/cc @morgante

### DIFF
--- a/src/releasers/terraform-module.ts
+++ b/src/releasers/terraform-module.ts
@@ -24,6 +24,7 @@ import {Commit} from '../graphql-to-commits';
 import {Changelog} from '../updaters/changelog';
 // Terraform specific.
 import {ReadMe} from '../updaters/terraform/readme';
+import {ModuleVersion} from '../updaters/terraform/module-version';
 
 export class TerraformModule extends ReleasePR {
   static releaserName = 'terraform-module';
@@ -85,6 +86,19 @@ export class TerraformModule extends ReleasePR {
         packageName: this.packageName,
       })
     );
+    // Update versions.tf to current candidate version.
+    // A module may have submodules, so find all versions.tf to update.
+    const versionFiles = await this.gh.findFilesByFilename('versions.tf');
+    versionFiles.forEach(path => {
+      updates.push(
+        new ModuleVersion({
+          path: this.addPath(path),
+          changelogEntry,
+          version: candidate.version,
+          packageName: this.packageName,
+        })
+      );
+    });
 
     await this.openPR({
       sha: commits[0].sha!,
--- a/test/releasers/terraform-module.ts
+++ b/test/releasers/terraform-module.ts
@@ -116,7 +116,7 @@ describe('terraform-module', () => {
           const fileContent = readFileSync(
             resolve(fixturesPath, readFilePath),
             'utf8'
-          );
+          ).replace(/\r\n/g, '\n');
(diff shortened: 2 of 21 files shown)


test/updaters/module-version.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import {readFileSync} from 'fs';
+import {resolve} from 'path';
+import * as snapshot from 'snap-shot-it';
+import {describe, it} from 'mocha';
+import {ModuleVersion} from '../../src/updaters/terraform/module-version';
+
+const fixturesPath = './test/updaters/fixtures';
+
+describe('versions.tf', () => {
```
morgante, line 23: @bharathkkb Can we add a test fixture for a submodule?

bharathkkb: @morgante sg, currently logic for submodule is in the releaser so we may need to add a test there.
https://github.com/googleapis/release-please/blob/25a9b5296831d147e24e1239eef770188a874601/src/releasers/terraform-module.ts#L90

bcoe: @bharathkkb when adding tests for releasers, take a look at these tests:

https://github.com/googleapis/release-please/blob/master/test/releasers/simple.ts#L43

I've been switching to using sinon and mocking for tests, rather than nock (_which requires you enumerate every HTTP endpoint you hit with the GitHub API). I think you'll find sinon easier.

bharathkkb: @morgante @bcoe added some tests primarily testing the behavior for modules, modules+sub modules and backward compatibility if no versions.tf or module versions defined.

src/releasers/terraform-module.ts
```diff
@@ -85,6 +86,19 @@ export class TerraformModule extends ReleasePR {
         packageName: this.packageName,
       })
     );
+    // Update versions.tf to current candidate version.
+    // A module may have submodules, so find all versions.tf to update.
+    const versionFiles = await this.gh.findFilesByFilename('versions.tf');
```
bcoe, line 91: The one annoying caveat here, is I believe GitHub doesn't index files until they've been merged for a little while -- you might bump into occasional hiccups as new `versions.tf` files are added.

test/releasers/terraform-module.ts
```diff
@@ -0,0 +1,154 @@
...
+      .returns(Promise.resolve([]));
+
+    // Lookup the default branch name:
+    sandbox.stub(releasePR.gh, 'getDefaultBranch').resolves('main');
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+  describe('run', () => {
```
bcoe, line 89: These tests seem really solid to me, do you think it was easier to mock the methods in `github` rather than trying to mock the routes with `nock`?

bcoe (approved):

src/releasers/terraform-module.ts
bharathkkb: Ack, it seems to be a couple of minutes but most of our repos have this so its hopefully indexed.

test/releasers/terraform-module.ts
bharathkkb: I have never used `nock` but the simple-module was a great starting point for me with Sinon.

PR #638, opened 2020-11-24
feat: return pr number, if one was opened

bcoe:
the PR number of an opened PR will now be returned, this allows us to run a GitHub action as a side effect of a PR being opened.


sofisl (approved):

src/release-pr.ts
```diff
@@ -128,29 +128,31 @@ export class ReleasePR {
...
         'snapshot releases not supported for this releaser',
         CheckpointType.Failure
       );
       return;
     }
-    const pr: GitHubReleasePR | undefined = await this.gh.findMergedReleasePR(
-      this.labels
-    );
-    if (pr) {
+    const mergedPR:
```
JustinBeckwith, line 139: No need to be explicit with the type declaration here

src/release-pr.ts
```diff
@@ -128,29 +128,31 @@ export class ReleasePR {
...
       checkpoint(
-        `pull #${pr.number} ${pr.sha} has not yet been released`,
+        `pull #${mergedPR.number} ${mergedPR.sha} has not yet been released`,
         CheckpointType.Failure
       );
+      return undefined;
     } else {
-      return this._run();
+      const openedPR = this._run();
+      return openedPR ? openedPR : undefined;
```
JustinBeckwith, line 151: Huh - is this trying to normalize null and undefined?

JustinBeckwith (approved):

src/release-pr.ts
bcoe: we return `0` from our github helper, I figured for the benefit of the GitHub action `undefined` would be a better signal that we didn't do anything.

I suppose we could switch to `undefined` everywhere.

bcoe: switched to returning `undefined` everywhere.

PR #646, opened 2020-11-30
feat(conventional-commits): support googleapis extension

bcoe:
Adds support for the parsing extension proposed by the googleapis repository.

---

@JustinBeckwith I was mistaken about markdown being stripped out during rendering, I'm open to trying out this approach; if you think it's worth providing customers with the additional context?

### DIFF
--- a/src/conventional-commits.ts
+++ b/src/conventional-commits.ts
@@ -94,15 +94,42 @@ class PostProcessCommits extends Transform {
   ) {
     chunk.notes.forEach(note => {
       let text = '';
+      let i = 0;
+      let extendedContext = false;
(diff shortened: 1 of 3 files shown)


JustinBeckwith (approved):

src/conventional-commits.ts
```diff
@@ -105,6 +116,22 @@ class PostProcessCommits extends Transform {
   }
 }
 
+// If someone wishes to include additional contextual information for a
+// BREAKING CHANGE, they can do so by starting the line after the initial
```
bcoe, line 120:
```suggestion
// BREAKING CHANGE using markdown, they can do so by starting the line after the initial
```

PR #659, opened 2020-12-07
fix(go): pass monorepo-tags to open-pr; omit commits when publishing gapic submodule

bcoe:
we were skipping passing `monorepoTags` with our more complicated approach.

### DIFF
--- a/src/releasers/go-yoshi.ts
+++ b/src/releasers/go-yoshi.ts
@@ -84,6 +84,19 @@ export class GoYoshi extends ReleasePR {
             }
           }
         }
+
+        // In the gapic repo, it's required that commits are prefixed with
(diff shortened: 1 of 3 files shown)


src/releasers/go-yoshi.ts
```diff
@@ -166,7 +166,7 @@ export class GoYoshi extends ReleasePR {
       changelogEntry,
       updates,
       version: candidate.version,
-      includePackageName: false,
+      includePackageName: this.monorepoTags,
```
bcoe, line 169: This logic is the same as all of the other releasers now, we include the package name if the release is for a submodule.

steffnay (approved):

PR #581, opened 2020-10-08
feat: Command and "Runner" Pattern for release-please

jbottigliero:
- Moves Yargs command parser to an isolated file (for reuse and testing).
- Updates command handlers to use a shared "runner" module.
  - This runner is meant to resemble the current `release-please-action`
- Adds `changelog-types` option
- Adds the ability to configure the CLI tool using a JSON file (via `--config` flag)

#### `src/runner.ts`

This module is being used to introduce a new entrypoint to this package that can be used by the internal command _and_ consumers to centralize execution (and configuration mapping) for `release-please` in a CI context. As an example, the `release-please-action` could be replaced with something like the following*:

<details>
 <summary>release-please-action/index.js</summary>

```js
const core = require('@actions/core')
const releasePlease = require('release-please/build/src/runner')

const command = core.getInput('command') ? core.getInput('command') : undefined

const options = {
  label: 'autorelease: pending',
  bumpMinorPreMajor: Boolean(core.getInput('bump-minor-pre-major')),
  monorepoTags: Boolean(core.getInput('monorepo-tags')),
  packageName: core.getInput('package-name'),
  path: core.getInput('path') ? core.getInput('path') : undefined,
  releaseType: core.getInput('release-type'),
  token: core.getInput('token'),
  fork: core.getInput('fork') ? true : undefined,
  changelogTypes: core.getInput('changelog-types'),
  repoUrl: process.env.GITHUB_REPOSITORY
}

releasePlease(options, command, (release) => {
  const { upload_url, tag_name } = release
  core.setOutput('release_created', true)
  core.setOutput('upload_url', upload_url)
  core.setOutput('tag_name', tag_name)
}).catch((e) => {
  core.setFailed(`release-please failed: ${err.message}`)
});
```
</details>

The change is intended to allow for consumers to only need to focus on _process_ input mapping and error/success handling. As another example, using the runner the GitHub Action behaviors can now be ported for usage via Google Cloud Build using [a simple Docker image](https://gist.github.com/jbottigliero/29e8de2dda4db7d3f07626dc99547026)*.

_*_ **This code hasn't been tested.** 

---

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/release-please/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
[...]

PR #674, opened 2020-12-30
refactor: migrate to @conventional-commits/parser and update conventional-changelog

### LINKED ISSUE #601: Investigate bug with Conventional Commits and Node 10 streams
A bug with streams is causing the CHANGELOG ordering to vary, this seems to only crop up on Node 10.

Refs https://github.com/googleapis/release-please/pull/554

bcoe:
* moves to new [@conventional-commits/parser](https://www.npmjs.com/package/@conventional-commits/parser): eliminating the need for stream-based parsing, simplifying the code, and giving us more flexibility to add our own custom functionality to the parser.
* pulls in latest version of `conventional-changelog-writer` which fixes #601.
* drop `conventional-commits-parser` dependency.

Fixes #601 
Refs #581

### DIFF
--- a/src/conventional-commits.ts
+++ b/src/conventional-commits.ts
@@ -15,7 +15,6 @@
 import chalk = require('chalk');
 import * as semver from 'semver';
 import {ReleaseType} from 'semver';
-import {Readable} from 'stream';
 
 import {checkpoint, CheckpointType} from './util/checkpoint';
 import {Commit} from './graphql-to-commits';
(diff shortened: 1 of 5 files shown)


src/conventional-commits.ts
```diff
@@ -201,78 +180,62 @@ export class ConventionalCommits {
...
+    const parsedCommits = [];
+    for (const commit of this.commits) {
+      try {
+        const parsedCommit = postProcessCommits(
+          toConventionalChangelogFormat(parser(commit.message))
+        ) as CommitWithHash;
+        parsedCommit.hash = commit.sha;
+        parsedCommits.push(parsedCommit);
+      } catch (_err) {
+        // Commit is not in conventional commit format, it does not
```
alexander-fenster, line 192: Actually, this makes it very easy to miss an important commit in the change log if it was not named properly. Maybe we could have a section in the change log listing unparsed commits as is?

PR #684, opened 2021-01-06
feat: add rust releaser & cargo-toml updater

### LINKED ISSUE #683: Add support for Rust crates / updating Cargo.toml (with monorepo support)
I'd love to use `release-please` for some of my Rust repositories, it should be fairly straightforward to bump the version in `Cargo.toml` instead of, say, `package.json`.

In particular, I think it should:

  * not assume the crate is going to be published on crates.io (third-party registries exist, like cloudsmith)
  * work with cargo workspaces (that have a top-level `Cargo.toml` file that list members, and one `Cargo.toml` file per member, in subdirectories)

Opening this issue as per the contributing guidelines, I do have a PR ready for this which I'll file in the next few minutes.

fasterthanlime:
This adds supports for releasing Rust crates, by bumping their version
in `Cargo.toml`, and dependency specs in any of their local dependents.

(Note: `Cargo.toml` edition does not perserve formatting - in Rust,
the `toml_edit` crate allows that, but in npm land, we simply don't have)

It supports two use cases: either the repository contains a single crate
that's also a package, or it's a monorepo: a cargo workspace that has
multiple members.

In the first case, everything is delightfully simple. The singular
`Cargo.toml` has its `package.version` bumped, and we're home free.

In the second case, `release-please` should be invoked separately
for each member of the workspace, with `path`, `packageName` and
`monorepoTags` set.

For example, a Cargo workspace that looks like the following:

```
  - Cargo.toml
  - crates/
    - crate1/
      - Cargo.toml
    - crate2/
      - Cargo.toml
```

Where `crate2` depends on `crate1`, like so:

```
[package]
name = "crate2"
version = "0.1.0"

[dependencies]
crate1 = { version = "0.1.0", path = "../crate1" }
```

Whenever `release-please` is invoked with either:

  * path=crates/crate1, packageName=crate1, monorepoTags=true
  * path=crates/crate2, packageName=crate2, monorepoTags=true

Then we attempt to discover a top-level `Cargo.toml`. If it's a
workspace, we attempt to update the `Cargo.toml` for *all the crates*.

For example, if `crate1` is bumped, its `Cargo.toml` will have its
`package.version` field bumped. For `crate2`, its `Cargo.toml` will have
`dependencies.crate1.version` (or `dev-dependencies`, or
`build-dependencies`) bumped, *if and only if* that dependency spec
has a `path` component.

What does this enable?

  * When developing locally, only `dependencies.x.path` properties are
    used
  * When cargo-publishing, `dependencies.x.path` properties are ignored,
    that's why crate2's `Cargo.toml` needs to be updated, otherwise it would
    immediately fail to build.

Each crate of a monorepo can be published independently, with their own
prefixed tags. There's currently no way to trigger a release of a
specific crate without touching at least one file under its path with a
conventional commit. The behavior of the "commit-as" trailer is
untested.

While adding this functionality, I ran into a limitation of
`ReleasePR.commits`: in the base class, if a `sha` *and* a `path`
is passed, but the `sha` is for a commit that does not touch that
path, it's completely ignored, and *all commits* are returned.

That's why the `Rust` releaser adds its own `commits()` method, which
fetches two sets of commits:

  * All commits (currently, the last 100) for a given path
  * All commits up until `sha` (for any path)

It then takes the intersection of both of those sets, and that gets
passed to `generateChangelogEntry`. Without that workaround, the
generated changelog used to have duplicate entries - it would list
the last 100 conventional commits for every release, no matter what.

---

- [x] Tracking issue: https://github.com/googleapis/release-please/issues/683
- [x] The tests and linter pass
- [x] Coverage increased by ~0.49%
- [ ] I have not updated the docs yet

Fixes #683 🦕


fasterthanlime:
Re docs: I see a bunch of releasers are missing from the README - the README has `node`, `python`, `terraform-module`, and `simple`, but the CLI has:

```
  --release-type          what type of repo is a release being created for?
  [choices: "go-yoshi", "go-yoshi", "java-auth-yoshi", "java-bom", "java-yoshi",
                  "node", "php-yoshi", "python", "ruby-yoshi", "ruby", "simple",
                                   "terraform-module", "rust"] [default: "node"]
```

I don't see much else in the way of docs, am I missing something?

Happy to do whatever is needed to get this to land.

Re: `Rust.commits` - I think a similar fix should be applied to the base `ReleasePR` class, but it seems out of scope for this PR.

bcoe:
@fasterthanlime thank you 👏 I've never worked with rust, so was worried if I attempted to add this functionality I wouldn't get some of the nuances correct.

test/releasers/rust.ts
```diff
@@ -0,0 +1,240 @@
...
+  );
+  const workspaceManifest = readFileSync(
+    resolve(fixturesPath, 'Cargo-workspace.toml'),
+    'utf8'
+  );
+  const graphql = JSON.parse(
+    readFileSync(resolve(fixturesPath, 'commits.json'), 'utf8')
+  );
+
+  let req = nock('https://api.github.com')
```
bcoe, line 47: I was finding nock was getting pretty painful, so in some of the newer unit tests I've been mocking methods on the `github` class instead:

https://github.com/googleapis/release-please/blob/master/test/releasers/terraform-module.ts#L101

I have a slight preference for this, but you've already put the footwork in for nock... so I'm open to using your approach.

What are your thoughts?

src/updaters/cargo-toml.ts
```diff
@@ -0,0 +1,136 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {checkpoint, CheckpointType} from '../util/checkpoint';
+import {Update, UpdateOptions, VersionsMap} from './update';
+import {GitHubFileContents} from '../github';
+import * as TOML from '@iarna/toml';
```
bcoe, line 18: Hey, I used to work with @iarna 😄 I'm glad this is the TOML parser you landed on.

bcoe (changes requested):
No major nits, this looks great to me:

* If you have the cycles, using `sinon` might make future updates easier for people (not a hard blocker for me though).
* mind running `npm run fix`, it will cleanup those quote warnings on the PR.

Thank you for this great contribution.

src/updaters/cargo-toml.ts
fasterthanlime: this is the first I'm hearing of a PR adding a dependency that sparks joy! 😄 

(that toml package is also used in renovate, which I'm also drafting a PR for this week, so at least if it breaks, it'll break for everyone the same way!)

test/releasers/rust.ts
fasterthanlime: I wasn't familiar with `nock` (it was quite the adventure) but I am familiar with `sinon` so this should be quick! I'm on it.

fasterthanlime:
> * mind running `npm run fix`, it will cleanup those quote warnings on the PR.

Ah darn, I had run it before rebasesquashing, but not after. Will do!

fasterthanlime:
Alright!

  * Tests migrated to sinon, which is much cleaner and clearer, I stole all the great comments from `test/releasers/simple.ts`. The `stringifyExpectedChanges` helper is great!
  * Ran `npm run fix` on the PR
  * Added some more notes about limitations in the already-novel-sized commit message

fasterthanlime:
update: Codecov was very unhappy about `Rust.commits()` being not covered at all, so I stubbed `gh.commitsSinceSha` instead, should be good now 🤞

test/releasers/rust.ts
```diff
@@ -0,0 +1,251 @@
...
+    it('creates a release PR for non-monorepo', async () => {
+      const releasePR = new Rust({
+        repoUrl: 'fasterthanlime/rust-test-repo',
+        releaseType: 'rust',
+        packageName: 'crate1',
+        apiUrl: 'https://api.github.com',
+      });
+
+      // Indicates that there are no PRs currently waiting to be released:
+      sandbox
```
bcoe, line 44: thanks for putting in this work, I'm finding this pattern is more readable.

bcoe (approved):

bcoe:
@fasterthanlime to me this seems like a really good starting point, I'm sure we might find and fix a few bugs as we start wiring it up to a few Rust repositories.

There's already an open feature request [here](https://github.com/google-github-actions/release-please-action/issues/163) asking for support, so we can ask them for additional help testing.

Are you using the action as well? I'll release this within the next hour, and we can update the action as well.

test/releasers/rust.ts
fasterthanlime: Agreed, I feel a lot better about those tests now!

fasterthanlime:
Thanks for reviewing and merging this! A starting point is 100% the right way to look at this, I'm very happy this landed early.

Looking forward to finding and fixing bugs as everyone reaches out with their custom setup.

(We are using the action as well, cutting a release of it as well would be much appreciated).

Feel free to ping me whenever rust related issues come up if you want — I can't guarantee response times, but I'm happy to help out where I can, plus I'm somewhat familiar with the codebase now ✨

bcoe:
@fasterthanlime please let me know when you have a chance to test this out, I've released the action.

fasterthanlime:
@bcoe Just tested it, everything appears to be working as intended for me!

joeldodge79:
@fasterthanlime I'm working on something similar for node to support npm/yarn workspaces. I thought I'd bounce my ideas off you in case you have any input:

1. Separately but related I have the need for a Release PR that includes updates/changelogs for multiple packages. A PR for this "aggregate" configuration is coming soon but the main idea is that we'll add a `--packages` flag/input that specifies the necessary releaser config per package. The aggregate releaser will gather all the commits, split them by package-path, pass them to individual releaser instances whose job it will be only to reply with proposed updates, and create a single PR with the result. This feature will apply to homogeneous as well as heterogeneous monorepos (within the bounds of currently implemented releasers).

2.`node` npm/yarn workspace dependency management: for any `--packages` of type `node` for which there is corresponding npm/yarn workspaces configuration: update each package's workspace-dependencies to the latest version. I wrote a [proof of concept](https://github.com/googleapis/release-please/pull/726/files/e23083b601540b874869a11a1e4ed1119478dcf4#diff-82396a2a416ee095585d98d1b81352a96eb6e50a66071453d10923dc8ea43399R17) using lerna libraries. I haven't connected it yet but somehow the aggregate releaser gets all the package.json contents for each package from the individual releaser (which has already bumped the top level version) and passes them through this lerna wrapper before including them in the PR.

I have a question about your approach here using the currently available "individual" configuration. When updating workspace dependencies do you have to be careful about what order you merge in the separate Release PRs? In your example, I could be making changes to `crate1` and separately making changes to `crate2` which depend on the `crate1` changes. I now have two separate Release PRs.

PR for crate2
- crate2.version bump
PR for crate1 
- crate1.version bump
- crate2.depencency-on-crate1 version bump

In this simple case I know that I need to merge the crate1 PR first and regenerate crate2's PR to pick up the depencency-on-crate1 version bump before merging it. But for more complex dependency graphs that could be complicated to figure out.
 
I'll keep you posted on the work in case you're interested in adapting the `rust` releaser to fit this aggregate model.

PR #694, opened 2021-01-11
feat: add addFilesByExtension method & allow leading slash in path

### LINKED ISSUE #692: Adding a path that starts with / causes an error
## Problem

Using a path starting with `/` causes an error when interacting with GitHub API - e.g. `ReleasePR.addPath('/esy.json')`

I already have this fixed. I'm just creating this to have an issue to link to in my PR.

#### Environment details

  - OS: Mac
  - Node.js version: 14.15.1
  - npm version: 6.14.8
  - `release-please` version: 9.0.0

#### Steps to reproduce

  1. In a releaser, try adding a path starting with `/`
  2. See the error it causes when interacting with github.

Thanks!

### LINKED ISSUE #693: GitHub API should provide method for fetching files by extension

While building the OCaml releaser, I found that I need to query GitHub for files ending with `.opam` and `.json`, so I need a way to fetch all those files.

I already added a method called `findFilesByExtension` that acts like `findFilesByFilename`, but only takes an extension to search for.

I considered requiring a user to specify the files they want to update, but there doesn't seem to be a great way to add those config options.

phated:

Fixes #692 Fixes #693 🦕


src/release-pr.ts
```diff
@@ -329,7 +329,11 @@ export class ReleasePR {
 
   static addPathStatic(file: string, path?: string) {
     if (path === undefined) {
-      return file;
+      if (file.startsWith('/')) {
```
bcoe, line 332: Does it make sense to also check for `\`?

src/github.ts
```diff
@@ -1034,6 +1034,31 @@ export class GitHub {
...
     });
   }
+
+  async findFilesByExtension(
+    extension: string,
+    prefix?: string
+  ): Promise<string[]> {
+    let q = `extension:${extension}+repo:${this.owner}/${this.repo}`;
+    if (prefix) {
+      prefix = prefix.replace(/[/\\]$/, '');
```
bcoe, line 1044: IF we find ourselves doing this too many more times, it might be worth pulling into a helper, `removeLeadingSlash`?

Could just be in the `github.js` class I suppose.

bcoe (changes requested):
Only real nit, is that we should be consistent about removing the leading `\`.

I'm fine with not abstracting the removing of a leading slash into a helper yet.

----

Disclaimer I always give folks using the search based API, GitHub sometimes doesn't immediately index all files. The approach works, but I think can have some hiccups when a new file is initially added (_they eventually correct themselves._)

src/release-pr.ts
phated: Sure! I was originally getting these back from the github API, which always returns `/`, but I could see someone trying to do a path normalization on windows.

bcoe (approved):
Awesome, thanks for the contribution 🎉

PR #697, opened 2021-01-11
feat: add ocaml releaser and esy/opam updaters

### LINKED ISSUE #695: OCaml Releaser
I want to use ReleasePlease on an OCaml project. There are 2 packages managers that they often support, Opam & Esy. Each uses a different package manifest (often `something.opam` and `esy.json`).

A releaser for OCaml that has both package manifests supported. I have work on releaser done and ran it against https://github.com/grain-lang/binaryen.ml/pull/48 - I'll submit the PR shortly.

N/A

phated:
## Context

This adds an initial implementation of an OCaml releaser that supports both `esy.json` and `opam` files. Those package managers are very flexible with filenames, so the releaser searches for `*.json` and `*.opam` files to update within the repository.

I still need to do research/work around getting this to work in a monorepo, but getting a good baseline for a project like Binaryen.ml (see https://github.com/grain-lang/binaryen.ml/pull/48) is my first milestone.

## Template

Fixes #695  🦕


phated:
I don't understand the Windows failure? Is it line-endings or something?

bcoe:
> I don't understand the Windows failure? Is it line-endings or something?

@phated  I'm going to bet it's line encoding related, I can try to take a look tomorrow at some point if nothing jumps out at you.

src/releasers/ocaml.ts
```diff
@@ -0,0 +1,128 @@
...
+
+const CHANGELOG_SECTIONS = [
+  {type: 'feat', section: 'Features'},
+  {type: 'fix', section: 'Bug Fixes'},
+  {type: 'perf', section: 'Performance Improvements'},
+  {type: 'revert', section: 'Reverts'},
+  {type: 'docs', section: 'Documentation'},
+  {type: 'chore', section: 'Miscellaneous Chores'},
+  {type: 'refactor', section: 'Code Refactoring'},
+  {type: 'test', section: 'Tests'},
```
bcoe, line 39: Is the goal here to show additional categories of PRs in CHANGELOG, do you think wanting to show these additional sections will be common in the OCaml community?

My concern is that showing every docs change, chore, etc., can get a little noisy.

src/updaters/ocaml/esy-json.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  constructor(options: UpdateOptions) {
+    this.create = false;
+    this.path = options.path;
+    this.changelogEntry = options.changelogEntry;
+    this.version = options.version;
+    this.packageName = options.packageName;
+  }
+
+  updateContent(content: string): string {
```
bcoe, line 36: If I'm understanding correctly, we're going to attempt to update every `.json` file in the repository, even though some might not be `esy` files?

Could we add a helper that attempts to detect whether it's the appropriate file type, by looking at some well known fields.

bcoe (changes requested):
Left some initial nits, in general this is looking really good to me.

src/releasers/ocaml.ts
phated: The common convention in the ocaml community is to show all of this. Ideally, these would be configurable by consumers and just provide a sensible default.

src/updaters/ocaml/esy-json.ts
phated: Yeah, the weirdness with esy is that you can have different json files for different sanboxes, those might contain an `esy` section, `overrides` section, or none of them (and just look like a package.json file...). Maybe we just check for the `version` key and update it? Even if an esy package contains an `esy.json` and `package.json`, it will be common that they both need to be updated anyway.

Thoughts?

bcoe: If you think the version will be okay, I think that's a good start (we'll at least get things working for your needs).

My only concern is as other folks with fixtures, etc., start to use your release strategy -- maybe at that point we can introduce an option to pass a list of files rather than using search?

README.md
```diff
@@ -84,6 +84,8 @@ Release Please automates releases for the following flavors of repositories:
 | node              | [A Node.js repository, with a package.json and CHANGELOG.md](https://github.com/yargs/yargs) |
 | python            | [A Python repository, with a setup.py, setup.cfg, and CHANGELOG.md](https://github.com/googleapis/python-storage) |
 | terraform-module  | [A terraform module, with a version in the README.md, and a CHANGELOG.md](https://github.com/terraform-google-modules/terraform-google-project-factory) |
+| rust              | A Rust repository, with a Cargo.toml (either as a crate or workspace) and a CHANGELOG.md |
```
bcoe, line 87: thanks for adding this.

bcoe (approved):
Awesome  🎉

PR #698, opened 2021-01-12
fix(github-release): increasae page size finding release

bcoe:
In monorepos there's enough traffic on PRs that looking at the last 25 didn't catch a release.


chingor13 (approved):
This should probably use search with pagination.

alexander-fenster (approved):

PR #703, opened 2021-01-13
feat(github-release): create a draft release

joeldodge79:
Added a "draft" option which defaults to false. If true, a draft release
is created (release.draft == true and no tag is created).`tag_name` and
`target_commitish` are still associated with the release for future tag
creation when draft is changed to false. Using the web UI this is
prepopulated on the release form. Via the API it uses the previously
stored values unless the call explicitly changes them.


joeldodge79:
waiting for https://github.com/googleapis/release-please/pull/702 to land before publishing this PR for review

bcoe:
@joeldodge79 went ahead and merged the snapshot update.

src/bin/release-please.ts
```diff
@@ -126,6 +126,14 @@ const argv = yargs
           describe: 'include library name in tags and release branches',
           type: 'boolean',
           default: false,
+        })
+        .option('draft', {
```
joeldodge79, line 130: @bcoe how do you feel about this cli option name? I think the release-please-action input should be `create-release-as-draft` as you suggested but that felt repetitive here. However, I can also see the argument for consistency.

bcoe: Why don't we add [an alias](https://github.com/yargs/yargs/blob/master/docs/api.md#optionskey-opt), `create-release-as-draft`, but keep the shorthand too.

In the GitHub Action, we can document the long form arguments, but for folks experimenting with the CLI, save a few key strokes.

bcoe (changes requested):
My only tiny nit would be, why don't we add long-form aliases for keys that have shorthands.

By adding these longer form options, perhaps it will make it clearer in the action which options are for Release PRs, and which are for GitHub Releases.

bcoe:
_Perhaps the two prefixes we could use are **pr-foo**, when the option is specific to PRs, and **release-foo** when the option is specific to GitHub releases.

src/bin/release-please.ts
joeldodge79: got a little confused about this suggestion in combination with your **{pr,release}-foo** prefixing suggestions below (which I like). Assuming that suggestion was for the cli (as opposed to the action) should I

- [ ] add an alias `create-release-as-draft`
- [ ] add an alias `release-create-as-draft`
- [ ] add aliases for options that only pertain to one command: (i.e. `pr-default-branch`, `pr-version-file`, `pr-default-branch`, `pr-fork`, `pr-snapshot`, `pr-last-package-version`, `release-release-type`, `release-changelog-path`)
- [ ] change the option name to `release-draft`
- [ ] keep the option name `draft`

bcoe: Let's keep the option `draft` 👍 

If you're so inclined, add `release-create-as-draft`.

bcoe (approved):
I'll leave adding an alias up to you 👍 If you think it would be better to have a longer name for the action, let's add it.

test/github-release.ts
```diff
@@ -68,7 +68,69 @@ describe('GitHubRelease', () => {
...
+        packageName: 'foo',
+        apiUrl: 'https://api.github.com',
+        draft: true,
+      });
+      const requests = nock('https://api.github.com')
+        // check for default branch
+        .get('/repos/googleapis/foo')
+        .reply(200, repoInfo)
+        .get(
+          '/repos/googleapis/foo/pulls?state=closed&per_page=50&sort=merged_at&direction=desc'
```
bcoe, line 107:
```suggestion
          '/repos/googleapis/foo/pulls?state=closed&per_page=50&sort=updated&direction=desc'
```

src/bin/release-please.ts
joeldodge79: I'm getting a type error when adding the `alias` key. not sure why but maybe I'll leave it off for now.

```
src/bin/release-please.ts:137:11 - error TS2345: Argument of type '{ describe: string; type: string; default: false; alias: string; }' is not assignable to parameter of type 'YargsOptions'.
  Object literal may only specify known properties, and 'alias' does not exist in type 'YargsOptions'.

137           alias: 'release-create-as-draft',
```

PR #705, opened 2021-01-14
feat(rust): Update Cargo.lock, preserve formatting in Cargo.{toml,lock}

### LINKED ISSUE #704: Rust: update Cargo.lock and attempt formatting-preserving Cargo.toml update
I've been using the initial Rust support ( https://github.com/googleapis/release-please/pull/684 ) for a week now and I want to come up with some improvements:

## Cargo.toml

The editing of `Cargo.toml` that is *not* formatting-preserving is more annoying than anticipated.

The initial PR is noisy (as expected) and the subsequent PRs aren't as noisy (also as expected), which is what is is, but acceptable.

What's less ideal is that it also strips all TOML comments from the file, so comments like "revert to upstream once PR #ZZZ lands" get lost in translation.

## Cargo.lock

I initially considered `Cargo.lock` too hard to update, since it would require running a command like `cargo check`, which requires a fully-functioning Rust toolchain, potentially with a C compiler, an alternative linker (like LLVM), etc.

I also didn't want to edit it "by hand" (without running `cargo`) since `[[package]]` entries contain checksums.

As it turns out, not updating `Cargo.lock` prevents `cargo publish` from working after a release has been created, since the first thing it does is updated the `Cargo.lock`. It then fails because there are "dirty" (uncommitted) files in the repository.

This is only an issue for binary crates that are *not* in a monorepo, because:

  * library crates have their `Cargo.lock` gitignored
  * monorepos have their `Cargo.lock` at the top level, not in `crates/mycrate`

Nevertheless, I want to try and address both of these - I'll open a relevant PR soon.

fasterthanlime:
The initial Rust support worked, but it didn't preserve formatting in `Cargo.toml`, because it parsed the whole file using `@iarna/toml` and serialized it again, turning inputs like these:

```toml
[dependencies]
# revert to upstream once https://github.com/seanmonstar/warp/pull/753 lands	
warp = { version = "0.2.5", registry = "netlify-trove" }
```

Into this:

```toml
  [dependencies.warp]
  version = "0.2.5"
  registry = "netlify-trove"
```

(note: the spacing is part of the output).

The problem is not so much that it reformats the whole file (also that's annoying too), but that an important comment was lost in translation.

Additionally, with the initial support, `Cargo.lock` was not updated. This resulted in follow-up commits (whenever developers ran `cargo check`, `cargo build`, `cargo test` etc. locally) that had "`Cargo.lock` update noise", but also, it prevented a `cargo publish` step from ever working on CI for non-monorepos that were binary crates (and thus have their `Cargo.lock` file committed).

This PR builds on `@iarna/toml`'s low-level API: it introduces a `TaggedTOMLParser`, which returns objects instead of string values. Those objects contain the location of the string in the original .toml file. A single function, `replaceTomlString`, is exported, which parses a TOML payload, attempts to follow a path into the "object tree", replaces the string, and validates that the result is valid TOML.

Both the `CargoToml` and `CargoLock` updaters now use that function. Additionally, the code was reorganized a little, some comments were added, common parts were extracted to their own files.

The `rust` releaser changes are fairly simple, and its existing tests are mostly untouched (modulo updating the snapshots, since it now preserves formatting!), I'm going to add some more tests to make sure it works both with `Cargo.lock` present and absent, both in a monorepo and non-monorepo scenario.

Fixes #704 🦕


fasterthanlime:
Another note: this also includes some partial typings for `@iarna/toml`'s low-level interface, which is not covered by the `@types/iarna__toml` package unfortunately.

Since this is the first "typing extensions" for this repo, I had to touch `tsconfig.json`, here's the relevant docs: https://www.typescriptlang.org/tsconfig#typeRoots

bcoe:
> Another note: this also includes some partial typings for @iarna/toml's low-level interface, which is not covered by the @types/iarna__toml package unfortunately.

@fasterthanlime seems reasonable as a stop gap, but could we open a tracking ticket on DefinitelyTyped?

---

Let me know when you've added all the tests you'd like and I'll review.

fasterthanlime:
> @fasterthanlime seems reasonable as a stop gap, but could we open a tracking ticket on DefinitelyTyped?

✅  https://github.com/DefinitelyTyped/DefinitelyTyped/issues/50618

fasterthanlime:
codecov is happy, I'm mostly happy - still going to add some unit tests to `toml-edit`, but I think you're good to start a proper review @bcoe 🎉

fasterthanlime:
So, the windows tests are failing in CI, I got curious and ran them locally (Windows 10, node v15.5.0) and they run just fine here. It's a mystery! 🔎

bcoe:
@fasterthanlime I think you're probably getting bitten by line endings, I've had to add `replace('\r\n', '\n');` in a few places when taking snapshots.

test/updaters/cargo-lock.ts
```diff
@@ -0,0 +1,50 @@
...
+        versions,
+        packageName: 'delf',
+      });
+      const newContent = cargoToml.updateContent(oldContent);
+      const pkg = parseCargoLockfile(newContent).package![4];
+      expect(pkg).to.deep.include({
+        name: 'delf',
+        version: '0.2.0',
+      });
+      snapshot(newContent);
```
bcoe, line 47:
```suggestion
      snapshot(newContent.replace(/\r\n/g, '\n'));
```

tsconfig.json
```diff
@@ -4,8 +4,10 @@
     "lib": ["es2018", "dom"],
     "rootDir": ".",
     "outDir": "build",
+    "typeRoots": ["./typings", "./node_modules/@types"]
```
bcoe, line 7: TIL, I've never gotten this fancy with providing custom typings.

bcoe (approved):
This is looking good to me, and I'm comfortable landing the way it is.

The only real concern I have is if we bump into any significant bugs with the TOML parser, and need to keep fleshing out that test suite -- it might eventually make sense for it to be its own repo.

bcoe:
@fasterthanlime went ahead and addressed the newline issues for you, IMO this is ready to go. Shall I get it released for you tomorrow?

fasterthanlime:
@bcoe sounds good to me! Thanks for addressing the newline issue, I keep forgetting Git can be configured to check out with CRLF 🙈

I share your general concern re "parsers vs real world input", although 1) TOML is a relatively simple format compared to, say, YAML 2) I'm happy I found a solution based on that same parser, and not something entirely custom or a pile of regexps.

fasterthanlime:
Just thought of some more good test cases for the toml-edit part (mostly re: quoted keys and non-basic strings), maybe hold off on the release until I add them in ⏳

fasterthanlime:
Glad I took the time to look at some of the funkier TOML out there, all kinds of strings are now supported, I just want to add support for platform-specific dependencies before this lands: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies

fasterthanlime:
Platform-specific dependency support is in, adding a few more tests for good measure... ⌛

fasterthanlime:
All unit tests are here, running this against a sample repo to be 101% sure.

fasterthanlime:
Ran successfully on a couple repositories - a complex private one, and this simple public one: https://github.com/fasterthanlime/release-please-sample/pull/1

@bcoe This PR is ready for prime time! 🚀🐿

bcoe:
@fasterthanlime thanks for the contribution 👍 I might just need to learn rust.

fasterthanlime:
@bcoe thanks for maintaining this repo! Contributing has been a real pleasure ☺️

If you do find some time to learn some Rust, [I have quite a few articles](https://fasterthanli.me/tags/rust) you may enjoy ✨

PR #708, opened 2021-01-15
fix(pagination): when tagging release, use updated, vs., created

bcoe:
There's a danger that someone comments on an old merged release PR, brining it earlier in chronological order, this would mean that we attempt to re-release the same version given how last release is calculated.

This is not a problem when tagging a release, which only occurs once after a PR is merged.


feywind (approved):
To paraphrase Asimov, we had the `answer that satisfied` but now we have the `answer that was true` :D

PR #711, opened 2021-01-21
fix(go): support processing footers in gapics

codyoss:
Because we need to filter the commits found in the footer of a
commit we needed a hook to filter commits for the changelog. This
can't be done in the releaser as it only has the concept of raw
commits.Added the idea of supplying an optional commit filter to
ConventionalCommits.


src/conventional-commits.ts
```diff
@@ -71,7 +72,10 @@ interface Note {
   text: string;
 }
 
-function getParsedCommits(commits: Commit[]): CommitWithHash[] {
+function getParsedCommits(
+  commits: Commit[],
+  commitFilter?: (c: ConventionalChangelogCommit) => boolean
```
bcoe, line 77: this is a clever approach.

src/conventional-commits.ts
```diff
@@ -81,6 +85,9 @@ function getParsedCommits(commits: Commit[]): CommitWithHash[] {
         const commitWithHash = postProcessCommits(
           parsedCommit
         ) as CommitWithHash;
+        if (commitFilter?.(parsedCommit)) {
```
bcoe, line 88: what if we default the `commitFilter` to a filter that always returns `true`, the current behavior, then we don't need to use this optional call?

bcoe (changes requested):
I like the approach you've come up with 👍 one thought, perhaps we could just default the filter to: `() => true`, then we don't need to do the optional call logic.

src/conventional-commits.ts
codyoss: Defaulted to false, as that will allow a commit to be processed.

bcoe (approved):

PR #712, opened 2021-01-21
feat: add add GitHub helpers for findFilesByFilenameAndRef and findFilesByExtensionAndRef

### LINKED ISSUE #656: Fail to update pom file versions when proposing release PR
Below proposed release PR is missing pom file updates:
https://github.com/googleapis/java-bigquery/pull/972/files

### LINKED ISSUE #710: File search only considers the default branch
[`findFilesByFilename`](https://github.com/googleapis/release-please/blob/9ae5620274836acd39662ef0fb55502d06585423/src/github.ts#L1018-L1040) and [`findFilesByExtension`](https://github.com/googleapis/release-please/blob/9ae5620274836acd39662ef0fb55502d06585423/src/github.ts#L1042-L1065) use the code search API which only uses the default branch (and also may not work on forked repos).

This prevents releasers that use this functionality from working on non-default branches.

chingor13:
Fixes #656 - using the git tree API should be instantaneous and not reliant on the search API.

Fixes #710 - provides extra helpers for finding files by filename/extension within a branch.

Refactors findFiles to use the tree API for listing files.


src/github.ts
```diff
@@ -1015,52 +1015,151 @@ export class GitHub {
...
+          path.endsWith(filename) &&
+          // match the prefix if provided
+          (!prefix || path.startsWith(prefix))
+        );
+      })
+      .map(file => {
+        let path = file.path!;
+        // strip the prefix if provided
+        if (prefix) {
+          const pfix = new RegExp(`^${prefix}[/\\\\]`);
```
bcoe, line 1063: this is elegant, I like the approach.

src/github.ts
```diff
@@ -1015,52 +1015,151 @@ export class GitHub {
...
+   *
+   * @param extension The file extension used to filter results.
+   *   Example: `js`, `java`
+   * @param prefix Optional path prefix used to filter results
+   * @param ref Optional git reference. Defaults to the default
+   *   branch's HEAD.
+   */
+  async findFilesByExtensionAndRef(
     extension: string,
+    ref: string,
```
bcoe, line 1107: Should ref be optional, based on the description of the method?

test/releasers/java-yoshi.ts
```diff
@@ -734,4 +664,162 @@ describe('JavaYoshi', () => {
...
+    );
+    const pomContents = readFileSync(resolve(fixturesPath, 'pom.xml'), 'utf8');
+    const googleUtilsContent = readFileSync(
+      resolve(fixturesPath, 'GoogleUtils.java'),
+      'utf8'
+    ).replace(/\r\n/g, '\n');
+    const graphql = JSON.parse(
+      readFileSync(resolve(fixturesPath, 'commits-yoshi-java.json'), 'utf8')
+    );
+    const req = nock('https://api.github.com')
```
bcoe, line 698: I'm wondering if it would be worth moving some of this logic to `sinon`, like [here](https://github.com/googleapis/release-please/blob/master/test/releasers/rust.ts#L147) I've been finding that it feels a bit more maintainable than enumerating all of the GitHub endpoints we'll be interacting with.

bcoe (changes requested):
The comment re: using sinon for mocking doesn't need to be a blocker, but I wonder if it makes the codebase more maintainable long term.

I've found people really wrestle with `nock`, and it might be better to only use it for select things, where we want to be darn sure we've exercised code paths that rely on it.

src/github.ts
chingor13: I originally had it optional as an overload of `fileFilesByExtension(extension, prefix?, ref?)`, but then opted to make it a separate explicit method. I'll fix the docs to not make it optional.

chingor13: Done

test/releasers/java-yoshi.ts
chingor13: I like this a lot. Let me do it in a separate, follow-up PR to avoid muddying up this PR.

bcoe (approved):

PR #714, opened 2021-01-21
fix: use the GitHub release label to help find the correct PR

### LINKED ISSUE #713: Rebasing the main branch makes release-please get confused
I work on a small team that is moving quick and we sometimes make mistakes in our git history/commits. If we rebase or update the main branch and push to correct issues, the release-please PR that is generated contains all changes since the start of the project.

#### Steps to reproduce

  1. Make a mistake and push it to the main branch
  2. Check that the release-please PR has the correct list of changes
  3. Fix the mistake with rebasing etc
  4. Push the update to the main branch
  5. Check the release PR again

If the release PR commit hashes change, release-please cannot find a commit hash that matches. Since the release PRs are mostly immutable once they are merged, it is not possible to ignore a previous (potentially failed) release PR. One possible solution is to use the assumption that the github-release step is going to tag released PRs.

safarmer:
Looks for PRs with the `autorelease: tagged` label when considering release PRs. This helps prevent
creating release PRs that include every change in the repository as opposed to changes since the
last release when the main branch has been rebased or has had other issues.

The change to `findMergedReleasePR` means it is possible to manually update the tags on merged PRs
to either have them included or excluded from the list of candidates used to find the last release.

This also changes test imports to use `resolve` so that they can be run more easily from IDEs and
tools such as wallaby.js.

Based on my testing, this looks like a safe change to make, but perhaps it should be placed behind a
command line/feature flag?

Fixes #713

### DIFF
--- a/src/github.ts
+++ b/src/github.ts
@@ -66,6 +66,7 @@ import {
   PREdge,
 } from './graphql-to-commits';
 import {Update} from './updaters/update';
+import {resolve} from 'path';
 
 // Short explanation of this regex:
 // - skip the owner tag (e.g. googleapis)
@@ -73,6 +74,8 @@ import {Update} from './updaters/update';
 // - take everything else
 // This includes the tag to handle monorepos.
 const VERSION_FROM_BRANCH_RE = /^.*:release-?([\w-.]*)-(v[0-9].*)$/;
+
+export const GITHUB_RELEASE_LABEL = 'autorelease: tagged';
 
 export interface OctokitAPIs {
   graphql: Function;
@@ -153,7 +156,7 @@ export class GitHub {
         baseUrl: this.apiUrl,
         headers: {
           'user-agent': `release-please/${
-            require('../../package.json').version
+            require(resolve('package.json')).version
           }`,
           // some proxies do not require the token prefix.
           Authorization: `${this.proxyKey ? '' : 'token '}${this.token}`,
@@ -457,7 +460,12 @@ export class GitHub {
     prefix?: string,
     preRelease = false
   ): Promise<GitHubTag | undefined> {
-    const pull = await this.findMergedReleasePR([], 100, prefix, preRelease);
+    const pull = await this.findMergedReleasePR(
+      [GITHUB_RELEASE_LABEL],
+      100,
+      prefix,
+      preRelease
+    );
     if (!pull) return await this.latestTagFallback(prefix, preRelease);
     const tag = {
       name: `v${pull.version}`,
--- a/src/github-release.ts
+++ b/src/github-release.ts
@@ -16,12 +16,11 @@ import chalk = require('chalk');
 import {checkpoint, CheckpointType} from './util/checkpoint';
 import {packageBranchPrefix} from './util/package-branch-prefix';
 import {ReleasePRFactory} from './release-pr-factory';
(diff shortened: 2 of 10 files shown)


safarmer:
I can also add the wallaby.js config I have that seems to be working if that would be useful for anyone.

bcoe:
@safarmer thank you for the contribution 👏 

You've caught us right in the middle of a refactor that might remove this logic entirely, instead looking for prior `tags` using GraphQL -- which I think would similarly solve your problem.

Regarding `wallaby.js` where does its configuration live? we're mainly vscode users here, but if it's in a reasonable home like `.editors`, I'm open to adding the config.

src/github.ts
```diff
@@ -153,7 +156,7 @@ export class GitHub {
         baseUrl: this.apiUrl,
         headers: {
           'user-agent': `release-please/${
-            require('../../package.json').version
+            require(resolve('package.json')).version
```
joeldodge79, line 159: oh sweet - does this mean when I setup vscode debugging per [this article](https://medium.com/@benlesh/debugging-typescript-mocha-tests-with-vscode-89310051531) I won't have to go manually change these paths?

joeldodge79 (commented):
another thought is: switch to using a stable release branch name (e.g. `release-please-<packagename>`) and additionally passing in the `head` parameter (using the {owner:branch-name} format) to the `/repos/pulls` API which should basically find exactly the one PR we want.

I'm curious to know if you think that would handle whatever situation prompted proposing this refinement? (something about default branch got rebased/force pushed or something?)

src/github.ts
safarmer: It would seem so as I was using the VSCode debugger to step through the code. Would it be useful for me to split this change into another PR?

safarmer:
> @safarmer thank you for the contribution 
> 
> You've caught us right in the middle of a refactor that might remove this logic entirely, instead looking for prior `tags` using GraphQL -- which I think would similarly solve your problem.
> 
> Regarding `wallaby.js` where does its configuration live? we're mainly vscode users here, but if it's in a reasonable home like `.editors`, I'm open to adding the config.

The config lives in $root/wallaby.js. The tool runs inside VSCode/WebStorm/IntelliJ giving test feedback. It should not interfere with editor setup.

safarmer:
@bcoe @joeldodge79 I have created a new PR (#715) with the resolve change and a wallaby.js config file since they still seem like useful changes.

@jdpedrie re your suggestion: I cannot think of any reason that won't work off the top of my head. I would need to make sure the stable release branch was protected so as not to be auto-deleted on merge.

I have added some more details with an example repo in the issue #713

src/github.ts
bcoe: @joeldodge79 just landed @safarmer's other PR with the resolve logic, I think we could then just add `tsnode` as a dev dependency, and add an alternate test script (like `test-vs`) for the benefit of VS Code's debugger.

I'd be fine with us checking in configuration in the `.vscode` folder.

bcoe:
@safarmer an update, I think @chingor13's work here can address #715, it proposes moving towards a stable branch name when finding release PRs.

bcoe:
@safarmer perhaps you can confirm against this branch that it does the trick for you? (or can do the trick for you).

PR #720, opened 2021-01-26
feat!: support stable release branch names

### LINKED ISSUE #718: Put releasing branch name in commit title for branch releases
To make it clearer which branch is getting released

### LINKED ISSUE #719: release branch name collisions when releasing from multiple branches
It's possible when releasing from multiple feature branches that release-please tries to use the same version. In this case, release-please would stomp on the other feature branches' release versions.

We probably need to construct a release branch naming structure such that they can never collide across branches.

chingor13:
Fixes #718 
Fixes #719 

* Adds support for building/parsing release branch names (language agnostic feature, but the languages choose an implementation)
* Moves GitHub release/tag logic into the individual releasers and will allow us to override release behavior on a language-by-language basis.
* Moves release note functionality into separate module (language agnostic feature)
* Migrates java-yoshi releaser to use stable branch names (all other languages should maintain their exact same behavior)

feat: add `GitHub#findMergedPullRequest` to filter through merged pull requests with pagination

BREAKING CHANGE:
removed per page parameter from `GitHub#findMergedReleasePR` and moved some internal helpers


src/util/branch-name.ts
```diff
@@ -0,0 +1,115 @@
...
+    if (match) {
+      this.targetBranch = match[1];
+    }
+  }
+  toString(): string {
+    return `release-please/branches/${this.targetBranch}`;
+  }
+}
+
+const COMPONENT_PATTERN = /^release-please\/branches\/([^/]+)\/components\/([^/]+)$/;
```
bcoe, line 99: you can use named capture groups to cleanup code a bit:

```
?<blerg>
```

bcoe (commented):
looking great to me so far, I like that you've broken work out the branch handling into its own helper.

src/github-release.ts
```diff
@@ -189,27 +173,12 @@ export class GitHubRelease {
...
+      defaultBranch: this.defaultBranch,
       owner,
       repo,
       apiUrl: this.apiUrl,
       proxyKey: this.proxyKey,
       octokitAPIs,
     });
   }
-
-  static extractLatestReleaseNotes(
```
chingor13, line 200: This is technically breaking. It was moved to a helper in `utils`

src/releasers/go-yoshi.ts
```diff
@@ -183,7 +183,7 @@ export class GoYoshi extends ReleasePR {
     return '0.1.0';
   }
 
-  static tagSeparator(): string {
+  tagSeparator(): string {
```
chingor13, line 186: This is technically breaking.

src/releasers/ruby.ts
```diff
@@ -108,7 +108,7 @@ export class Ruby extends ReleasePR {
     });
   }
 
-  static tagSeparator(): string {
+  tagSeparator(): string {
```
chingor13, line 111: This is technically breaking

src/util/release-notes.ts
```diff
@@ -12,19 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// map from a packageName to the prefix used in release branch creation.
-export function packageBranchPrefix(packageName: string, releaseType?: string) {
```
chingor13, line 16: This is technically breaking

src/release-pr.ts
```diff
@@ -198,7 +228,7 @@ export class ReleasePR {
     return Promise.resolve(undefined);
   }
 
-  static tagSeparator(): string {
+  tagSeparator(): string {
```
chingor13, line 231: This is technically breaking

test/github-release.ts
```diff
@@ -291,61 +273,56 @@ describe('GitHubRelease', () => {
...
+        .resolves();
+
       const created = await release.createRelease();
-      strictEqual(created!.tag_name, 'v1.0.3');
-      requests.done();
+      expect(created).to.not.be.undefined;
+      strictEqual(created!.tag_name, 'foo-v1.0.3');
     });
 
     it('attempts to guess package name for release', async () => {
```
bcoe, line 328: I might make this test case more generic, something like:

```
it('falls back to lookupPackageName() if no package name provided')
```

test/github-release.ts
```diff
@@ -451,72 +450,161 @@ describe('GitHubRelease', () => {
...
-        resolve(fixturesPath, './CHANGELOG-old-new.md'),
-        'utf8'
-      ).replace(/\r\n/g, '\n');
-      const latestReleaseNotes = GitHubRelease.extractLatestReleaseNotes(
-        changelogContent,
-        'v1.0.0'
-      );
-      snapshot(latestReleaseNotes);
-    });
+    it('parses version from PR title (detectReleaseVersionFromTitle impl: base)', async () => {
```
bcoe, line 455: This seems like a nice simplification over some other approaches we've discussed, e.g., loading the version form `package.json` or `setup.py`.

src/release-pr.ts
```diff
@@ -340,4 +406,111 @@ export class ReleasePR {
...
+  // inferred from the release PR head branch)
+  protected async detectReleaseVersionFromCode(): Promise<string | undefined> {
+    return undefined;
+  }
+
+  private async detectReleaseVersion(
+    mergedPR: MergedGitHubPR,
+    branchName: BranchName | undefined
+  ): Promise<string | undefined> {
+    // try from branch name
```
bcoe, line 422: I like where you've landed with this functionality. Perhaps we could add one test that explicitly calls out order of operation, i.e., "it('uses version number in branch name over version number in title')".

src/release-pr.ts
bcoe: I think it's okay given the dependency counts on npm to safely move this.

src/releasers/java-yoshi.ts
```diff
@@ -287,4 +288,46 @@ export class JavaYoshi extends ReleasePR {
     }
     return newVersions;
   }
+
+  // Begin release configuration
+
+  // Override this method to use static branch names
+  // If you modify this, you must ensure that the releaser can parse the tag version
+  // from the pull request.
+  protected async buildBranchName(
```
bcoe, line 297: Am I following that each language would opt in to this new functionality? Is the reasoning to allow for a gradual upgrade and testing process?

src/github.ts
```diff
@@ -547,6 +582,59 @@ export class GitHub {
...
+        labels: associatedPullRequest.labels.nodes.map(label => label.name),
+        title: associatedPullRequest.title,
+        body: associatedPullRequest.body,
+      };
+    });
+  }
+
   // The default matcher will rule out pre-releases.
   // TODO: make this handle more than 100 results using async iterator.
   async findMergedReleasePR(
```
bcoe, line 640: Is this dead code now that you've pulled `findMergedRelease` into `release-pr`?

src/releasers/node.ts
```diff
@@ -39,9 +38,7 @@ export class Node extends ReleasePR {
     if (pkg.name) this.packageName = pkg.name;
 
     const latestTag: GitHubTag | undefined = await this.gh.latestTag(
```
bcoe, line 40: In the cases of releasers with a stable branch name, should `latestTag` also be looking up the stable name first, and falling back to iterating over `findMergedReleasePR`?

src/releasers/node.ts
```diff
@@ -39,9 +38,7 @@ export class Node extends ReleasePR {
...
-      this.monorepoTags
-        ? `${packageBranchPrefix(this.packageName, 'node')}-`
-        : undefined
+      this.monorepoTags ? `${this.packagePrefix}-` : undefined
```
bcoe, line 41: Wasn't immediately following why we can drop the helper `packageBranchPrefix` here.

Also, if the goal is hopefully to move most the releasers over to a stable branch name, I wonder if the idea of a `prefix` makes sense, should we just be passing in the full branch name?

bcoe (changes requested):
Left some initial comments and suggestions.

I like the direction things are going in, a couple things that jumped out at me:

1. it seems like we'd want the lookup of `latestTag` and "commits since last release" to handle the stable branch names too -- unless I'm missing something.
2. I'd love to make sure we move some of the other releasers over to the new branch naming approach ASAP, once we've addressed your immediate need.

test/releasers/node.ts
```diff
@@ -289,4 +289,31 @@ describe('Node', () => {
       req.done();
     });
   });
+
+  describe('coercePackagePrefix', () => {
```
joeldodge79, line 293: I think we need a test where `packageName` is not provided to `new Node({})` and that `releasePR.packagePrefix` is set correctly after `releasePR._run` is invoked since that's where the Node releaser [currently sets it](https://github.com/googleapis/release-please/pull/720/files?file-filters%5B%5D=.json&file-filters%5B%5D=.ts#diff-e6f703994fb6593e7ef82b0d3cdf990ebdd51764ac01a2723ffa45136d504730R38)

note: that might not be the best place for a releaser to set it but since it comes from inspecting source code, any usage of `this.packagePrefix` needs to wait till the releaser has had a chance to look it up dynamically.

src/releasers/node.ts
joeldodge79: It's been appropriately moved over into a new `ReleasePR.coercePackagePrefix` - I like that change! However, see my note on the tests as I think it's not quite all the way there yet for Node.

also, in the spirit of "Small CLs" that change could probably have been pulled out as a separate PR ahead of this change.

src/release-pr.ts
```diff
@@ -274,24 +304,60 @@ export class ReleasePR {
...
+  ): Promise<string> {
+    return includePackageName
+      ? `chore: release ${this.packageName} ${version}`
+      : `chore: release ${version}`;
+  }
+
+  // Override this method to detect the release version from code (if it cannot be
+  // inferred from the release PR head branch)
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  protected detectReleaseVersionFromTitle(title: string): string | undefined {
```
joeldodge79, line 320: @bcoe I recall your aversion to the instability of labels to store the version - I'd argue that the PR title is just as fragile - it could be changed after the last release-please run but before merging. It can also be changed after the PR has already been merged

@chingor13 is this here in support of keeping the version name out of the branch? If so, i think we should stick with just the manifest approach - I promise I'll have that up for review soon!

src/release-pr.ts
```diff
@@ -340,4 +406,111 @@ export class ReleasePR {
   addPath(file: string) {
     return ReleasePR.addPathStatic(file, this.path);
   }
+
+  // BEGIN release functionality
+
+  // Override this method to detect the release version from code (if it cannot be
+  // inferred from the release PR head branch)
+  protected async detectReleaseVersionFromCode(): Promise<string | undefined> {
+    return undefined;
```
joeldodge79, line 415: I don't think we're ready to expose this yet, or even if it's a good idea at all

joeldodge79 (changes requested):
Nice work Jeff! Really beautifully written code here and some very common sense refactors. I do have some thoughts:

1. There's a lot going on in this PR. Is it possible to tease some things apart to submit separately? Maybe something like
  - `packageBranchPrefix` -> `coercePackagePrefix` change
  - the switch to using graphql for finding the last merge release if that's the way we decide to go
  - adding the `BranchName` class and tests w/out wiring them up yet
  - wiring up `BranchName` usage?
  - custom PR body and title
Smaller changes are easier to reason about and catch bugs in :-)

2. I'm not sure how I feel about the proliferation of methods for finding a version. On the one hand flexibility gives users more options. On the other hand, it creates higher risk for bugs and might make it harder to keep straight. The aggregate releaser is already going to add another version storing method: the manifest. I might even present an argument to you folks that *all* release-please version tracking should be migrated to that file (the structure of which might need to be updated to scoped by targetReleaseBranch / LTS branch).

src/github.ts
```diff
@@ -547,6 +582,59 @@ export class GitHub {
     return tags;
   }
 
+  async findMergedPullRequests(number?: number): Promise<MergedGitHubPR[]> {
+    const targetBranch = await this.getDefaultBranch();
+    const response = await this.graphqlRequest({
+      query: `query findMergedPullRequests($owner: String!, $repo: String!, $num: Int!) {
```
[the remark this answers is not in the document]
chingor13: I was using this only for detecting the release PR/commit after the merge was completed and we're looking for the commit on the branch to tag (which in theory should be very close to head of the branch).

@bcoe pointed out that I missed the logic for fetching the commits since the last release on a target branch would also need to check the new branch names (which this PR is not yet doing).

For some reason, I thought that the REST endpoint couldn't filter by base branch, but [it totally can](https://docs.github.com/en/rest/reference/pulls#list-pull-requests). I will switch back to the REST endpoint

src/release-pr.ts
chingor13: This would be the hook for looking up the version(s) from the manifest.

src/release-pr.ts
chingor13: This is fallback to support versionless branch names (until we implement the code detection).

src/releasers/java-yoshi.ts
chingor13: Yes, this allows each releaser to choose its strategy for branch naming and will allow gradual upgrades. This is `protected` so it's not part of the public surface area. If we're happy with using `release-please/branches/[branch]/(components/[component])?`globally, then we can refactor this to be done in the base class without breaking anyone.

src/releasers/java-yoshi.ts
```diff
@@ -287,4 +288,46 @@ export class JavaYoshi extends ReleasePR {
...
+  // inferred from the release PR head branch)
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  protected detectReleaseVersionFromTitle(title: string): string | undefined {
+    const pattern = /^chore\((?<branch>[^(]+)\): release ?(?<component>.*) (?<version>\d+\.\d+\.\d+)$/;
+    const match = title.match(pattern);
+    if (match?.groups) {
+      return match.groups['version'];
+    }
+    return undefined;
+  }
```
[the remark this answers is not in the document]
chingor13: I also considered making a `PullRequestTitle` resource name-like class similar to the `BranchName` that knows how to parse and build itself.

src/releasers/node.ts
chingor13: This was an oversight and is missing in this PR. Finding commits since the last release also needs to be able to find the stable release on the branch.

src/util/branch-name.ts
```diff
@@ -0,0 +1,126 @@
...
+  version?: string;
+
+  static parse(branchName: string): BranchName | undefined {
+    const branchNameClass = getAllResourceNames().find(clazz => {
+      return clazz.matches(branchName);
+    });
+    if (!branchNameClass) {
+      return undefined;
+    }
+    return new branchNameClass(branchName);
```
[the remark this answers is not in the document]
chingor13: I was running into typescript issues, but I agree that it's not ideal duplicating the regex match. I will try this again.

src/github.ts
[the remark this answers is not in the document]
chingor13: Ah, I remember additional context - it fixes the ordering. Currently, we're sorting pull requests by `created` which is when the PR was opened. If a release PR has sat open for a long time with many PRs open, it may be buried in the list of closed pull requests and could require much paging to find. Coming from the commit side would give you the history order or when it's merged.

It might be possible to switch to `updated` for the sort order, but still a PR being modified after merge might affect this field too.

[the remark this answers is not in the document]
joeldodge79: > detecting the release PR/commit after the merge was completed and we're looking for the commit on the branch to tag

I actually think there should be one and only one way of finding this - whether it's for creating a tag or for creating/updating the next Release PR. The question is the same regardless of why it's being asked: "what is the commit on targetReleaseBranch that represents the last release" and I think we need to stick with the definition of "it's the last-merged-release-pr's merge commit". I think there should be a single `github.ts` method for this that has these inputs:

- targetReleaseBranch (defaultBranch/LTS branch)
- pattern for releasePRBranch name (constructed from your BranchName stuff with a regex placeholder for version if it's a `AutoreleaseBranchName` branch)

and finds the last merged PR given the above combo.

src/github.ts
```diff
@@ -547,6 +582,59 @@ export class GitHub {
...
+          repository(owner: $owner, name: $repo) {
+            ref(qualifiedName: $targetBranch) {
+              target {
+                ... on Commit {
+                  history(first: $num) {
```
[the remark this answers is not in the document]
chingor13: I probably need to do some testing. 

If you merge or rebase/merge and all the commits are associated to the same PR, this method could return duplicates and could fill the entire page(s) of responses. We would likely still need to paginate the graphql query as well.

src/release-pr.ts
joeldodge79: do we need versionless branch names right now? i.e. is there a need outside the aggregate PR to keep version names out of branches?

chingor13: On of our Java repos has a edge case need for this.

src/github.ts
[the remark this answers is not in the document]
chingor13: The bug Ben pointed out is exactly this. The 2 need to use the same logic.

src/release-pr.ts
joeldodge79: ah, I interpreted `FromCode` to mean per-type specific source code (e.g. version.py, package.json)

I'd prefer to wait on this change till we have something actually ready to implement it. I'd actually like to propose that we use a eventually only use a manifest file for tracking versions. I'd add it as part of the aggregate PR work, providing a migration path that seamlessly switches from encoding the version in the branch name to writing it to the manifest.

For now I feel like we should stick with the current functionality of version-in-branch-name

test/releasers/node.ts
bcoe: As long as we don't break the existing logic, I think it's shaken out both in this PR and in #725, that the fallback `packageName` lookup should be refactored a bit.

I wonder if we could pull it up to `releasea-pr.ts` instead, and have all releasers attempt the lookup., if it's supported by their language.

src/github.ts
```diff
@@ -547,6 +582,59 @@ export class GitHub {
...
+                  }
+                }
+              }
+            }
+          }
+        }`,
+      owner: this.owner,
+      repo: this.repo,
+      branch: targetBranch,
+      num: number ?? 10,
```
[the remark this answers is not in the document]
bcoe: 👍 I think perhaps this doesn't need to grab as many merged release PRs, because it's going directly to a well known branch name? or perhaps I'm missing something.

src/github.ts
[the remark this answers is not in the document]
bcoe: I wonder if rather than paginating, there's a sort we can apply on the lookup, so the most recent interaction with the well known branch name is at the top? we should be able to test this in the sandbox.

src/release-pr.ts
bcoe: If this unblocks @chingor13's work on Java, I'm tempted to run with this approach for the time being.

An additional concern I had with labels, is that our forking GitHub actions have permission to open PRs, but not to label.  So using labels had an additional permission headache.

I like that the lookup logic has been encapsulated and tested.

bcoe:
> I'm not sure how I feel about the proliferation of methods for finding a version. 

@joeldodge79 I'm a little concerned about this too, I think @chingor13's work to address the `latestTag` bug will hopefully move things closer together.

joeldodge79 (commented):
looking good! one more suggestion for ya and a question about the java edge case for not using version in branch name.

src/github.ts
```diff
@@ -547,79 +558,155 @@ export class GitHub {
...
+      }
+      page += 1;
+      mergedPullRequests = await this.findMergedPullRequests(
+        targetBranch,
+        page
+      );
+    }
+    return undefined;
+  }
+
```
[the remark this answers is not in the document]
chingor13: I think we can skip the optimization on filtering the exact release branch name until we migrate all releasers to start using the stable branch names -- we need to support the fallback to the old, versioned style during the migration period.

test/releasers/node.ts
joeldodge79: actually, I think this is still broken for node as setting `this.packagePrefix` only happens in the constructor on this branch but `this.packageName` gets reset in `Node._run`...

chingor13: Yes, looks like this will cause the PR to be opened with the wrong packagePrefix

joeldodge79 (approved):
LGTM 🌲

bcoe (approved):

bcoe (approved):
This is looking good to me, if you're feeling it's ready to land?

I will test the finished state on a few repositories I manage.

@chingor13 would you be able to run the CLI against a testing Java repo, to make sure that the behavior is as expected?

chingor13:
For Java, it's opening pull requests to the right branches https://github.com/chingor13/java-bigtable-hbase/pull/9 (commits are weird in this one because there's no releases to master since the code was migrated here), https://github.com/chingor13/java-bigtable-hbase/pull/10

src/github.ts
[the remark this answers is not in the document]
chingor13: I remember an additional reason why I was using commits to the branch to find the pull requests. The `base` ref of the associated pull request may not have been the release branch we're looking for.

Consider 2.0.0, 2.1.0 are both released from `main`. We now create a `2.0.x` branch. To find the latest release on the `2.0.x` branch, we need to consider *any* release pull request in history (the `2.0.0` release PR was against `main`, not the `2.0.x` branch - the `base` ref is not important).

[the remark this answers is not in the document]
joeldodge79: @chingor13 I want to make sure I understand what you're saying here:

Were versions `"2.0.0"` and `"2.1.0"` releases both release-please PRs that were merged to `main`? How does branch `2.0.x` come into existence? Is it manually created by branching off `main`? Ah, I think I get it. The problem is "bootstrapping" the initial version on the release branch, right? I.e. when release-please is configured for the first time with `defaultBranch` set to `2.0.x` it won't find the last release/version? I think we can solve this with the manifest file (thoughts forthcoming...) but till then does the `release-as` footer suffice as a work-around for that first release on the LTS branch?

PR #725, opened 2021-01-29
feat(cli)!: refactor factory/CLI to be more testable

bcoe:
Taking inspiration from @jbottigliero's work on #581, this PR refactors our factory class, and the CLI entry-point to be more easily tested/shared:

* there's now a `factory.runCommand` helper that can be used as an entry-point to GitHub releases or Release PRs.
* `bin/release-please.ts` does not execute unless it's being run by `node`, allowing it to be tested in isolation.
* I refactored how the factory exports its helpers, making it easier to mock with `sinon`.

future work:

* we should flesh out unit tests for the CLI further:
  * adding test cases for loading config.
  * adding a test case that resembles how config will be passed in by the GitHub action.
* I did not pull over the improved conventional changelog section types that @jbottigliero wrote (_perhaps you could submit a patch?_).
* I have not added support for config files yet (@joeldodge79 given you use a GitHub action, you might want to think through how the config is actually loaded, since yargs assumes a file system. You will likely want to have something that reaches out using the GitHub API and pulls in config).

feat: packageName parameter is now optional
refactor!: helpers in factory class renamed.
refactor!: more options pulled to top level (some types changed in process)
refactor!: GitHub release now uses "run" rather than "createRelease" to execute
refactor!: drop unused proxy-key parameter.

CC: @joeldodge79


src/release-pr.ts
```diff
@@ -111,11 +98,13 @@ export class ReleasePR {
     this.repoUrl = options.repoUrl;
     this.token = options.token;
     this.path = options.path;
-    this.packageName = options.packageName;
+    // If no packageName has been provided, use the name
+    // of the repository:
+    const [, repo] = parseGithubRepoUrl(this.repoUrl);
+    this.packageName = options.packageName || repo;
```
joeldodge79, line 104: huh, so it looks like `github-release` effectively [requires a package name](https://github.com/googleapis/release-please/blob/ce0d2d8d5bc1b278b5a5f8ebd81c6ba16db60645/src/github-release.ts#L80). `release-pr` does not... but if we want to set one by default here I think we should follow that same lookup logic rather than just choosing the repo name?

For consistency I'd actually like to see `release-pr` require it too and remove the concept of choosing wether to `OpenPROptions.includePackageName` But that would be a breaking change (breaking how last release pr is found) and maybe the asymmetry between including the package name in the release/tag but not having it show up in the PR is fine... probably just my AR/OC tendencies getting tickled :-)

src/bin/release-please.ts
```diff
@@ -78,55 +67,30 @@ const argv = yargs
...
         })
-        .option('label', {
-          default: 'autorelease: pending',
-          describe: 'label to remove from release PR',
-        })
         .option('release-type', {
           describe: 'what type of repo is a release being created for?',
           choices: getReleaserNames(),
           default: 'node',
         })
```
joeldodge79, line 93: why did this `release-type` stay here? it's present below in the common options

test/cli.ts
```diff
@@ -0,0 +1,77 @@
...
+          return Promise.resolve(undefined);
+        }
+      );
+      parser.parse(
+        'release-pr --repo-url=googleapis/release-please-cli --package-name=cli-package'
+      );
+      assert.strictEqual(classToRun!.repoUrl, 'googleapis/release-please-cli');
+      assert.strictEqual(classToRun!.packageName, 'cli-package');
+      // Defaults to Node.js release type:
+      assert.strictEqual(classToRun!.releaseType, 'node');
```
joeldodge79, line 50: maybe a runtime `instanceof ReleasePR` assertion since both `GitHubRelease` and `ReleasePR` would pass these three assertions I think?

src/factory.ts
```diff
@@ -0,0 +1,72 @@
...
+  return releaser;
+}
+
+export const factory = {
+  githubRelease,
+  releasePR,
+  releasePRClass,
+  run,
+  runCommand,
+};
```
joeldodge79, line 72: I like this common factory. nifty how you collected them together like this for a single entry point.

joeldodge79 (changes requested):
nice improvements! some questions and a couple nits

regarding the aggregate `packages` config: I'd rather stay within the bounds of `yargs` for cli and action inputs defined in the workflow for the GH-action side:
- **cli**: `--packages` flag for the cli: json encoded string (or array of json strings)
  - consider adding `yargs.config()` to make cli users' lives easier supporting actual nested json in the json config file rather than requiring embedded json encoded strings in the json config file)
  - cli users may happen to choose to store their CLI config somewhere in their repo but that doesn't affect us
- **action**: `packages` input key: array of objects defined directly in the workflow file

src/factory.ts
bcoe: It's maybe note the most "TypeScripty" looking code, but it's an approach I learned from working on Node:

https://github.com/nodejs/node/blob/master/lib/fs.js#L2058

I've found it makes it really easy to mock the methods, vs., static methods.

src/bin/release-please.ts
```diff
@@ -170,10 +132,6 @@ const argv = yargs
     default: 'https://api.github.com',
     type: 'string',
   })
-  .option('proxy-key', {
-    describe: 'key used by some GitHub proxies',
-    type: 'string',
-  })
```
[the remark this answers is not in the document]
bcoe: @joeldodge79 this was used to interact with an internal proxy that has be decommissioned.

src/release-pr.ts
bcoe: I was on the fence about this change too, to minimize the code change in this PR perhaps I should just default to an empty string -- and we can improve things later?

src/bin/release-please.ts
bcoe: oversight, good catch.

test/cli.ts
bcoe: Done, good idea for a test.

src/release-pr.ts
bcoe: @joeldodge79 I suggest we just start with the empty string.

jbottigliero (approved):
This looks good to me! 

In comparison to #581, one of the things that it looks like the `google-github-actions/release-please-action` (and other potential Node API users) will still have to handle is the empty command flow ([calling both the `github-release` and `release-pr` in succession](https://github.com/google-github-actions/release-please-action/blob/main/index.js#L29)). I saw this logic being shared as way to create a single source of truth for release-please behaviors, especially with the idea of introducing new triggers (ie. [the Google Cloud Build example](https://gist.github.com/jbottigliero/29e8de2dda4db7d3f07626dc99547026)). It seems like that could maybe still be done in a follow-up patch in addition to the shared changelog options parsing?

src/release-pr.ts
```diff
@@ -111,11 +98,10 @@ export class ReleasePR {
     this.repoUrl = options.repoUrl;
     this.token = options.token;
     this.path = options.path;
-    this.packageName = options.packageName;
+    this.packageName = options.packageName || '';
```
joeldodge79, line 101: I'm ok with an empty string if you don't think the following is appropriate (moved further down where `this.{gh,releaseType}` are available):

```suggestion
    this.packageName = options.packageName || await factory
      .releasePRClass(this.releaseType)
      .lookupPackageName(this.gh, this.path);
```

joeldodge79 (approved):
LGTM (up to you if you want to incorporate my suggestion or not, I don't feel strongly either way)

bcoe:
> I saw this logic being shared as way to create a single source of truth for release-please behaviors, especially with the idea of introducing new triggers (ie. the Google Cloud Build example). It seems like that could maybe still be done in a follow-up patch in addition to the shared changelog options parsing?

@jbottigliero if it's okay with you, let's think about this in a follow up. One slight concern I have, is that creating a GitHub Release, vs., creating a Release PR, take enough divergent config options, that we might still want to do some validation in the CLI and action before calling a factory?

src/release-pr.ts
chingor13: Some release strategies prefer not having a package name in the release title (for Java, we usually just have the release title be `v1.2.3` rather than `some-library-name v1.2.3`.

src/factory.ts
```diff
@@ -0,0 +1,72 @@
...
+import {ReleasePR, ReleasePROptions} from './release-pr';
+import {RubyReleasePROptions} from './releasers/ruby';
+import {
+  GitHubReleaseOptions,
+  GitHubRelease,
+  ReleaseResponse,
+} from './github-release';
+import {getReleasers} from './releasers';
+
+export function runCommand(
```
chingor13, line 27: Do you need these extra exports if they are included into the single `factory` export?

chingor13: Is this extra factory necessary? The only place this is called is in the argv parsing and at that point you already know which command you're trying to execute.

src/index.ts
```diff
@@ -12,13 +12,25 @@
...
+
+export {ReleaseCandidate, ReleasePR, ReleasePROptions} from './release-pr';
+
+// Configuration options shared by Release PRs and
+// GitHub releases:
+export interface SharedOptions {
+  label?: string;
+  repoUrl: string;
+  path?: string;
+  packageName?: string;
```
chingor13, line 25: Can we call out `packageName` now being optional in the release notes?

chingor13 (commented):
A few comments/questions.

Also, can you write in the release notes what needs to be renamed? The GitHub app will need to make some code changes and it would be nice if the notes explicitly state `X` is renamed to `factory.Y`.

src/factory.ts
bcoe: I was thinking it might be nice in the action, but maybe it's overkill.

bcoe: @chingor13 actually, one reason I would like to keep this approach, was that it made it really nice to test the CLI:

```js
      let classToRun: GitHubRelease;
      // This logic is used to capture the class that would
      // be executed, allowing us to validate that the appropriate
      // properties were assigned from the argument parser:
      sandbox.replace(
        factory,
        'run',
        (runnable): Promise<undefined> => {
          classToRun = runnable as GitHubRelease;
          return Promise.resolve(undefined);
        }
      );
      parser.parse(
        'github-release --repo-url=googleapis/release-please-cli --package-name=cli-package'
      );
```

Because I was able to mock just the logic that runs the GitHub Release or Release PR, without actually running it.

chingor13 (approved):

JustinBeckwith (approved):

src/factory.ts
```diff
@@ -0,0 +1,72 @@
...
+
+function githubRelease(options: GitHubReleaseOptions): GitHubRelease {
+  return new GitHubRelease(options);
+}
+
+function releasePR(options: ReleasePROptions): ReleasePR {
+  const releaseOptions: ReleasePROptions | RubyReleasePROptions = options;
+  return new (factory.releasePRClass(options.releaseType))(releaseOptions);
+}
+export function releasePRClass(releaseType: string): typeof ReleasePR {
```
alexander-fenster, line 57: nit: missing empty line between functions

PR #726, opened 2021-01-31
feat: monorepo node package dependency updates with lerna

joeldodge79:
this is just a prototype PR to discuss the plumbing implementation - not intended to be merged

Note to readers: my domain experience in package management in general is slim to none. My experience configuring javascript monorepos with intra-repo dependencies is likewise very limited. I acknowledge that I'm operating off one way of setting this up: [yarn-classic workspaces in our sdk-codegen repo](https://github.com/looker-open-source/sdk-codegen/blob/20ca1915776b657f958cf3182843d72fc4ec1344/package.json#L7). please feel free to point out other approaches that this prototype might fail to address.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,9 +56,13 @@
   "dependencies": {
     "@conventional-commits/parser": "^0.4.1",
     "@iarna/toml": "^2.2.5",
+    "@lerna/package": "^3.16.0",
+    "@lerna/package-graph": "^3.18.5",
+    "@lerna/run-topologically": "^3.18.5",
     "@octokit/graphql": "^4.3.1",
     "@octokit/request": "^5.3.4",
     "@octokit/rest": "^18.0.4",
+    "@types/npm-package-arg": "^6.1.0",
     "chalk": "^4.0.0",
     "code-suggester": "^1.4.0",
     "conventional-changelog-conventionalcommits": "^4.4.0",
--- a/src/util/update-node-workspace-packages-dependencies.ts
+++ b/src/util/update-node-workspace-packages-dependencies.ts
@@ -1,0 +1,64 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Package = require('@lerna/package');
+import PackageGraph = require('@lerna/package-graph');
+import runTopologically = require('@lerna/run-topologically');
+
+export interface PackageJson {
+  name: string;
+  version: string;
+  dependencies?: Record<string, string>;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any;
+}
+
+export type PathPackages = Record<string, PackageJson>;
+
+// Inspired by https://github.com/lerna/lerna/blob/72414ec1c679cf8a7ae4bfcefce52d50a6120a70/commands/version/index.js#L495
+// Expected input is a mapping of 'path/to/dirContainingPkgJson' => JSON.parse(package.json)
+// It is also assumed that each package.json version is the new version.
+// Each package.json dependency on another package will be updated to "^pkg.latest.version"
+export async function updateNodeWorkspacePackagesDependencies(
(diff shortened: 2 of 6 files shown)


src/util/update-node-workspace-packages-dependencies.ts
```diff
@@ -0,0 +1,64 @@
...
+  dependencies?: Record<string, string>;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any;
+}
+
+export type PathPackages = Record<string, PackageJson>;
+
+// Inspired by https://github.com/lerna/lerna/blob/72414ec1c679cf8a7ae4bfcefce52d50a6120a70/commands/version/index.js#L495
+// Expected input is a mapping of 'path/to/dirContainingPkgJson' => JSON.parse(package.json)
+// It is also assumed that each package.json version is the new version.
```
joeldodge79, line 31: the caller (aggregate releaser) of this function would have already run the node releaser per package so each package's `version` would already be updated.  The caller then passes them all into this function which in turn updates each packages intra-repo dependencies

package.json
```diff
@@ -56,9 +56,13 @@
   "dependencies": {
     "@conventional-commits/parser": "^0.4.1",
     "@iarna/toml": "^2.2.5",
+    "@lerna/package": "^3.16.0",
+    "@lerna/package-graph": "^3.18.5",
+    "@lerna/run-topologically": "^3.18.5",
```
bcoe, line 61: It's neat how well lerna is decomposed into submodules.

bcoe (commented):
This is a surprisingly small amount of code.

One question, what about `package-lock.json`? will we need to pre-load this file if it exists too? I don't think this would add too much overhead, since we'll need to load the contents when we update it any ways.

test/util/update-node-workspace-packages-dependencies.ts
```diff
@@ -0,0 +1,75 @@
...
+import {
+  updateNodeWorkspacePackagesDependencies,
+  PathPackages,
+} from '../../src/util/update-node-workspace-packages-dependencies';
+import {describe, it} from 'mocha';
+import {expect} from 'chai';
+
+describe('updateNodeWorkspacePackagesDependencies', () => {
+  it('updates package dependencies', async () => {
+    function getPackages(updated: boolean): PathPackages {
```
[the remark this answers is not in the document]
joeldodge79: I'm still thinking about where/when this code is called. Somehow the aggregate releaser gets all the package.json contents for each package from the individual releaser (which has already bumped the top level version) and passes them through this lerna wrapper before including them in the PR.

> enough information to run the lerna tree resolution 

I think I have a bug in this implementation: I'm ignoring the workspace configuration and defining the entire workspace only from the `PathPackages` input. Instead I need to load up all the package versions (currently being updated or not) so that any package being currently updated can have it's dependencies set to the latest available whether that dependency is being updated in this run or not. I'll add tests for this, e.g:


**release1**: all packages version and dependencies bump
```
pkgA 1.1.1 => 1.1.2
pkgB 1.1.1 => 1.1.2
- pkgA "^1.1.1" => "^1.1.2"
pkgC 1.1.1 => 1.1.2
- pkgB "^1.1.1" => "^1.1.2"
```

**release2**: A and C version bumps
```
pkgA 1.1.2 => 1.1.3
// pkgB unchanged since it isn't part of this release
pkgC 1.1.2 => 1.1.3
- // pkgB dep unchanged
```

**release3**:  B version bump
```
pkgB 1.1.2 => 1.1.3
- pkgA "^1.1.2" => "^1.1.3"  // pkgA dep must be updated to the latest even though A isn't part of this release.
// pkgC unchanged since it isn't part of this release
```

**release4**: A version bump
```
pkgA 1.1.3 => 1.1.4
// pkgB, pkgC unchanged since they aren't part of this release.
```

**release5**:
```
pkgB 1.1.3 => 1.1.4
- pkgA "^1.1.3" => "^1.1.4"  // pkgA dep must be updated to the latest even though A isn't part of this release.
pkgC 1.1.3 => 1.1.4
- pkgB "^1.1.2" => "^1.1.4"
```

src/util/update-node-workspace-packages-dependencies.ts
```diff
@@ -0,0 +1,64 @@
...
+// Each package.json dependency on another package will be updated to "^pkg.latest.version"
+export async function updateNodeWorkspacePackagesDependencies(
+  packages: PathPackages
+): Promise<PathPackages> {
+  const lernaPkgs: Package[] = [];
+  const newVersions: Record<string, string> = {};
```
joeldodge79, line 37: note to self. `newVersions` needs to be `allVersions` - a map of packageName => currentOrUpdatedVersion

what defines `allVersions`? either the `--packages` aggregate config or the top level package.json `workspaces`?

joeldodge79:
Closing for now. I'll submit a real PR once the aggregate releaser stuff is done

PR #730, opened 2021-02-02
Revert "fix: use resolove to find imports in tests"

bcoe:
This breaks the packaged version of the library.

Reverts googleapis/release-please#715


bcoe:
@safarmer #715 broke a few things, because `require` is relative absolute to the file that `require` is happening within, where as `resolve` is relative to the working directory.

I'm not sure what the best way to make things work for continuous running of tests, with package dependencies is, but it would need to also use paths relative to the files.

PR #734, opened 2021-02-02
test: missed release-notes snapshot file

chingor13:
From #720


chingor13:
Not sure why this is popping up as #720 isn't merged yet

PR #737, opened 2021-02-04
refactor: use RELEASE_PLEASE for 'release-please'

joeldodge79:


src/util/branch-name.ts
```diff
@@ -39,14 +44,14 @@ export class BranchName {
...
   static ofTargetBranch(targetBranch: string): BranchName {
-    return new DefaultBranchName(`release-please/branches/${targetBranch}`);
+    return new DefaultBranchName(`${RELEASE_PLEASE}/branches/${targetBranch}`);
   }
   static ofComponentTargetBranch(
     component: string,
     targetBranch: string
   ): BranchName {
     return new ComponentBranchName(
-      `release-pleaase/branches/${targetBranch}/components/${component}`
```
joeldodge79, line 49: this cute typo was the motivation for this change :-)

src/util/branch-name.ts
```diff
@@ -12,6 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// cannot import from '..' - transpiled code references to RELEASE_PLEASE
+// at the script level are undefined, they are only defined inside function
+// or instance methods/properties.
+import {RELEASE_PLEASE} from '../constants';
```
joeldodge79, line 18: I really wanted to import from '..' but, even though typescript was happy and transpiled the code (`const __1 = require("..")`), the references to `__1.RELEASE_PLEASE` at the script level (where the `{COMPONENT,DEFAULT}_PATTERN` need to use them) were undefined. `__1.RELEASE_PLEASE` was only defined inside functions or instance methods/properties (I also tried static properties and it was undefined there too).

for some reason it works fine like this. transpiled code:
```
const constants_1 = require("../constants");
.
.
.
// happy as a clam
const DEFAULT_PATTERN = `^${constants_1.RELEASE_PLEASE}/branches/(?<branch>[^/]+)$`;
```

src/index.ts
```diff
@@ -29,7 +29,7 @@ export interface SharedOptions {
   octokitAPIs?: OctokitAPIs;
 }
 
-export const DEFAULT_LABELS = 'autorelease: pending';
+export {DEFAULT_LABELS, RELEASE_PLEASE} from './constants';
```
joeldodge79, line 32: Given the odd behavior I encountered in `branch-name.ts` I wonder if we should even re-export these names here?

bcoe (approved):

bcoe (approved):
should we also update `bin/release-please.ts`? I think it populates default labels.

src/util/branch-name.ts
bcoe: I'm surprised by this behavior. For now, I suggest we stop exporting the constants in `index.ts`, like you suggest, to prevent hiccups.

joeldodge79: I tried really hard to find a reference on the internet but my searching skills were not up to the task. I'll update to remove the re-export from `index.ts`

joeldodge79:
> this test didn't cause any tests to break for Java

I don't quite follow... do you mean _this change_ ? which java tests might you think would have broken?

> add a smoke test

do you mean a round-trip test that verifies that `RELEASE_PLEASE` is actually set to the string `release-please`? the branch-name tests cover it indirectly [here](https://github.com/googleapis/release-please/blob/f8b036d5d97847643afa9788eea599a13bfee57e/test/util/branch-name.ts#L48) and [here](https://github.com/googleapis/release-please/blob/f8b036d5d97847643afa9788eea599a13bfee57e/test/util/branch-name.ts#L58) but I could add a more explicit test simply against the imported const (something like):

```typescript
import {RELEASE_PLEASE} from ../constants
...
it('exports RELEASE_PLEASE', () => {
  expect(RELEASE_PLEASE).to.equal('release-please');
});
```

joeldodge79:
per offline discussion:

- I'll remove the `src/index.ts` re-exports of the `src/constants.ts` (anyone have any clue why importing from `'..'` causes the names to be undefined at the top script level and class static level?)
- I'll add a test that exercises `ofComponentTargetBranch` which would have caught this bug
- investigation: why did [this snapshot](https://github.com/googleapis/release-please/pull/720/files#diff-8a12a0c638ecdf4c6dc65c5b7e393b349ae6b252b6c00fd68d96f2237aa8261fR183) not capture the `release-pleaase` bug? I would have expected "branch": "release-pleaase/branches/1.x"

bcoe (approved):

joeldodge79:
@bcoe in retrospect I feel a little silly that I didn't initially narrow in on why the `"release-pleaase"` bugfix didn't either break any tests or require any test changes. I got distracted by trying to do what seemed super simple (add a constant to `src/index.js` and use it) and having it not work in a really weird way that I still don't understand. When I finally debugged/characterized that issue I'd mostly forgotten about the initial bug! Anyway, thanks for bringing it back to the original problem :-)

PR #741, opened 2021-02-04
feat: add GitHub#commitsSince and GitHub#findMergeCommit

### LINKED ISSUE #740: `GitHub#latestTag` fails to find release to different branch
Consider the following scenario.

* versions 1.0.0, 1.1.0, and 1.2.0 are released from `main`.
* branch a `1.0.x` branch from the `1.0.0` commit and backport a fix into that branch
* when considering the `1.0.x` branch, the `1.0.0` tag should be the latest tag found

Currently, the logic fails to find this release because we are filtering pull requests whose base branch is `1.0.x`. At the time of that release, however, the base branch was actually `main`.

Instead, we need to look through the latest commits to that branch and find the latest one that looks like a release commit.

chingor13:
Adds a GraphQL query that looks at commits to the current branch and iterates/pages through them to find a commit that matches certain conditions. We use this to find the last commit to a branch that looks like a release. 

`GitHub#latestTag` now uses this method of finding the last release PR. The previous method looks at the `base` branch of a pull request.

Fixes #740

### DIFF
--- a/__snapshots__/github.js
+++ b/__snapshots__/github.js
@@ -529,3 +529,87 @@ exports['GitHub findMergedPullRequests finds merged pull requests with labels 1'
     "body": "(cherry picked from commit a03d7dcdd8458d5c3542ec2914a031afd69815cf)\r\n"
   }
 ]
+
+exports['GitHub commitsSince finds commits up until a condition 1'] = [
+  {
+    "sha": "e6daec403626c9987c7af0d97b34f324cd84320a",
+    "message": "Merge pull request #7 from chingor13/feature-branch-plain-merge\n\nfeat: feature that will be plain merged",
+    "files": []
+  }
+]
+
+exports['GitHub commitsSince paginates through commits 1'] = [
+  {
+    "sha": "e6daec403626c9987c7af0d97b34f324cd84320a",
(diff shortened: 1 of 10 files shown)


bcoe:
@chingor13 since this makes the "setting up your repo" initially story worse. I'd love for us to introduce the ability to provide a `--first-sha`, as an argument before we remove looking at tags.

Otherwise this seems like a good refactor to me.

bcoe (approved):
I was initially thinking it was the fallback that was being refactored, not the initial lookup of `latestTag` (before the tag fallback).

This seems good to me (_once you run lint_).

bcoe (changes requested):
Two concerns with our continue adoption of GraphQL here:

1. I think we should have a bound on how far we paginate, for the case of people bootstrapping a repository (250 is probably fine).
2. I found that graphql queries often initially timeout, which is why I had retry logic built in (could we figure out a way to abstract retrying).

chingor13:
> 2. I found that graphql queries often initially timeout, which is why I had retry logic built in (could we figure out a way to abstract retrying).

Let's push the 502 retries into `GitHub#graphqlRequest` in a separate PR

bcoe (approved):

PR #745, opened 2021-02-05
refactor: stricter releaseType typing

joeldodge79:
releasers/index.js is the source of truth for available releasers. This
change strengthens that by introducing the ReleaseType as the typed
definition of what releasers are available.

Keeping the source of truth (ReleaseType) separate from actual ReleasePR
subclasses allows multiple types to point to the same subclass (e.g. the
case of "go" and "go-yoshi"). Prior to this change it would not have
been possible (via yargs cli options) to specify `--release-type go` (or
at least it's not listed in the help output)

Since the only case for ReleasePR.releaserName is now obviated it's been
removed. If there a need in the future for a "display friendly"
representation of the releaser instance then we can add it back.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,3 +1,9 @@
 {
-  "extends": "./node_modules/gts"
+  "extends": "./node_modules/gts",
+  "rules": {
+    "@typescript-eslint/no-unused-vars": [
+      "error",
+      {"argsIgnorePattern": "^_", "varsIgnorePattern": "^_$"}
+    ]
+  }
 }
--- a/src/releasers/index.ts
+++ b/src/releasers/index.ts
@@ -27,30 +27,50 @@ import {TerraformModule} from './terraform-module';
 import {Rust} from './rust';
 import {OCaml} from './ocaml';
 
-// TODO: add any new releasers you create to this list:
-export function getReleasers(): {[key: string]: typeof ReleasePR} {
-  const releasers = {
-    go: GoYoshi, // TODO(codyoss): refactor this into a more generic go strategy.
-    'go-yoshi': GoYoshi,
-    'java-bom': JavaBom,
-    'java-yoshi': JavaYoshi,
-    node: Node,
-    'php-yoshi': PHPYoshi,
-    python: Python,
-    'ruby-yoshi': RubyYoshi,
-    ruby: Ruby,
-    simple: Simple,
-    'terraform-module': TerraformModule,
-    rust: Rust,
-    ocaml: OCaml,
-  };
+// add any new releasers you create to this type as well as the `releasers`
+// object below.
+export type ReleaseType =
+  | 'base' // not exposed to end users
+  | 'go'
+  | 'go-yoshi'
+  | 'java-bom'
+  | 'java-yoshi'
+  | 'node'
+  | 'php-yoshi'
+  | 'python'
+  | 'ruby-yoshi'
+  | 'ruby'
+  | 'simple'
+  | 'terraform-module'
+  | 'rust'
+  | 'ocaml';
+
+type Releasers = Partial<Record<ReleaseType, typeof ReleasePR>>;
+
+const releasers: Releasers = {
+  go: GoYoshi, // TODO(codyoss): refactor this into a more generic go strategy.
+  'go-yoshi': GoYoshi,
+  'java-bom': JavaBom,
+  'java-yoshi': JavaYoshi,
+  node: Node,
+  'php-yoshi': PHPYoshi,
+  python: Python,
+  'ruby-yoshi': RubyYoshi,
+  ruby: Ruby,
+  simple: Simple,
+  'terraform-module': TerraformModule,
+  rust: Rust,
+  ocaml: OCaml,
+};
+
+export function getReleasers(): Releasers {
   return releasers;
 }
 
-export function getReleaserNames(): string[] {
-  const releasers = getReleasers();
-  return Object.keys(releasers).map(key => {
-    const releaser = releasers[key];
-    return releaser.releaserName;
-  });
+export function getReleaserTypes(): readonly ReleaseType[] {
+  const names: ReleaseType[] = [];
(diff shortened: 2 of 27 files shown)


test/github-release.ts
```diff
@@ -85,68 +89,30 @@ describe('GitHubRelease', () => {
...
       strictEqual(created!.patch, 3);
       strictEqual(created!.draft, false);
     });
 
+    it('creates and labels release on GitHub with invalid semver', async () => {
+      const releaser = gitHubReleaserForVersion('1A.B.C');
+      const created = await releaser.run();
+      expect(created).to.be.undefined;
+    });
+
```
joeldodge79, line 110: I was gaming the coverage a little for this PR by adding this (and realizing the mock setup should be DRY'd out). Kind of broke my own "single purpose" PR rule. Kept it as a separate commit so I can easily remove this change if it's too distracting from the main idea of a `ReleaseType`

src/releasers/index.ts
```diff
@@ -27,30 +27,50 @@ import {TerraformModule} from './terraform-module';
...
-    rust: Rust,
-    ocaml: OCaml,
-  };
+// add any new releasers you create to this type as well as the `releasers`
+// object below.
+export type ReleaseType =
+  | 'base' // not exposed to end users
+  | 'go'
+  | 'go-yoshi'
+  | 'java-bom'
```
bcoe, line 36: I feel like this makes the CLI much more ergonomic, good refactor 👍

test/factory.ts
```diff
@@ -93,4 +93,16 @@ describe('factory', () => {
       expect(githubRelease.releaseType).to.be.undefined;
     });
   });
+  describe('run', () => {
+    it('runs a runnable', async () => {
+      const runnable = factory.releasePR({
+        repoUrl: 'googleapis/simple-test-repo',
```
bcoe, line 99: Nice, thanks for bumping the coverage up a bit.

bcoe (changes requested):
As long as we're fleshing out tests a bit, why don't we add a cli test for the CLI throwing on an unknown releaser.

You can test like this:

```
parser.parse('some command', (err, argv, output) => {
  // assert something.
}
```

src/releasers/index.ts
```diff
@@ -27,30 +27,50 @@ import {TerraformModule} from './terraform-module';
...
+  'terraform-module': TerraformModule,
+  rust: Rust,
+  ocaml: OCaml,
+};
+
+export function getReleasers(): Releasers {
   return releasers;
 }
 
-export function getReleaserNames(): string[] {
```
chingor13, line 50: Note that this is a breaking change (the GitHub bot references this one)

chingor13 (approved):
I wouldn't block this PR on adding the new test for the failure case. We can add that in a small separate PR.

src/releasers/index.ts
joeldodge79: oh, good point. Should I leave it in (calling `getReleaserTypes()` under the hood)?

joeldodge79: if so is it ok if the return type is `ReleaseType[]`?

.eslintrc.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": "./node_modules/gts"
+  "extends": "./node_modules/gts",
+  "rules": {
+    "no-unused-vars": [{ "argsIgnorePattern": "^_"}]
+  }
```
joeldodge79, line 5: is this ok? it would get rid of some `// eslint-disable-next-line @typescript-eslint/no-unused-vars`

src/releasers/index.ts
```diff
@@ -53,20 +53,24 @@ const releasers: Releasers = {
...
 };
 
 export function getReleasers(): Releasers {
   return releasers;
 }
 
+export function getReleaserNames(): readonly ReleaseType[] {
+  return getReleaserTypes();
+}
+
```
joeldodge79, line 73: put this back in for backwards compatibility

.eslintrc.json
joeldodge79: oops, guess not :-(

joeldodge79: actually, this made `gts` happy: `"@typescript-eslint/no-unused-vars": ["error", { "argsIgnorePattern": "^_"}]`

src/releasers/index.ts
chingor13: I think it should be. `ReleaseType` is a limited set of strings and the typing is only for compile time.

test/cli.ts
```diff
@@ -50,6 +52,41 @@ describe('CLI', () => {
       // Defaults to Node.js release type:
       assert.strictEqual(classToRun.releaseType, 'node');
     });
+    it('validates releaseType choices', () => {
```
bcoe, line 55: nit, we should probably have a `(done)` callback here, and call it after our assertions, in case the parser were to simply not be called.

.eslintrc.json
```diff
@@ -1,3 +1,9 @@
 {
-  "extends": "./node_modules/gts"
+  "extends": "./node_modules/gts",
+  "rules": {
```
bcoe, line 3: I believe this file is still autogenerated in `synth.py` nightly, you can just add an `excludes` on `templates.copy` sorry for the nuisance.

bcoe (approved):
two small nits from me that we should address, but otherwise looks good.

synth.py
```diff
@@ -11,7 +11,7 @@
   'README.md',
   'CONTRIBUTING.md',
   '.eslintignore',
-  '.eslintrc.yml',
+  '.eslintrc.json',
```
joeldodge79, line 14: are [these](https://github.com/googleapis/synthtool/tree/master/synthtool/gcp/templates/node_library) the templates? if so there's no longer a `.eslintrc.yml`

bcoe: I wonder if we removed it when we migrated to a newer version of gts at some point.

PR #748, opened 2021-02-05
feat: add support for helm

### LINKED ISSUE #747: Support Helm Chart Packages

Helm Chart is not supported 

I want it automatically update Helm Chart version while creating a release
 **Additional context**

abhinavkhanna-sf:

Fixes #747 🦕


__snapshots__/chart-yaml.js
```diff
@@ -0,0 +1,5 @@
+exports['Chart.yaml updateContent updates version in Chart.yaml 1'] = `
+version: 1.1.0
+
```
bcoe, line 3: these extra few newlines are a bit weird, perhaps confirm this is only in the snapshot, and not happening when the PR is created?

package.json
```diff
@@ -51,7 +51,9 @@
     "nock": "^13.0.0",
     "sinon": "^9.0.3",
     "snap-shot-it": "^7.0.0",
-    "typescript": "^3.8.3"
+    "typescript": "^3.8.3",
+    "js-yaml": "^4.0.0",
```
bcoe, line 55: I believe we need `js-yaml` to be an actual dependency, not a `dev-dependency` -- since you'll want it when running the application.

`@types/js-yaml` can stay as a dev-dependency.

src/releasers/helm.ts
```diff
@@ -0,0 +1,144 @@
...
+      commits,
+      githubRepoUrl: this.repoUrl,
+      bumpMinorPreMajor: this.bumpMinorPreMajor,
+      changelogSections: this.changelogSections,
+    });
+    const candidate: ReleaseCandidate = await this.coerceReleaseCandidate(
+      cc,
+      latestTag
+    );
+    const changelogEntry: string = await cc.generateChangelogEntry({
```
bcoe, line 67: I'll trust you're the expert where the actual Helm update process is concerned 😄 

I've just asked that folks own the first few bugs in their releaser, as other people adopt it.

__snapshots__/chart-yaml.js
bcoe, line 3: should this snapshot also contain `name:` like in your fixture.

bcoe (changes requested):
This is looking really solid, a few notes:

* I saw the unused `test/releasers/fixtures/helm/commits.json`, I'm guessing it's your intention to create some tests for the releaser (which might be where this file came from). If you use `sinon` and mock the functions in `github.ts`, I think you can avoid needing to create a `commits.json` fixture. See [this example from rust](https://github.com/googleapis/release-please/blob/master/test/releasers/rust.ts#L34).
* It seemed odd to me that your snapshot file didn't include the name from your fixture -- I think you should probably dig into this.

Structurally, everything is looking good to me.

__snapshots__/chart-yaml.js
abhinavkhanna-sf: fixed

package.json
abhinavkhanna-sf: fixed

__snapshots__/chart-yaml.js
abhinavkhanna-sf: Updated updaters fixtures files

bcoe:
@abhinavkhanna-sf this is looking good 😄 what was the bug leading to fields missing in the fixture?

----

Let me know if you're having any trouble writing the tests for the releaser, and I can try to pitch in (_it might be a few days before I have cycles though_).

Thanks for your work adding this.

abhinavkhanna-sf:
@bcoe No bug for the missing fields. there were two Chart.yaml files with different data. One in releaser/fixtures and other one in updater/fixtures. Still two files but both with same data. Let me know if this needs to be changed.

Let me try writing test cases for releaser. Need to find some time to understand and do it. I will let you know if I need help.

Thanks for your help and reviewing the PR.

bcoe:
@abhinavkhanna-sf #763 is a signifiant refactor, and I'd love to ideally land this first.

If you haven't had time to write the additional tests yet, perhaps  we could just get this initial work merged and follow up with the tests?

bcoe:
FYI, tests are currently failing because `latestTag` was moved into the release class from the GitHub class.

bcoe (approved):
@abhinavkhanna-sf I'm going ahead and landing this, just so you don't get bitten by a fairly major refactor that @joeldodge79 is taking on.

I'd love it if we could follow on with tests, so that we can push coverage back up.

Thanks for the contribution .

PR #752, opened 2021-02-06
refactor: commit-split enhancements

joeldodge79:
The aggregate releaser will get all commits on the default branch and
use this function to partition them per configured package path to send
to individual releasers.

- packagePaths: restrict splits to specific paths
- includeEmpty: include `git commit --allowEmtpy` commits. Each path gets
  a copy.

default behavior is the original: split by 1st child dir, drop empty
commits.

### DIFF
--- a/src/commit-split.ts
+++ b/src/commit-split.ts
@@ -15,31 +15,66 @@
 import {Commit} from './graphql-to-commits';
 import {relative} from 'path';
 
-interface CommitSplitOptions {
+export interface CommitSplitOptions {
+  // Defaults to './'
   root?: string;
+  // Include empty git commits: each empty commit is included
+  // in the list of commits for each path.
+  includeEmpty?: boolean;
+  // rather than split by top level folder, split by these package
+  // paths (e.g. ["packages/pkg1", "python/pkg1"]). Commits that
+  // only touch files under paths not specified here are ignored.
+  packagePaths?: string[];
 }
 
 export class CommitSplit {
   root: string;
+  includeEmpty: boolean;
+  packagePaths?: string[];
   constructor(opts?: CommitSplitOptions) {
     opts = opts || {};
     this.root = opts.root || './';
+    this.includeEmpty = !!opts.includeEmpty;
+    if (opts.packagePaths) {
+      const paths: string[] = [];
+      for (let newPath of opts.packagePaths) {
+        newPath = newPath.replace(/[/\\]$/, '');
+        for (const exPath of paths) {
+          if (newPath.indexOf(exPath) >= 0 || exPath.indexOf(newPath) >= 0) {
+            throw new Error(
+              `Path prefixes must be unique: ${newPath}, ${exPath}`
+            );
+          }
+        }
+        paths.push(newPath);
+      }
+      this.packagePaths = paths;
+    }
   }
+
   split(commits: Commit[]): {[key: string]: Commit[]} {
     const splitCommits: {[key: string]: Commit[]} = {};
     commits.forEach((commit: Commit) => {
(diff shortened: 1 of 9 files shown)


src/commit-split.ts
```diff
@@ -15,32 +15,67 @@
 import {Commit} from './graphql-to-commits';
 import {relative} from 'path';
 
-interface CommitSplitOptions {
+export interface CommitSplitOptions {
+  // Defaults to './'
   root?: string;
+  // Include empty git commits: each empty commit is included
+  // in the list of commits for each path.
+  includeEmpty?: boolean;
```
bcoe, line 23: Curious what workflow you use includeEmpty for, I'm guessing to force a release for all packages?

src/commit-split.ts
```diff
@@ -15,32 +15,67 @@
...
-        const splitPath = relative(this.root, file).split(/[/\\]/);
+        const path = relative(this.root, file);
+        const splitPath = path.split(/[/\\]/);
         // indicates that we have a top-level file and not a folder
         // in this edge-case we should not attempt to update the path.
         if (splitPath.length === 1) continue;
-        const pkgName = splitPath[0];
-        if (dedupe.has(pkgName)) continue;
+        const pkgName = this.packagePaths
+          ? this.packagePaths.find(p => path.indexOf(p) >= 0)
```
bcoe, line 67: I might add a comment to this line, pointing out that "if packagePaths has been set, only track paths within this set".

test/commit-split.ts
```diff
@@ -29,12 +31,134 @@ describe('CommitSplit', () => {
...
+        path.resolve(fixturesPath, 'commits-yoshi-php-monorepo.json'),
         'utf8'
       )
     );
     const commits: Commit[] = (await graphqlToCommits(github, graphql)).commits;
     const cs = new CommitSplit();
     snapshot(cs.split(commits));
   });
+
+  type ExpectedCommitSplit = Record<string, Commit[]>;
```
bcoe, line 43: I'd toss a comment above the two helpers, and above the actual setup for the assertions, pointing out:

1. what state each helper helps configure.
2. what methods we're testing with the assertion.

bcoe (changes requested):
Left nits, this is looking good to me.

src/commit-split.ts
joeldodge79: yeah, I'd like to extend the existing functionality (using `--allow-empty` to control the version) into the aggregate*. My personal use-case would mostly be in conjunction with the future `release-<package/path>-as` footer functionality to set versions per package:

```
chore: set version 21.2.0

release-pkgA-as: 21.2.0
release-pkgB-as: 21.2.0
// the remainder just semver bump according to Conventional Commits
```

Of course you could also auto bump all packages with
```
feat: bump minor on all packages
```

Or coerce all packages to a version with
```
chore: set all to version 7.7.7

release-as: 7.7.7
```

I'll add to the comments to that effect here.


*I haven't tested but I'm pretty sure `monorepo-tags` configuration isn't currently compatible with this workflow. Probably worth making consistent at some point?

joeldodge79:
I'm having a hard time understanding how these changes reduce each releasers' coverage. halp?

test/commit-split.ts
```diff
@@ -144,10 +182,14 @@ describe('CommitSplit', () => {
     });
   }
 
+  // Test invalid CommitSplitOptions.packagePaths combinations.
+  // Intentionally inconsistent trailing slashes to test path normalization.
   const invaldPaths = [
```
bcoe, line 187: whoops, missed a typo -> `invalidPaths`?

bcoe (approved):

PR #757, opened 2021-02-10
fix(github-release): release name is packageName

joeldodge79:
Forming the contents of a release was factored out of GitHubRelease in
https://github.com/googleapis/release-please/pull/720
The argument provide for the release name was accidentally swapped to be
the packagePrefix (e.g. sans `@scope/` for node packages).


test/github-release.ts
```diff
@@ -283,8 +289,15 @@ describe('GitHubRelease', () => {
...
       sandbox
         .stub(release.gh, 'createRelease')
-        .withArgs('foo', 'foo-v1.0.3', 'abc123', '\n* entry', false)
+        .withArgs(
+          '@google-cloud/foo',
+          'foo-v1.0.3',
+          'abc123',
+          '\n* entry',
+          false
+        )
```
joeldodge79, line 298: A few times now this `withArgs` has caused me to scratch my head for a minute when the code path stops sending the exact args. The result is `undefined`. Reading the docs you'd think that not matching the args would fall through to the original implementation but that's not the case. I know this is a "stub" as opposed to a "mock" so it shouldn't really be in the business of verifying that it's used properly. However, the mis-matched `withArgs` behavior results in confusing failures: `"undefined has no property html_url"` so... what to folks think about adding a `rejects` or `throwsArg(0)` to any stub that we use `withArgs` on in order to make diagnosing test failures easier?

It's done once here on L288 above. I suggest we try to always add it (except in cases where we need the stub to [reject calls mimicking a 403 or 404 from the GH api](https://github.com/googleapis/release-please/blob/2405916a7aa9a4b16734c0b0c3a81748df2815e3/test/releasers/node.ts#L98)) Thoughts?

joeldodge79 (commented):
some explanatory comments inline

joeldodge79:
cc @chingor13

test/github-release.ts
chingor13: I agree, it'd be nice if it threw on an unexpected input which we do similar to the get file stubs. We can grab the stub and have the fallthrough case reject, but it's more verbose to write in the test.

chingor13 (approved):

bcoe (approved):

PR #759, opened 2021-02-10
feat: add async interator for searching commit history

chingor13:
This functionality should be covered by the tests for `commitsSince` and `findMergeCommit` which were refactored to use the async iterator.


bcoe (approved):
Like this refactor 👍

PR #762, opened 2021-02-11
refactor: squash helper method into ReleasePR.latestTag

chingor13:
Also removes extra unused filter criteria (labels). When searching for latestTag, we don't care about the PR labels so remove that filter check.

### DIFF
--- a/src/release-pr.ts
+++ b/src/release-pr.ts
@@ -504,61 +504,16 @@ export class ReleasePR {
     prefix?: string,
     preRelease = false
   ): Promise<GitHubTag | undefined> {
+    const branchPrefix = prefix?.endsWith('-')
+      ? prefix.replace(/-$/, '')
+      : prefix;
     // only look at the last 250 or so commits to find the latest tag - we
     // don't want to scan the entire repository history if this repo has never
     // been released
-    const pull = await this.findMergedReleasePR([], prefix, preRelease, 250);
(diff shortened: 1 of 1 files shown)


src/release-pr.ts
```diff
@@ -504,64 +504,19 @@ export class ReleasePR {
...
         continue;
       }
 
-      // If labels specified, ensure the pull request has all the specified labels
-      if (
-        labels.length > 0 &&
-        !this.hasAllLabels(labels, mergedPullRequest.labels)
-      ) {
-        continue;
-      }
```
chingor13, line 563: When called in this helper, labels were always empty.

src/release-pr.ts
```diff
@@ -504,64 +504,19 @@ export class ReleasePR {
...
     // only look at the last 250 or so commits to find the latest tag - we
     // don't want to scan the entire repository history if this repo has never
     // been released
-    const pull = await this.findMergedReleasePR([], prefix, preRelease, 250);
-    if (!pull) return await this.gh.latestTagFallback(prefix, preRelease);
-
-    // TODO: shouldn't need to do this twice
-    const branchName = BranchName.parse(pull.headRefName)!;
-    const version = await this.detectReleaseVersion(pull, branchName);
-    const normalizedVersion = semver.valid(version)!;
```
chingor13, line 516: This is now no longer necessary with the async iterator - don't need to return the pull request and parse this information again.

bcoe (approved):

PR #763, opened 2021-02-11
refactor!: normalize configuration across classes

joeldodge79:
This is a first take at normalizing the configuration options used to
construct GitHub, ReleasePR, and GitHubRelease instances as well as
higher level configuration used by the factory entrypoints to build
them. The principal motivation is to bring the code more inline with
IoC/DI patterns (https://stackoverflow.com/a/3140/13802304)

refactor!: interfaces for constructor options moved to src/index.ts
and signficantly reorganized
refactor!: GitHub constructor: GitHubConstructorOptions
refactor!: GitHubRelease constructor: GitHubReleaseConstructorOptions
refactor!: ReleasePR constructor: ReleasePRConstructorOptions
refactor!: ReleasePR package name handling refactored -
lookupPackageName removed in favor of getPackageName
refactor!: ConventionalCommits constructor takes owner and repo instead
of just repoUrl
refactor!: remove GitHubRelease.labels in favor of ReleasePR.labels

### DIFF
--- a/src/bin/release-please.ts
+++ b/src/bin/release-please.ts
@@ -16,11 +16,10 @@
 
 import chalk = require('chalk');
 import {coerceOption} from '../util/coerce-option';
-import {GitHubReleaseOptions} from '../github-release';
-import {ReleasePROptions} from '../release-pr';
 import {factory} from '../factory';
 import {getReleaserTypes, ReleaseType} from '../releasers';
 import * as yargs from 'yargs';
+import {GitHubReleaseFactoryOptions, ReleasePRFactoryOptions} from '..';
 
 interface ErrorObject {
   body?: object;
@@ -41,46 +40,52 @@ interface YargsOptionsBuilder {
   option(opt: string, options: YargsOptions): YargsOptionsBuilder;
 }
 
+function releaseType(ya: YargsOptionsBuilder, defaultType?: string) {
+  const relTypeOptions: YargsOptions = {
+    describe: 'what type of repo is a release being created for?',
+    choices: getReleaserTypes(),
+  };
+  if (defaultType) {
+    relTypeOptions.default = defaultType;
+  }
+  ya.option('release-type', relTypeOptions);
+}
+
 export const parser = yargs
   .command(
     'release-pr',
     'create or update a PR representing the next release',
+    // options unique to ReleasePR
     (yargs: YargsOptionsBuilder) => {
-      yargs
-        .option('version-file', {
-          describe: 'path to version file to update, e.g., version.rb',
-        })
-        .option('last-package-version', {
-          describe: 'last version # that package was released as',
-        })
-        .option('fork', {
-          describe: 'should the PR be created from a fork',
-          type: 'boolean',
-          default: false,
-        })
-        .option('snapshot', {
-          describe: 'is it a snapshot (or pre-release) being generated?',
-          type: 'boolean',
-          default: false,
-        })
-        .option('default-branch', {
-          describe: 'default branch to open release PR against',
-          type: 'string',
-        })
-        .option('monorepo-tags', {
-          describe: 'include library name in tags and release branches',
-          type: 'boolean',
-          default: false,
-        });
+      releaseType(yargs, 'node');
     },
-    (argv: ReleasePROptions) => {
+    (argv: ReleasePRFactoryOptions) => {
       factory.runCommand('release-pr', argv).catch(handleError);
+    }
+  )
+  .command(
+    'latest-tag',
+    'find the sha of the latest release',
+    // options unique to ReleasePR
+    (yargs: YargsOptionsBuilder) => {
+      releaseType(yargs, 'node');
+    },
+    (argv: ReleasePRFactoryOptions) => {
+      const releasePR = factory.releasePR(argv);
+      releasePR
+        .latestTag()
+        .catch(handleError)
+        .then(latestTag => {
+          console.log(latestTag);
+        });
     }
   )
   .command(
     'github-release',
     'create a GitHub release from a release PR',
+    // options unique to GitHubRelease
     (yargs: YargsOptionsBuilder) => {
+      releaseType(yargs);
       yargs
         .option('changelog-path', {
           default: 'CHANGELOG.md',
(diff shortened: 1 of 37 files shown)


joeldodge79:
found some holes in `src/factory.ts` while manually testing the cli against some example repos I have. Now that the factory has much more responsibility for correctly constructing different pieces it should have much better testing. working on that.

src/bin/release-please.ts
```diff
@@ -133,7 +123,7 @@ export const parser = yargs
     type: 'boolean',
   })
   .option('default-branch', {
-    describe: '',
+    describe: 'The branch to open release PRs against and tag releases on',
```
bcoe, line 126: Good catch.

src/factory.ts
```diff
@@ -37,22 +40,91 @@ function runCommand(
...
+    releaseAs,
+    snapshot,
+    monorepoTags,
+    fork,
+    changelogSections,
+    lastPackageVersion,
+    versionFile,
+  });
+
+  return new GitHubRelease({github, releasePR, ...remaining});
```
bcoe, line 95: I like this approach, this ends up meaning that the GitHub release can use any releaser specific methods on `releasePR`?

src/index.ts
```diff
@@ -12,25 +12,100 @@
...
+  versionFile?: string;
+}
+
+// GitHub Ctor options
+export interface GitHubCtorOptions extends GitHubOptions {
+  owner: string;
+  repo: string;
+}
+
+// Used by GitHubRelease and ReleasePR Ctor
```
bcoe, line 64: nit: might spell this out constructor, had to Google the shorthand.

src/releasers/go-yoshi.ts
```diff
@@ -95,9 +92,10 @@ export class GoYoshi extends ReleasePR {
     });
     const cc = new ConventionalCommits({
       commits: commits,
-      githubRepoUrl: this.repoUrl,
+      owner: this.gh.owner,
```
bcoe, line 95: Just standardizing on splitting own `owner` and `repo` in more places?

src/releasers/node.ts
```diff
@@ -119,9 +117,28 @@ export class Node extends ReleasePR {
...
+    if (!this.pkgJsonContents) {
+      this.pkgJsonContents = await this.getPkgJsonContent();
+    }
+    if (!this._packageName) {
+      const pkg = JSON.parse(this.pkgJsonContents.parsedContent);
+      this.packageName = this._packageName = pkg.name ?? this.packageName;
+    }
+    return {
+      name: this.packageName,
+      refSafeName: () =>
```
bcoe, line 131: If I'm understanding `refSafeName` properly, it codifies dropping the `@pkg/` prefix from a branch name?

bcoe (commented):
Just a couple initial questions, will give a more thorough review later today.

In general I really like the approach, it seems like you've gone the route that: rather than pulling more functionality into ReleasePR, so that it's available when creating a release, you now have an instance of a ReleasePR in a GithubRelease?

This seems really smart to me.

src/factory.ts
joeldodge79: `GitHubRelease` [already relies on a `ReleasePR`](https://github.com/googleapis/release-please/blob/b277b8909e78047797796c0428f545ce482a0b2a/src/github-release.ts#L85) instance for to perform it's one job (`run`) - this change is simple removing the knowledge of how to construct it (and thus also removing any duplicate constructor options) and centralizing it in case other future classes (aka aggregate) need one too. Same pattern as pulling out the building of a `GitHub` instance because that was already being duplicated (by `ReleasePR`)

src/index.ts
joeldodge79: will do. I had to look it up too not that long ago (finally got tired of seeing it everywhere on SO and not really knowing what it mean). I come from python where `init` is both short and accurate :-)

src/releasers/go-yoshi.ts
joeldodge79: the idea is to consider `repoUrl` as denormalized shorthand user input for `factory.ts` and that it should be normalized only once into `{owner, repo}` and everything else inside the release-please codebase should rely on those directly rather than either carry around this copy of the short hand that everyone then has to convert. It would be a different story if different actors converted it differently but they all just use `parse-github-repo-url` to get the owner and repo.

src/releasers/node.ts
joeldodge79: yeah, but I think I'll change it to `getComponent()` per a discussion with @chingor13

> In this refactor I tried to steer away as much as possible from functionality refactoring or re-naming. The one exception is that darned quagmire of (packagePrefix vs packageName) x (set via configuration vs releaser knows how to get it): there was a nasty tangle of responsibility leaking that I needed to cleanup to normalize the configuration. Maybe `PackageName.refSafeName()` is better named `PackageName.getComponent()` - that's more inline with the domain vocabulary we've chosen and being safe to use as/in a git ref is just a requirement of a component rather than the name

src/bin/release-please.ts
```diff
@@ -41,46 +40,52 @@ interface YargsOptionsBuilder {
   option(opt: string, options: YargsOptions): YargsOptionsBuilder;
 }
 
+function releaseType(ya: YargsOptionsBuilder, defaultType?: string) {
+  const relTypeOptions: YargsOptions = {
+    describe: 'what type of repo is a release being created for?',
+    choices: getReleaserTypes(),
+  };
+  if (defaultType) {
+    relTypeOptions['default'] = defaultType;
```
bcoe, line 49: nit: `relTypeOptions.default`.

src/factory.ts
```diff
@@ -37,22 +40,91 @@ function runCommand(
...
+    packageName,
+    bumpMinorPreMajor,
+    releaseAs,
+    snapshot,
+    monorepoTags,
+    fork,
+    changelogSections,
+    lastPackageVersion,
+    versionFile,
+    ...remaining
```
bcoe, line 70: I seem to be missing where `remaining` is defined?

src/release-pr.ts
```diff
@@ -88,60 +70,65 @@ export interface OpenPROptions {
...
-  // Prefix for tags/branches
-  packagePrefix: string;
+  changelogSections?: ChangelogSection[];
 
-  constructor(options: ReleasePROptions) {
+  constructor(options: ReleasePRConstructorOptions) {
     this.bumpMinorPreMajor = options.bumpMinorPreMajor || false;
-    this.defaultBranch = options.defaultBranch;
     this.fork = !!options.fork;
     this.labels = options.label
```
bcoe, line 88: Feels a bit weird that `label` starts as a string, that's split, and then we pass it around as an array of strings to the GitHub instance.

We could coerce labels into an array in the CLI, using `.coerce` and in the GitHub action, and perhaps make labels an array from the factory onwards?

src/releasers/node.ts
```diff
@@ -27,25 +32,17 @@ import {PackageJson} from '../updaters/package-json';
 import {SamplesPackageJson} from '../updaters/samples-package-json';
 
 export class Node extends ReleasePR {
-  async getOpenPROptions(
+  private pkgJsonContents?: GitHubFileContents;
```
bcoe, line 35: Is this just an optimization so that we only load once?

test/factory.ts
```diff
@@ -25,16 +27,57 @@ describe('factory', () => {
     sandbox.restore();
   });
 
+  describe('gitHubInstance', () => {
+    const owner = 'google';
+    const repo = 'cloud';
+    const repoUrls = [
+      `${owner}/${repo}`,
+      `${owner}/${repo}`,
+      `https://github.com/${owner}/${repo}.git`,
```
bcoe, line 36: I appreciate these tests, and that we're so forgiving with the input.

bcoe (approved):
One question for you, was scratching my head trying to figure out where `remaining` is assigned (maybe I'm misunderstanding something magic about spread operators).

It also jumped out at me that perhaps `label` should only be a string during parsing.

src/bin/release-please.ts
joeldodge79: ug, brain frazzle (when I get tired the python slips back in)

src/factory.ts
joeldodge79: I can't hyperlink to the exact location but search for the string "Rest in Object Destructuring " in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

in this case `remaining` is whatever is left of the `options` parameter that wasn't pulled out in the destructuring.

src/release-pr.ts
joeldodge79: I like it. Do you want me to do that in this PR?

src/releasers/node.ts
joeldodge79: yeah - the original `_run` code loads the contents at the beginning only once and passes it to the `PackageJson` Updater as well as sets `self.packageName` - but now that I've tried to make `getPackageName()` available potentially prior to any "run" I thought I'd cache it here. But maybe potential caching bugs aren't worth the small cost of an extra API call or two?

test/factory.ts
joeldodge79: oops, I have the same input listed twice. I was just trying to copy some examples from https://github.com/repo-utils/parse-github-repo-url#features

src/release-pr.ts
joeldodge79: I'm doing it in this PR - and it's leading to another `GitHubRelease` option culling: that class doesn't need `this.labels` now that it has a `releasePR` instance. I hadn't looked closely but it turns out ReleasePR.labels is supposed be the same as GitHubRelease.labels. Back before GitHubRelease had a releaser instance that duplication was necessary, but not anymore

src/factory.ts
bcoe: neat, haven't used that pattern.

src/factory.ts
```diff
@@ -37,22 +41,106 @@ function runCommand(
...
+  } = options;
+  const github = gitHubInstance({
+    repoUrl,
+    defaultBranch,
+    token,
+    apiUrl,
+    octokitAPIs,
+  });
+
+  const labels = getLabels(label);
```
bcoe, line 126: Nice, I like this approach.

bcoe (approved):

PR #764, opened 2021-02-11
fix: find first associate pull request

### LINKED ISSUE #756: latestTag() does not find stable-branch name without a version in the branch name
`latestTag()` assumes that the version can be parsed from the branch name which is not possible when using stable-branch names

We will need to improve/relax the logic for how we find the latest tag.

chingor13:
For some reason, a commit can have multiple associated pull requests. We actually want the first result, not the last and this appears to correctly be the associated pull request that we want.

Fixes #756


bcoe (approved):
Let's see if this works.

PR #767, opened 2021-02-12
refactor!: rename factory.run -> call

joeldodge79:
now that there's an entry point other than `run()` on an instance
(i.e. `ReleasePR.latestTag()`), `runCommand()` needs to be able to
specify what it is. `call()` (formerly `run()`) now accepts this `method`
argument.

Using function overloads as a means to keep a clean call type signature.

### DIFF
--- a/src/bin/release-please.ts
+++ b/src/bin/release-please.ts
@@ -75,7 +75,12 @@ export const parser = yargs
         .runCommand('latest-tag', argv)
         .catch(handleError)
         .then(latestTag => {
-          if (latestTag) console.log(latestTag);
+          if (latestTag) {
+            console.log(latestTag);
+          } else {
+            console.log('No latest tag found.');
+            process.exitCode = 1;
+          }
         });
     }
   )
@@ -191,8 +196,10 @@ interface HandleError {
 // the request object, don't do this in CI/CD).
 export const handleError: HandleError = (err: ErrorObject) => {
   let status = '';
-  if (!handleError.yargsArgs) {
-    return;
+  if (handleError.yargsArgs === undefined) {
+    throw new Error(
+      'Set handleError.yargsArgs with a yargs.Arguments instance.'
+    );
   }
   if (!handleError.logger) {
     handleError.logger = console;
--- a/src/bin/release-please.ts
+++ b/src/bin/release-please.ts
@@ -228,12 +228,11 @@ let argv: yargs.Arguments;
 if (require.main === module) {
   argv = parser.parse();
   handleError.yargsArgs = argv;
+  process.on('unhandledRejection', err => {
+    handleError(err as ErrorObject);
+  });
+
+  process.on('uncaughtException', err => {
+    handleError(err as ErrorObject);
+  });
 }
-
-process.on('unhandledRejection', err => {
-  handleError(err as ErrorObject);
-});
-
-process.on('uncaughtException', err => {
-  handleError(err as ErrorObject);
-});
--- a/src/bin/release-please.ts
+++ b/src/bin/release-please.ts
@@ -71,12 +71,11 @@ export const parser = yargs
       releaseType(yargs, 'node');
     },
     (argv: ReleasePRFactoryOptions) => {
-      const releasePR = factory.releasePR(argv);
-      releasePR
-        .latestTag()
+      factory
+        .runCommand('latest-tag', argv)
         .catch(handleError)
         .then(latestTag => {
-          console.log(latestTag);
+          if (latestTag) console.log(latestTag);
         });
     }
   )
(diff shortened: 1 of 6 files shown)


src/bin/release-please.ts
```diff
@@ -71,12 +71,11 @@ export const parser = yargs
...
     (argv: ReleasePRFactoryOptions) => {
-      const releasePR = factory.releasePR(argv);
-      releasePR
-        .latestTag()
+      factory
+        .runCommand('latest-tag', argv)
         .catch(handleError)
         .then(latestTag => {
-          console.log(latestTag);
+          if (latestTag) console.log(latestTag);
```
bcoe, line 78: maybe in the `else` we could say `no latest tag found`? might also be worth setting `process.exitCode = 1`?

src/factory.ts
```diff
@@ -46,8 +47,17 @@ function isGitHubRelease(
...
+function run(
+  runnable: ReleasePR,
+  cmd: 'run' | 'latestTag'
+): Promise<number | GitHubTag | undefined>;
+function run(
+  runnable: GitHubRelease,
+  cmd: 'run'
+): Promise<ReleaseResponse | undefined>;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function run(runnable: any, cmd: any = 'run') {
```
bcoe, line 59: I wish we could avoid this `any` type, what if we just introduce a Runnable interface, that's something like:

```
interface Runnable {
  run: (): ReleaseResponse | undefined | number
}
```

☝️ which GitHubRelease and GitHubRelease implement? Then your mocks could just implement this.

bcoe (changes requested):
One nit, that doesn't need to be blocking.

src/factory.ts
joeldodge79: I'll spend some time thinking about this. I kind of reached for function overloads as a quick fix to support GitHubRelease.run ReleasePR.run, and now ReleasePR.latestTag inside of this `run` method. Reading the doc examples they use `any` for the actual method implementation. But I can avoid it with a big fat pile of unions and a type assertion in the implementation:

```suggestion
function run(
  runnable: ReleasePR | GitHubRelease,
  cmd: 'run' | 'latestTag'
): Promise<number | GitHubTag | ReleaseResponse | undefined> {
  if (runnable instanceof GitHubRelease) {
    return runnable[cmd as 'run']();  // type assertion less bad than `any`
  }
  return runnable[cmd]();
}
```

It still has the benefit of the type overloads to help enforce correct calling but also has a higher level of type correctness itself, just looks kinda ugly...

But maybe there's a better pattern for accomplishing this single method to dispatch the execution? I'll keep pondering :-)

joeldodge79: ended up refactoring this a bit more. renamed `run` to `call` since that's what this dispatcher is really doing now: calling a method on an instance.

src/factory.ts
```diff
@@ -28,26 +28,98 @@ import {DEFAULT_LABELS} from './constants';
...
+};
+type IsGitHubReleaseCmd = {
+  command: Command;
+  options: GitHubReleaseFactoryOptions;
+};
+
+// shorthand aliases for instance/method call return types
+export type ReleasePRRunResult = Promise<number | GitHubTag | undefined>;
+export type GitHubReleaseRunResult = Promise<GitHubReleaseResponse | undefined>;
+export type RunResult = ReleasePRRunResult | GitHubReleaseRunResult;
```
joeldodge79, line 52: a bit of an explosion of types here... I'm definitely open to a simpler approach. But this is where I got adding support for `latest-tag` while keeping the nice/testable pattern of

CLI <-> translate to instance/method (`runCommand()`) <-> execute (`call()`)

I could see other CLI commands being added too (maybe `validate/list-package-config` for the aggregate) so I think it's a worthwhile endeavor.

src/factory.ts
```diff
@@ -28,26 +28,98 @@ import {DEFAULT_LABELS} from './constants';
...
+  let result: RunResult;
+  const cmdOpts = {command, options};
+  if (isReleasePRCmd(cmdOpts)) {
+    let method: ReleasePRMethod = 'run';
+    if (command === 'latest-tag') {
+      method = 'latestTag';
+    }
+    result = factory.call(releasePR(cmdOpts.options), method);
+  } else if (isGitHubReleaseCmd({command, options})) {
+    result = factory.call(githubRelease(cmdOpts.options), 'run');
```
joeldodge79, line 92: I augmented the `is{GitHubRelease,ReleasePR}Cmd` functions' type predicates specifically to be able to narrow the typing such that `cmdOpts.options` is the correct type going into `releasePR()` and `githubRelease()`

test/cli.ts
```diff
@@ -14,48 +14,100 @@
...
+  method: GitHubReleaseMethod
+): GitHubReleaseRunResult;
+function callStub(
+  instance: ReleasePR | GitHubRelease,
+  _method: Method
+): RunResult {
+  instanceToRun = instance;
+  return Promise.resolve(undefined);
+}
+
```
joeldodge79, line 54: trying to be ruthless about not admitting `any` into the codebase - so I had to duplicate the overload definitions here.

bcoe (approved):

PR #770, opened 2021-02-15
refactor!: move fork down to GitHub

joeldodge79:
fork is a low level GitHub plumbing configuration. GitHub can make the
decision when it will be able or unable to add/remove labels and close
prs.

BREAKING CHANGE: fork is no longer in ReleasePRConstructorOptions, it is
now in GitHubConstructorOptions.

### DIFF
--- a/src/bin/release-please.ts
+++ b/src/bin/release-please.ts
@@ -20,6 +20,7 @@ import {factory} from '../factory';
 import {getReleaserTypes, ReleaseType} from '../releasers';
 import * as yargs from 'yargs';
 import {GitHubReleaseFactoryOptions, ReleasePRFactoryOptions} from '..';
+import {GH_API_URL} from '../constants';
 
 interface ErrorObject {
   body?: object;
@@ -125,12 +126,17 @@ export const parser = yargs
   .option('token', {describe: 'GitHub token with repo write permissions'})
   .option('api-url', {
     describe: 'URL to use when making API requests',
-    default: 'https://api.github.com',
+    default: GH_API_URL,
     type: 'string',
   })
   .option('default-branch', {
     describe: 'The branch to open release PRs against and tag releases on',
     type: 'string',
+  })
+  .option('fork', {
+    describe: 'should the PR be created from a fork',
+    type: 'boolean',
+    default: false,
   })
   .option('repo-url', {
     describe: 'GitHub URL to generate release for',
@@ -169,11 +175,6 @@ export const parser = yargs
   })
   .option('last-package-version', {
     describe: 'last version # that package was released as',
-  })
-  .option('fork', {
-    describe: 'should the PR be created from a fork',
-    type: 'boolean',
-    default: false,
   })
   .option('snapshot', {
     describe: 'is it a snapshot (or pre-release) being generated?',
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -128,20 +128,20 @@ function getLabels(label?: string): string[] {
 
 function githubRelease(options: GitHubReleaseFactoryOptions): GitHubRelease {
   const {
-    label,
     repoUrl,
     defaultBranch,
+    fork,
     token,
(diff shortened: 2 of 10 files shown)


src/bin/release-please.ts
```diff
@@ -125,13 +126,18 @@ export const parser = yargs
...
+    default: GH_API_URL,
     type: 'string',
   })
   .option('default-branch', {
     describe: 'The branch to open release PRs against and tag releases on',
     type: 'string',
   })
+  .option('fork', {
+    describe: 'should the PR be created from a fork',
+    type: 'boolean',
```
bcoe, line 138: Reading through `factory.ts`, It didn't look like `fork` is passed to `githubInstance` in the `releasePR` helper?

Does it still need to be set here so that the branch can be created from a fork?

bcoe (changes requested):
This is looking good 👍

It might be good to open a work in progress PR on `release-please-action` that captures some of these additional breaking changes, such as switching from `run` to `call`.\

We can keep publishing candidate releases of `release-please` as we continue refactoring work.

src/bin/release-please.ts
joeldodge79: good catch, that means I'm missing a test! will fix (and I think there's an even simpler way to pass around Factory config options to the different factory helpers)

src/index.ts
```diff
@@ -85,19 +85,24 @@ interface FactoryOptions {
...
+    ReleaserFactory {
+  label?: string;
+}
 
 // GitHubRelease factory/builder options
 export interface GitHubReleaseFactoryOptions
   extends GitHubReleaseOptions,
     ReleasePROptions,
-    Omit<ReleasePRFactoryOptions, 'releaseType'>,
+    ReleasePRFactoryOptions,
```
joeldodge79, line 105: decided maintaining a required vs optional `releaseType` depending on the factory was too complicated - I think we can on the cli defaulting to 'Node' for `release-pr`. Anyone else who accidentally ends up with a ReleasePR instance when they wanted an actual releaser to run will find out soon enough as `run()` or `getOpenPROptions` will go boom.

src/factory.ts
```diff
@@ -82,19 +82,17 @@ function runCommand(
...
     }
     result = factory.call(releasePR(cmdOpts.options), method);
-  } else if (isGitHubReleaseCmd({command, options})) {
-    result = factory.call(githubRelease(cmdOpts.options), 'run');
   } else {
     throw new Error(
-      `Invalid command(${cmdOpts.command}) with options(${JSON.stringify(
-        cmdOpts.options
-      )})`
+      `Invalid command(${command}) with options(${JSON.stringify(options)})`
```
joeldodge79, line 95: I'm not a huge fan of this complex type checking code - when I made `releaseType` optional on both `{GitHubRelease,ReleasePR}FactoryOptions` it introduced a type error in the original code here saying `cmdOpts` had ["acquired the `never` type"](https://www.typescriptlang.org/docs/handbook/basic-types.html#never) after the `isReleasePRCmd` type predicate/guard check. I'm not exactly sure why, perhaps because now `ReleasePRFactoryOptions` is a subset of `GitHubReleaseFactoryOptions` (it used to have a required `releaseType` property). It hurts my brain a little but it works switching the if/else-if conditionals around

test/factory.ts
```diff
@@ -66,23 +82,40 @@ describe('factory', () => {
     });
   });
   describe('releasePR', () => {
-    it('returns instance of dynamically loaded releaser', async () => {
+    it('returns a ReleasePR with all the things', async () => {
       const releasePR = factory.releasePR({
         repoUrl: 'googleapis/ruby-test-repo',
+        fork: true,
+        token: 'some-token',
```
bcoe, line 89: I like this smoke test 👍 I should probably do something similar in the action.

bcoe (approved):

PR #774, opened 2021-02-17
fix(java): loosen version lookup regex

bcoe:
Loosen regex for release.


chingor13 (changes requested):
Add a quick test for parsing both cases?

chingor13 (approved):
I can send a followup PR

PR #776, opened 2021-02-18
fix(release-pr): update default initial version

### LINKED ISSUE #558: Support initial development phase (aka Major version zero)
I have software in the [initial development phase](https://semver.org/#spec-item-4), and would like to cut releaseses with a major version zero.

If the major component of my current version is `0`, `release-please` could:
* bump the patch component for features and fixes. (example: `0.1.2` -> `0.1.3`)
* bump the minor component for breaking changes. (example: `0.1.2` -> `0.2.0`)
* (as a consequence) never bumps the major component.

Of course there will be a time where I'll eventually want to bump to `1.0.0`. In order to let release-please that I'm ready for the big "one dot oh" there could be a flag for the release-please CLI, and option for the github action.

Use release-please only for the changelog generation and always manually override the version numbers. But this is quite error prone, and maybe not using release-please would be a better solution. :-/

 **Additional context**
I would like to raise two arguments in favor of this:

1. Initial development phase is part of the semver specification: https://semver.org/#spec-item-4
2. In rust, for example, there are explicit quality standards for libraries that may "force" people to release their library with a major version zero (under special conditions): https://rust-lang.github.io/api-guidelines/necessities.html#c-stable

yi-Xu-0100:

Fixes #558 https://github.com/google-github-actions/release-please-action/issues/168 🦕

### DIFF
--- a/src/releasers/node.ts
+++ b/src/releasers/node.ts
@@ -154,4 +154,8 @@ export class Node extends ReleasePR {
     }
     return this.pkgJsonContents;
   }
+
+  protected defaultInitialVersion(): string {
+    return this.bumpMinorPreMajor ? '0.1.0' : '1.0.0';
+  }
 }
--- a/src/releasers/node.ts
+++ b/src/releasers/node.ts
@@ -154,8 +154,4 @@ export class Node extends ReleasePR {
     }
     return this.pkgJsonContents;
   }
-
-  protected defaultInitialVersion(): string {
-    return this.bumpMinorPreMajor ? '0.1.0' : super.defaultInitialVersion();
-  }
 }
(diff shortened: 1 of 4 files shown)


src/releasers/node.ts
```diff
@@ -154,4 +154,8 @@ export class Node extends ReleasePR {
     }
     return this.pkgJsonContents;
   }
+
+  protected defaultInitialVersion(): string {
+    return this.bumpMinorPreMajor ? '0.1.0' : super.defaultInitialVersion();
```
bcoe, line 159: @yi-Xu-0100 I think we can move this into the super class, this behavior should be supported by all languages.

bcoe (approved):

PR #785, opened 2021-02-19
refactor(manifest): code layout, file format

joeldodge79:
The Manifest class uses a manifest.json file to track version info for
one or more packages* in a repo. The structure is an object of this
shape:

`{"<packagePath>": "<version>"}`

The manifest file must exist at least at the tip of the `defaultBranch`
even if it is empty at first. The Manifest class will attempt to find
the last merged Release PR and use the package versions set in the
manifest at that time as the current versions. Failing to find a package
version it will read the manifest at the tip of the `defaultBranch` to
see if a version has been added. "packagePath" additions are the only
change considered (changed or deleted packagePath entries are ignored).
If a version still cannot be determined then the Manifest falls back to
the individual ReleasePR instance's `defaultInitialVersion()`.

Configuration for the Manifest class is split. It needs the basic
options required to create a GitHub instance but then reads a
config.json file from the tip of the `defaultBranch` in the repository
to know what packagePaths it manages and what their specific ReleasePR
instance configuration should be. To start, only a subset of the current
ReleasePR options are supported. Also, top level ReleasePR options can
be set which will be the default values used by any configured packages.

An absurd configuration to demonstrate functionality:
```
{
  // optional package default overrides
  "release-type": "python", // default "node"
  "bump-minor-pre-major": true, // default false

  // default is conventional-changelog-conventionalcommits default
  // preset
  "changelog-sections": [
    {
      type: 'feat',
      section: 'Fancy Features Section'
    }
  ]

  // packages inheret above config but can override
  packages: {
    // path/to/package
    "py/foo": {
      // required for releasers that don't implement src lookup
      "package-name": "foopy",
      "bump-minor-pre-major": false
    },
    "node/pkg1": {
      "release-type": "node",
      "changelog-sections": [
        {
          type: 'feat',
          section: 'Node Features Section'
        },
        {
          type: 'fix',
          section: 'Node Bugs Section'
        }
      ]
    }
  }

}
```

This PR is just a skeleton to consider code layout and the
manifest/config design. The implementation to follow is:

- `pullRequest` - Get all the commits since last release and parcel
them out per package. Pass them along with current version info to a
releaser instance per package and get back the updates needed for a PR.
Update the manifest with new versions and open a PR.

- `githubRelease`: pass the single commit that is the last merged
Release PR into CommitSplit to determine which packages were updated.
Create a Release/Tag for each of these

- connect to factory/cli and profit

\* In the release-please domain a package is a collection of
configuration options ("release-type", "package-name", "path") - the
unique/natural key is "path" which identifies a directory under the
repository root where relevant package metadata files live. All the
source code for this package must also have this path as an ancestor.

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -1,0 +1,176 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {GitHub, GitHubFileContents} from './github';
+import {ReleaseType} from './releasers';
+import {RELEASE_PLEASE_CONFIG, RELEASE_PLEASE_MANIFEST} from './constants';
+import {ManifestConstructorOptions} from '.';
+import {ChangelogSection} from './conventional-commits';
+import {CheckpointType, checkpoint, Checkpoint} from './util/checkpoint';
+import {GitHubReleaseResponse} from './github-release';
+
+interface ReleaserConfig {
+  'release-type'?: ReleaseType;
+  'bump-minor-pre-major'?: boolean;
+  'changelog-sections'?: ChangelogSection[];
+}
+
+interface ReleaserPackageConfig extends ReleaserConfig {
+  'package-name'?: string;
+}
+
+export interface ManifestConfig extends ReleaserConfig {
+  packages: Record<string, ReleaserPackageConfig>;
+}
+
+export class Manifest {
+  gh: GitHub;
+  configFileName: string;
+  manifestFileName: string;
+  checkpoint: Checkpoint;
+  configFile?: GitHubFileContents;
+  headManifest?: GitHubFileContents;
+
+  constructor(options: ManifestConstructorOptions) {
+    this.gh = options.github;
+    this.configFileName = options.configFile || RELEASE_PLEASE_CONFIG;
+    this.manifestFileName = options.manifestFile || RELEASE_PLEASE_MANIFEST;
+    this.checkpoint = options.checkpoint || checkpoint;
+  }
+
+  protected async getFile(
+    fileName: string,
+    sha?: string
+  ): Promise<GitHubFileContents> {
+    let data;
+    try {
+      if (sha) {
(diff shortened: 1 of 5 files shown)


src/manifest.ts
```diff
@@ -0,0 +1,176 @@
...
+  constructor(options: ManifestConstructorOptions) {
+    this.gh = options.github;
+    this.configFileName = options.configFile || RELEASE_PLEASE_CONFIG;
+    this.manifestFileName = options.manifestFile || RELEASE_PLEASE_MANIFEST;
+    this.checkpoint = options.checkpoint || checkpoint;
+  }
+
+  protected async getFile(
+    fileName: string,
+    sha?: string
```
joeldodge79, line 54: This `sha?` parameter and treatment really belong to the next PR but I didn't want to do too much git surgery so I left it in here.

joeldodge79 (commented):
@bcoe I wanted to get a smaller PR out first to handle discussions around config/manifest design and code layout (e.g. `src/manifest.ts` vs something else?)

I already have the `pullRequest` piece implemented [here](https://github.com/googleapis/release-please/compare/master...joeldodge79:aggregate-2-pullRequest) and I'm starting on the `githubRelease` part today

test/manifest.ts
```diff
@@ -0,0 +1,183 @@
...
+import {GitHub} from '../src/github';
+import {stubFilesFromFixtures} from './releasers/utils';
+import {expect} from 'chai';
+import {CheckpointType} from '../src/util/checkpoint';
+
+const fixturePath = './test/fixtures/manifest/repo';
+const defaultBranch = 'main';
+const sandbox = sinon.createSandbox();
+
+describe('Manifest', () => {
```
bcoe, line 28: For another TypeScript project I've been working on, my colleague @SurferJeffAtGoogle defined a schema using JSONSchema for our config format:

https://github.com/googleapis/repo-automation-bots/blob/master/packages/owl-bot/src/owl-bot-yaml-schema.json

and then used [ajv](https://www.npmjs.com/package/ajv) to perform validation. I thought this was a neat approach we might consider, since our manifest/config format is relatively complex.

src/manifest.ts
```diff
@@ -0,0 +1,176 @@
...
+        response = {valid: true, obj: obj};
+      }
+    } catch (e) {
+      this.checkpoint(`Invalid JSON in ${fileName}`, CheckpointType.Failure);
+    }
+    return response;
+  }
+
+  protected async validate(): Promise<boolean> {
+    const configValidation = await this.validateJsonFile<ManifestConfig>(
```
bcoe, line 111: you might be able to replace some of this validation logic with ajv, if we went that route.

bcoe (changes requested):
In general this is looking great to me 😄

bcoe:
Here's where we actually apply the schema:

https://github.com/googleapis/repo-automation-bots/blob/master/packages/owl-bot/src/config-files.ts#L68

bcoe (approved):

PR #786, opened 2021-02-20
refactor: snapshot cleanup and ordering

joeldodge79:
- Order snapshots for easier future diff comparisons per
  https://github.com/bahmutov/snap-shot-it#sorted-snapshots
- use `stringifyExpectedChanges` everywhere
- DRY out some `suggester.createPullrequest` stubbing


joeldodge79:
one more round of suggester.createPullRequest DRY out to do, will re-open

PR #787, opened 2021-02-21
refactor: snapshot cleanup and ordering

joeldodge79:
- Order snapshots for easier future diff comparisons per
  https://github.com/bahmutov/snap-shot-it#sorted-snapshots
- DRY out `suggester.createPullrequest` stubbing so that we have
  consistently formed PR snapshots for changes and metadata.


joeldodge79 (commented):
It looks like a lot is going on in this PR but not really. The main things to check are:

1. All changes happened to test code (except the single char release-pr.ts trailing whitespace fix)
2. The test code changes that happened didn't alter the code paths under test - they just centralize the "snapshot the PR changes" logic

I left inline comments on the `helm` releaser test to illustrate the pattern.

The massive __snapshots__ changes fall into these categories:
- One time alphabetical re-ordering changes for future easier diff reading
- file line length explosion: some tests had taken to snapshotting `JSON.stringify(expectedChanges)` which resulted in each changed file's contents appearing only on a single line, even if it's really up to 700+ lines long (e.g. pom.xml). While github can sometimes do a decent job of highlighting in extra bold the actual character changes on a single line, I think future diff comparison will be much easier when these files'  contents each have their own line.
- additions: I added snapshotting the `suggester.createPullRequest()`'s `options` argument so these now show up for every test as well.

bcoe (approved):

PR #799, opened 2021-02-24
build: ignore .mocharc.js

bcoe:
@joeldodge79 added a couple improvements to `.mocharc.js`, such as an `unhandledRejection` handler, which I'm not quite ready to roll out to all our libraries (since I know there are some codebases where we don't want this behavior, e.g., telemetry).

### DIFF
--- a/synth.py
+++ b/synth.py
@@ -18,5 +18,6 @@ s.copy(templates, excludes=[
   '.nycrc',
   '.kokoro/presubmit/node10/system-test.cfg',
   '.kokoro/continuous/node10/system-test.cfg',
-  '.kokoro/system-test.sh'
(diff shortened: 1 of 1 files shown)


JustinBeckwith (approved):

synth.py
```diff
@@ -18,5 +18,6 @@
   '.nycrc',
   '.kokoro/presubmit/node10/system-test.cfg',
   '.kokoro/continuous/node10/system-test.cfg',
-  '.kokoro/system-test.sh'
+  '.kokoro/system-test.sh',
+  '.mocharc.js'
```
joeldodge79, line 22: @bcoe I'd already done this :-) (see L15 above)

PR #800, opened 2021-02-24
refactor(manifest): pullRequest implementation

joeldodge79:


src/manifest.ts
```diff
@@ -144,12 +249,109 @@ export class Manifest {
...
+
   async pullRequest(): Promise<number | undefined> {
     const valid = await this.validate();
     if (!valid) {
       return;
     }
-    return 1;
+
+    const branchName = (await this.getBranchName()).toString();
+    const lastMergedPR = await this.gh.lastMergedPRByHeadBranch(branchName);
```
bcoe, line 322: what would happen if we accidentally missed on merge, one of our submodules wouldn't get released until another commit touched it?

src/github.ts
```diff
@@ -466,6 +497,77 @@ export class GitHub {
...
+   *
+   * @param {string} headBranch - e.g. "release-please/branches/main"
+   * @returns {MergedGitHubPRWithFiles} - if found, otherwise undefined.
+   */
+  async lastMergedPRByHeadBranch(
+    headBranch: string
+  ): Promise<MergedGitHubPRWithFiles | undefined> {
+    const baseBranch = await this.getDefaultBranch();
+    const response: Repository<PullRequests> = await this.graphqlRequest({
+      query: `query lastMergedPRByHeadBranch($owner: String!, $repo: String!, $baseBranch: String!, $headBranch: String!) {
```
bcoe, line 512: I wish we weren't continuing to collect new GraphQL calls, but this looks solid to me.

bcoe (approved):
This is looking really solid to me. If I'm understanding correctly, you've built out the Manifest logic, but it's not wired up anywhere yet. Will the Manifest class be used in a new releaser you introduce?

One thing that jumps out at me, once we have the manifest approach fully supported, I'd love to see if we can move @chingor13's LTS work towards this -- and potentially delete some logic.

src/manifest.ts
```diff
@@ -28,12 +39,33 @@ interface ReleaserConfig {
...
   packages: Record<string, ReleaserPackageConfig>;
 }
 
+interface Package {
+  path: string;
+  releaseType: ReleaseType;
+  packageName?: string;
+  bumpMinorPreMajor?: boolean;
+  changelogSections?: ChangelogSection[];
+  changelogPath?: string;
```
chingor13, line 55: Are these fields included in the manifest? I still feel the releaser configuration should not go into the manifest. IMO, the manifest should include the package name, last release version, and last release SHA.

The GitHub app config, CLI script, or GitHub action should have the releaser configuration and doesn't need to be duplicated in the manifest.

chingor13 (changes requested):
Can you add the manifest format to this PR description?

src/github.ts
joeldodge79: I could have gone this route in the manifest using existing GitHub functionality:
```typescript
const defaultBranch = await this.gh.getDefaultBranch();
const branchName = (await this.getBranchName()).toString()
const filter: MergedPullRequestFilter = (pr: MergedGitHubPR) => {
  return pr.baseRefName === defaultBranch && pr.headRefName === branchName;
}
const pr = this.gh.findMergedPullRequest(targetBranch, filter) // pages through `findMergedPullRequests()` results
const prWithFiles = this.gh.pullRequestFiles(pr.number);  // in a while loop to make sure we get all the files
```

but it seemed a little clunky when:
1/ given a stable branch name we can filter down to exactly the PR we want in the API call rather than in memory
2/ get everything in one* API call as opposed to a REST call and then a GraphQL call (*well, more than one if there are more than 100 files associated with the PR but that's true either way)

we should definitely take stock at some point of all the ways release-please needs to query the GitHub API to see if we can consolidate.

src/manifest.ts
joeldodge79: 100% agree with you. These fields are not in the manifest (see [this PR's description for details](https://github.com/googleapis/release-please/pull/785#issue-576524915) - the manifest file format is easy to miss because it's only a single line near the top of the description).

`Package` here is an internal interface representing the parsing that happens on the releaser config - a dumbed down version of what yargs does for our cli flags.

src/manifest.ts
joeldodge79: is your question

1. "what would happen if `lastMergedPR` here is `undefined`?" - in this case the releaser falls back to getting version info from `.release-please-manifest.json` at HEAD if it exists and if it doesn't then it falls back to each package releaser's `defaultInitialVersion` method. See [this test](https://github.com/googleapis/release-please/pull/800/files#diff-1c245f2069be81e5dae96c50d3f5bac65a1c14be82e61e7d49f9ac89587f3af6R506)
2. what if a submodule has no user facing commits since the last merged PR? in that case it doesn't show up at all in the release PR - only packages with user facing changes have updates in the PR. See [this test](https://github.com/googleapis/release-please/pull/800/files#diff-1c245f2069be81e5dae96c50d3f5bac65a1c14be82e61e7d49f9ac89587f3af6R386)

joeldodge79:
I'm getting a sense I didn't communicate clearly enough on #785 - `src/manifest.ts` *is the "aggregate" releaser*. I re-branded it because I feel like it's more generic than an "aggregate release PR on monorepo" - it could be configured for a single package repo too and it handles bootstrapping quite nicely.

The wiring up will be methods added the factory to handle new cli commands like `manifest-pr` (calls `Manifest.pullRequest`) and `manifest-release` (calls `Manifest.githubRelease`)

If this feels off, lets meet to discuss.

chingor13 (approved):

test/manifest.ts
```diff
@@ -14,22 +14,564 @@
...
+    }
+    addLabelsStub.throws(() => {
+      const argsArr = addLabelsStub.getCalls().map(c => c.args);
+      const unexpected = argsArr.filter(args => args[0][0] !== labelToAdd);
+      return new Error(
+        'addLabelsStub called with unexpected args: ' +
+          JSON.stringify(unexpected[0])
+      );
+    });
+  }
```
joeldodge79, line 96: @AprilArcus here's what I landed on for the sinon.stub functionality that gives me an error when called with unexpected arguments and tells me what they were. I should probably be using a [mock](https://sinonjs.org/releases/latest/mocks/) but reading the philosophy at the top of that dock made me shy away from it. Specifically

> In general you should have no more than one mock (possibly with several expectations) in a single test.

because my follow on PR has a few more stub/mock thingy's that I want to tell me when they've been used unexpectedly

joeldodge79: haha, and now that I finally wrote that down I think I understand. the "mock" unit here would actually be the `github` instance and the "several expectations" would be expectations I set on methods and how they should be called. I'm gonna give that a spin.

@bcoe If you and @chingor13 are ok with this PR as is, I'll change this "stub abusage" into "mock usage" in the next PR

bcoe:
> I'm getting a sense I didn't communicate clearly enough on #785 - src/manifest.ts is the "aggregate" releaser. I re-branded it because I feel like it's more generic than an "aggregate release PR on monorepo" - it could be configured for a single package repo too and it handles bootstrapping quite nicely.

I think this was my bad forgetting the facts in play. I like the design of it being an additional option for the factory.

src/manifest.ts
bcoe: It seems like we get the last merged PR? was wondering if we happened to miss running for one PR, what problems it would cause -- If I'm understanding the code correctly, it would self correct eventually?

src/github.ts
bcoe: > but it seemed a little clunky

I'm sold on the approach you've used.

src/manifest.ts
joeldodge79: > happened to miss running for one PR

I assume you mean creating github releases? if we went two rounds of

`manifest-pr` -> creates Release PR -> Release PR merged

without running `manifest-release` (Tag/Release) in between: the way I'm building `manifest-release` / `Manifest.githubRelease` currently you'd miss out on those tags/releases. I'll add a note for future functionality to run `manifest-release` with a specific sha or PR number to create releases for.

PR #801, opened 2021-02-24
refactor: commit split only needs / sep

joeldodge79:
All file paths coming back from GitHub GraphQL API calls will have the
`/` separator, no need to handle other platform separators

`root` is unused - remove to avoid the use of path.relative which seems to
normalize any hard-coded `/` separators into path.sep

### DIFF
--- a/src/commit-split.ts
+++ b/src/commit-split.ts
@@ -13,11 +13,8 @@
 // limitations under the License.
 
 import {Commit} from './graphql-to-commits';
-import {relative} from 'path';
 
 export interface CommitSplitOptions {
-  // Defaults to './'
-  root?: string;
   // Include empty git commits: each empty commit is included
   // in the list of commits for each path.
   // This allows using the `git commit --allow-empty` workflow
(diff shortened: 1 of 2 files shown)


src/commit-split.ts
```diff
@@ -71,16 +66,15 @@ export class CommitSplit {
       const dedupe: Set<string> = new Set();
       for (let i = 0; i < commit.files.length; i++) {
         const file: string = commit.files[i];
-        const path = relative(this.root, file);
```
joeldodge79, line 74: @bcoe I believe this usage of `path.relative` was automatically converting path separators. I still don't have access to a windows machine with node to verify but my manifest tests all pass with this change.

joeldodge79: In fact I think we should probably have a lint check to prohibit any `"path"` module usage under `src/` as it will do the wrong thing under windows (since github api always returns `/` separators)

bcoe (approved):

PR #803, opened 2021-02-25
fix(node): PackageJson updater sets this.contents

joeldodge79:
The node releaser passes contents it retrieves ahead of time for
"package.json" into the PackageJson update to avoid an exra call. This
was not being persisted.

Switched node releaser test to use a mock so we have stricter checks on
how GitHub is being called.

### DIFF
--- a/src/github.ts
+++ b/src/github.ts
@@ -1191,10 +1191,10 @@ export class GitHub {
    * @returns {string}
    */
   async getDefaultBranch(): Promise<string> {
-    if (this.defaultBranch) {
-      return this.defaultBranch;
+    if (!this.defaultBranch) {
+      this.defaultBranch = await this.getRepositoryDefaultBranch();
     }
-    return this.getRepositoryDefaultBranch();
+    return this.defaultBranch;
   }
 
   /**
--- a/src/updaters/package-json.ts
+++ b/src/updaters/package-json.ts
@@ -31,6 +31,7 @@ export class PackageJson implements Update {
     this.changelogEntry = options.changelogEntry;
     this.version = options.version;
     this.packageName = options.packageName;
+    this.contents = options.contents;
   }
 
   updateContent(content: string): string {
--- a/test/releasers/node.ts
+++ b/test/releasers/node.ts
@@ -19,26 +19,19 @@ import {GitHub} from '../../src/github';
 import {Node} from '../../src/releasers/node';
 import * as snapshot from 'snap-shot-it';
 import * as sinon from 'sinon';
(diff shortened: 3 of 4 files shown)


src/github.ts
```diff
@@ -1191,10 +1191,10 @@ export class GitHub {
    * @returns {string}
    */
   async getDefaultBranch(): Promise<string> {
-    if (this.defaultBranch) {
-      return this.defaultBranch;
+    if (!this.defaultBranch) {
+      this.defaultBranch = await this.getRepositoryDefaultBranch();
     }
-    return this.getRepositoryDefaultBranch();
+    return this.defaultBranch;
```
joeldodge79, line 1197: snuck this optimization in

src/updaters/package-json.ts
```diff
@@ -31,6 +31,7 @@ export class PackageJson implements Update {
     this.changelogEntry = options.changelogEntry;
     this.version = options.version;
     this.packageName = options.packageName;
+    this.contents = options.contents;
```
joeldodge79, line 34: this is the bugfix

test/releasers/node.ts
```diff
@@ -85,18 +146,25 @@ describe('Node', () => {
...
           });
-          mockRequest(releasePR);
-          // for Node.getPackageName
-          stubFilesToUpdate(releasePR.gh, ['package.json']);
 
           // no latestTag to pass to getOpenPROptions (never found a release)
           // releaser should set defaultInitialVersion
           const openPROptions = await releasePR.getOpenPROptions(COMMITS);
 
+          mock.verify();
```
joeldodge79, line 167: the important thing to check in this test file is that
1/ no assertions changed (other than adding in the `mock.verify()` calls
2/ no changes in __snapshots__/node.js

test/releasers/node.ts
```diff
@@ -52,21 +45,89 @@ const COMMITS = [
...
+    .atMost(2)
+    .resolves([]);
+  mock.expects('findMergedReleasePR').atMost(1).resolves(undefined);
+  let headRefName = 'release-';
+  if (branchComponent) {
+    headRefName += branchComponent;
+  }
+  headRefName += '-v0.123.4';
+  mock
+    .expects('mergeCommitIterator')
```
bcoe, line 71: I appreciate that we're getting better at asserting against inputs 👍

test/releasers/node.ts
bcoe: so this serves the same purpose as `.done()` when using nock 👍 

thanks for these test improvements.

bcoe (approved):

PR #806, opened 2021-02-28
fix: commit-split paths validation

joeldodge79:
discovered this bug that prevented configuring with paths like:
- path/path1
- path/path1-extended

went a little heavy on making it clear to future editors why we're using
hardcoded `/` path separators

### DIFF
--- a/test/commit-split.ts
+++ b/test/commit-split.ts
@@ -182,6 +182,21 @@ describe('CommitSplit', () => {
     });
   }
 
+  // Test valid CommitSplitOptions.packagePaths combinations.
+  // Intentionally inconsistent trailing slashes to test path normalization.
+  it('validates configured paths on / separator', () => {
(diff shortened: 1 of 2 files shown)


test/commit-split.ts
```diff
@@ -182,6 +182,21 @@ describe('CommitSplit', () => {
     });
   }
 
+  // Test valid CommitSplitOptions.packagePaths combinations.
+  // Intentionally inconsistent trailing slashes to test path normalization.
+  it('validates configured paths on / separator', () => {
+    const cs = new CommitSplit({
+      packagePaths: ['two-three', 'one-two', 'three/', 'one', 'one-two-three'],
```
bcoe, line 189: Would it be worth adding a `foo/bar` test case, since you mention this was one of the problems in issue?

bcoe (approved):

PR #807, opened 2021-03-01
refactor(github-release): prep for manifest

joeldodge79:
Manifest.githubRelease will find the last merged release PR, get each
package's version, and call
GitHubRelease.createRelease(version, mergedPR) for each package.

Also refactored test/github-release.pr to use mocks.

PR #809, opened 2021-03-03
refactor(manifest): githubRelease implementation

joeldodge79:


test/manifest.ts
```diff
@@ -829,6 +850,674 @@ describe('Manifest', () => {
...
+        },
+      };
+      const config = JSON.stringify(manifestConfig);
+
+      const github = new GitHub({
+        owner: 'fake',
+        repo: 'repo',
+        defaultBranch,
+      });
+      const mock = mockGithub({
```
joeldodge79, line 877: @bcoe I'd already adopted this "mock all the GitHub things in one method" approach prior to your feedback in #807 about splitting it up. At this point would you be ok with a follow-on PR to refactor it (I'd refactor the existing `pullRequest` describe as well)? I'll definitely take care of it, I just have a few more changes already in the pipe that would be easier to get through and then come back to refactoring this.

src/manifest.ts
```diff
@@ -463,26 +476,123 @@ export class Manifest {
...
+        // multiple `draft` releases can be created with the exact same inputs.
+        // (It's a tad confusing because `tag_name` still comes back populated
+        // in these calls but the tag doesn't actually exist).
+        // A draft release can even be created with a `tag_name` referring to an
+        // existing tag referenced by another release.
+        // However, GitHub will prevent "publishing" any draft release that
+        // would cause a duplicate tag to be created. release-please manifest
+        // users specifying the "release-draft" option could run into this
+        // duplicate releases scenario. It's easy enough to just delete the
+        // duplicate draft entries in the UI (or API).
```
joeldodge79, line 551: I spent more time just re-writing this comment. And as I did so, it occurred to me that maybe I should be checking for existing releases prior to creating, rather than relying on the API to `"already_exists"`. That would avoid this potential "duplicate draft releases" issue all together. However, it's kind of an edge case and not hard to manually clean up from. thoughts?

joeldodge79 (commented):
This PR introduces the manifest releaser's ability to create releases/tags for packages that were released in the prior release PR (though there is still no entry point to the manifest releaser via the cli or factory yet, that's a couple PRs down the pipe)

src/manifest.ts
```diff
@@ -463,26 +476,123 @@ export class Manifest {
     return pr;
   }
 
-  async githubRelease(): Promise<GitHubReleaseResponse[] | undefined> {
+  async githubRelease(): Promise<
```
bcoe, line 479: Am I reading this correctly that, unlike the current approach where we look specifically for merged pull requests that have `autorelease: pending`, `githubRelease()` simply pulls the most recently merged pull request, uses this manifest as a starting point, and performs the release logic?

What would trigger `githubRelease()`?

src/manifest.ts
bcoe: I'd start with just adding a `TODO` comment, that suggests the eventual improvement.

test/manifest.ts
```diff
@@ -829,6 +850,674 @@ describe('Manifest', () => {
     });
   });
 
+  describe('githubRelease', () => {
+    it('python draft and node package', async () => {
```
bcoe, line 854: nit: mind making these read a bit better, e.g.,

`it('handles node with draft false python with draft true`)`

I realize I'm being a little pedantic.

test/manifest.ts
bcoe: Follow on PR is absolutely fine 👍

bcoe (changes requested):
Just some nits about the test descriptions, I like to keep them as readable as possible as a sentence -- since it ends up being valuable in the test output.

---

Another question comes to mind though, `lastMergedPRByHeadBranch` will return any merged PR will it not? not just merged release PRs?

Should we be filtering to the last merged PR that has labels indicating that it's a release PR?

src/manifest.ts
joeldodge79: `githubRelease` will be called by `manifest-release` cli cmd.

Good catch, I should check the label! The way it's currently setup yes: it will only get the very latest merged "release PR"* since `lastMergedPRByHeadBranch` returns the most recent PR merged to `defaultBranch` from [`getBranchName()`](https://github.com/googleapis/release-please/blob/c24d0414edde8d6011d4b046d189002d36739277/src/manifest.ts#L94) regardless of tags. So multiple repeated runs will just keep getting the last merged PR and trying to re-create the tags/releases. I'll definitely add checking for the `auto-release:pending` to prevent that in this PR.

*I suppose there's a small hole here: if, prior to release-please creating the PR branch,  someone were to manually create a branch with exactly the same name as [`getBranchName()`](https://github.com/googleapis/release-please/blob/c24d0414edde8d6011d4b046d189002d36739277/src/manifest.ts#L94) (e.g. `release-please/branches/main`) and merge some random PR from it before release-please comes and clobbers their stuff by force-pushing the branch, then we'd have a non-release-please created PR turning up in our query. I think the solution is the same as above: just add checking for `auto-release:pending` will safeguard against this edge case. If someone manually selects our branch name for their random branch AND applies the `auto-release:pending` tag, then that's on them!

This "accidental" branch name shadowing case as well as the case you mentioned before of "forgetting to do a `manifest-release` run between release PR merges" leads me to think we need to consider iterating back through merged release PRs till we find one that is `auto-release:tagged` and then roll forward creating release/tags per release PR we found since the last `auto-release:tagged` PR that have `auto-release:pending`. Would you be ok with adding this robustness in a later PR?

bcoe: > Would you be ok with adding this robustness in a later PR?

For sure, I think in practice we'll find going to the last release PR is sufficient, since we'll eventually have the behavior wired up to GitHub actions that fire within a few moments of a merge.

I think we're on the same page that we were just missing the label check 😄

bcoe (approved):
I think, in practice, this will tend to work because we'll fire a GitHub action whenever a PR is merged -- so should capture all the merged Release PRs as they land.

What we did historically was just look back the last 50 merged PRs, looking for PRs that still had the `autorelease: pending` label. I think an approach like this would be more robust, but I'm open to doing this in a follow up.

PR #810, opened 2021-03-04
feat(java): add LTS versioning scheme and releaser

chingor13:
Adds a custom versioning scheme for Java.

To support service pack/LTS releases, we use a separate versioning scheme. The first LTS release after a mainline release will bump 1 patch version and add the `-lts.1` qualifier. Future LTS releases to that branch will bump the LTS number (e.g. the next version would be `-lts.2`.

This also converts the JavaBom releaser to use stable branch names.

Example PRs:
[snapshot after branch created](https://github.com/googleapis/java-trace/pull/444)
[LTS release PR](https://github.com/googleapis/java-trace/pull/445)


src/releasers/java-bom.ts
```diff
@@ -12,213 +12,71 @@
...
 
-export class JavaBom extends ReleasePR {
-  protected async _run(): Promise<number | undefined> {
-    const versionsManifestContent = await this.gh.getFileContents(
-      'versions.txt'
-    );
-    const currentVersions = VersionsManifest.parseVersions(
-      versionsManifestContent.parsedContent
-    );
+export class JavaBom extends JavaYoshi {
```
bcoe, line 30: awesome, I love that you can extend the `JavaYoshi` base class now.

src/releasers/java-yoshi.ts
```diff
@@ -50,10 +47,19 @@ const CHANGELOG_SECTIONS = [
 ];
 
 export class JavaYoshi extends ReleasePR {
+  private versionsManifestContent?: GitHubFileContents;
+
+  protected async getVersionManifestContent(): Promise<GitHubFileContents> {
+    if (!this.versionsManifestContent) {
+      this.versionsManifestContent = await this.gh.getFileContents(
+        'versions.txt'
```
bcoe, line 55: maybe add a TODO to consolidate with @joeldodge79's manifest format eventually, if we can figure out an approach that works for both needs?

src/releasers/java/version.ts
```diff
@@ -14,49 +14,57 @@
...
-    let extra = '';
-    let snapshot = false;
-    if (match[5]) {
-      extra = match[4];
-      snapshot = match[5] === '-SNAPSHOT';
-    } else if (match[4]) {
-      if (match[4] === '-SNAPSHOT') {
-        snapshot = true;
+    let lts = undefined;
+    if (match[4]) {
```
bcoe, line 58: I'd be tempted to use named regexes here, just so we have less magic numbers.

src/releasers/java/version.ts
```diff
@@ -14,49 +14,57 @@
 
 import {BumpType} from './bump_type';
 
-const VERSION_REGEX = /(\d+)\.(\d+)\.(\d+)(-\w+)?(-SNAPSHOT)?/;
+const VERSION_REGEX = /(\d+)\.(\d+)\.(\d+)(-.+)?/;
```
bcoe, line 17: wondering if it's worth pulling these files into a `utils/` subfolder, maybe it's not worth the extra nesting? i'm on the fence.

bcoe (approved):
I was picturing us making the version bumping configurable, I like the approach you've instead taken of introducing a new releaser -- seems like it's ultimately a bit less complex.

I think it might be worth us adding some guides, such as `docs/java-releases.md` to a `docs` folder, so that external users can learn about our best practices -- what releasers are good for what tasks.

Perhaps open an enhancement issue for this?

PR #814, opened 2021-03-09
refactor(manifest): cache parsed json for config and manifest

joeldodge79:
While trying to add `"release-as"` and `"boostrap-sha"` config options (PR to come), it became apparent the current method of storing only the `GitHubFileContents` version of the config and manifest made it awkward to refer to those objects again (i.e. having to re-parse the "raw" version)

- `this.headManifest` and `this.configFile` now refer to their parsed data
- moved `readConfig` parsing into `getConfigJson`

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -39,19 +39,19 @@ import {
 } from './github-release';
 import {OpenPROptions} from './release-pr';
 
-interface ReleaserConfig {
+interface ReleaserConfigJson {
   'release-type'?: ReleaseType;
   'bump-minor-pre-major'?: boolean;
   'changelog-sections'?: ChangelogSection[];
   'release-draft'?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


bcoe (approved):

src/manifest.ts
```diff
@@ -127,27 +132,24 @@ export class Manifest {
...
   }
 
-  protected async getManifest(): Promise<GitHubFileContents>;
-  protected async getManifest(
-    sha: string
-  ): Promise<GitHubFileContents | undefined>;
-  protected async getManifest(
-    sha?: string
-  ): Promise<GitHubFileContents | undefined> {
+  protected async getManifestJson(): Promise<object>;
```
bcoe, line 140: curious why we use the fairly generic `object` type here?

joeldodge79: kinda lame - I'm pushing an update that locks the types down more strictly.

PR #820, opened 2021-03-10
refactor(manifest): bootstrap-sha and release-as

joeldodge79:
Two new manifest config options:
1. bootstrap-sha: optional top-level only
  - The first time `manifest-pr` is run against a repo (or setup for
  the first time on a new `--default-branch`), it will not find a
  previously merged PR as the starting point to collect commits for
  determine the next release/changelog. The default behavior is to
  iteravily traverse back to the very first commit of the repo. This
  option provides an arbitrary place to start. Once at least one release
  PR has been merged in the history of the `--default-branch`, this
  option is ignored.
2. release-as: optional per package and top-level default
  - manually set the next release version for a package. This will not
  force a release (there still must be user facing commits since the
  last realease). Per-package release-as overrides any top level
  release-as setting. If there is a top-level release-as configured but
  you'd like some packages to auto-increment you can set their
  per-package release-as to an empty string to override the default
  value.

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -43,6 +43,7 @@ interface ReleaserConfigJson {
   'release-type'?: ReleaseType;
   'bump-minor-pre-major'?: boolean;
   'changelog-sections'?: ChangelogSection[];
+  'release-as'?: string;
   'release-draft'?: boolean;
 }
 
@@ -54,6 +55,7 @@ interface ReleaserPackageConfig extends ReleaserConfigJson {
 export interface Config extends ReleaserConfigJson {
   packages: Record<string, ReleaserPackageConfig>;
   parsedPackages: Package[];
+  'bootstrap-sha'?: string;
 }
 
 interface Package {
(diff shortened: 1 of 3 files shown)


src/manifest.ts
```diff
@@ -63,6 +65,7 @@ interface Package {
   bumpMinorPreMajor?: boolean;
   changelogSections?: ChangelogSection[];
   changelogPath?: string;
+  releaseAs?: string;
```
bcoe, line 68: For `release-as` I'm guessing a user would most likely use this during a manual run of the `relese-please` CLI?

test/manifest.ts
```diff
@@ -787,6 +811,99 @@ describe('Manifest', () => {
...
+          'Found version 0.1.2 for node/pkg2 in ' +
+            '.release-please-manifest.json at tip of main',
+          CheckpointType.Success,
+        ],
+        [
+          'Found version 1.2.3 for python in ' +
+            '.release-please-manifest.json at tip of main',
+          CheckpointType.Success,
+        ],
+        ['Processing package: Node(@node/pkg1)', CheckpointType.Success],
```
bcoe, line 901: I find this approach of using the output to the logs as an assertion  clever (in a good way).

bcoe (approved):

PR #822, opened 2021-03-11
fix(changelog): monorepoTags links in changelog

joeldodge79:
changelog links to comparing current to previous tags was broken for
node and python releasers (and others I suspect as well.) Limiting the
scope of the fix to releasers that test `monorepoTags`


src/releasers/rust.ts
```diff
@@ -57,10 +51,10 @@ export class Rust extends ReleasePR {
...
     const changelogEntry: string = await cc.generateChangelogEntry({
       version: candidate.version,
-      // those need to be prefixed so that the links generated by
-      // the changelog generator go to the proper tags
-      currentTag: addPrefix(`v${candidate.version}`),
-      previousTag: addPrefix(candidate.previousTag),
+      currentTag: await this.normalizeTagName(candidate.version),
+      previousTag: candidate.previousTag
+        ? await this.normalizeTagName(candidate.previousTag)
+        : undefined,
```
joeldodge79, line 57: cc @fasterthanlime - I made this fix for node and python and then I discovered you'd already found this discrepancy when you built the rust releaser. I changed it to be uniform with the others. The snapshot didn't change so we should be good.

bcoe (approved):
Good catch 👍

PR #824, opened 2021-03-12
feat(manifest): factory/cli integration

joeldodge79:
This PR finally connects `Manifest` class commands to the cli:

1. `Manifest.pullRequest()` - `manifest-pr`
2. `Manifest.githubRelease()` - `manifest-release`

the only cli flags available are
a) the base set used for a `GitHub` instance (like `--repo-url`, `--default-branch` etc) and
b) `--config` and `--manifest` to override the respective default locations of `release-please-config.json` and `.release-please-manifest.json`

### DIFF
--- a/test/factory.ts
+++ b/test/factory.ts
@@ -13,12 +13,12 @@
 // limitations under the License.
 
 import {describe, it, afterEach} from 'mocha';
-import * as assert from 'assert';
 import {
   factory,
   ReleasePRCommand,
   GitHubReleaseMethod,
   ManifestMethod,
+  ReleasePRMethod,
 } from '../src/factory';
 import * as sinon from 'sinon';
 import {expect} from 'chai';
@@ -164,19 +164,14 @@ describe('factory', () => {
       expect(packageName.getComponent()).to.equal('ruby-test-repo');
     });
     it('throws an error on invalid release type', () => {
-      let caught = false;
-      try {
+      expect(() =>
         factory.releasePR({
           repoUrl: 'googleapis/simple-test-repo',
           packageName: 'simple-test-repo',
           apiUrl: 'https://api.github.com',
           releaseType: 'unknown' as 'go', //hack the typing
-        });
-        assert.fail();
-      } catch (e) {
-        caught = true;
-      }
-      expect(caught).to.be.true;
+        })
+      ).to.throw();
     });
   });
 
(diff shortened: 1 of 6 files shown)


test/factory.ts
```diff
@@ -343,14 +385,36 @@ describe('factory', () => {
...
+      sandbox.stub(instance, 'pullRequest').resolves(32);
+      expect(await factory.call(instance, 'pullRequest')).to.eql(32);
+    });
+    it('errors with bad method on Manifest', async () => {
+      const instance = factory.manifest({
+        repoUrl: 'googleapis/simple-test-repo',
+      });
+      let caught = false;
+      let err: Error;
+      try {
```
bcoe, line 409: I'm to blame, because I think I started this pattern, but I've started moving towards using `assert.rejects` and `assert.throws` for this type of assertion, I think it's cleaner:

https://nodejs.org/api/assert.html#assert_assert_rejects_asyncfn_error_message

You just do:

```js
await assert.rejects(factory.call(instance, 'foo' as ManifestMethod));
```

bcoe (changes requested):
One very small nit, which I think is worth doing since it makes for slightly more elegant tests.

test/factory.ts
joeldodge79: I've tried your syntax and a few other incantations per the docs but I always end up with a test failure and this output (as if the error is bubbling past the `await assert.rejects` somehow):

```
  1) factory
       call                                                                                                                                                                                                                                                                                                                            errors with bad method on Manifest:
     Error: No such method(foo) on Manifest                                                                                                                                                                                                                                                                                         at Object.call (build/src/factory.js:79:15)
          -> /usr/local/google/home/joeldodge/release-please/src/factory.ts:161:11
at Context.<anonymous> (build/test/factory.js:392:52)
          -> /usr/local/google/home/joeldodge/release-please/test/factory.ts:407:36
at processImmediate (internal/timers.js:439:21)
```

joeldodge79: i could give chai-as-promised [a shot...](https://stackoverflow.com/a/32790325/13802304)

joeldodge79: I went with a different pattern:

```typescript
expect(() => factory.call(instance, 'foo' as ManifestMethod)).to.throw(
  'No such method(foo) on Manifest'
);
```

bcoe (approved):

PR #833, opened 2021-03-20
feat(manifest): add support for releasing root module

bcoe:
Add support for a special "." path which indicates that a root module should be released.

This is for the [googleapis use-case](https://github.com/googleapis/google-api-nodejs-client/pull/2559).

CC: @joeldodge79

### DIFF
--- a/__snapshots__/manifest.js
+++ b/__snapshots__/manifest.js
@@ -1,3 +1,134 @@
+exports['Manifest pullRequest allows root module to be published, via special "." path: changes'] = `
+
+filename: node/pkg1/CHANGELOG.md
+# Changelog
+
+## [4.0.0](https://www.github.com/fake/repo/compare/pkg1-v3.2.1...pkg1-v4.0.0) (1983-10-10)
+
+
+### ⚠ BREAKING CHANGES
+
+* **@node/pkg1:** major new feature
+
+### Features
+
+* **@node/pkg1:** major new feature ([e3ab0ab](https://www.github.com/fake/repo/commit/e3ab0abfd66e66324f685ceeececf35c))
+
+filename: node/pkg1/package.json
+{
+  "name": "@node/pkg1",
+  "version": "4.0.0"
+}
+
+filename: node/pkg2/CHANGELOG.md
+# Changelog
+
+## [2.0.0](https://www.github.com/fake/repo/compare/pkg2-v1.2.3...pkg2-v2.0.0) (1983-10-10)
+
+
+### ⚠ BREAKING CHANGES
+
+* **@node/pkg2:** major new feature
+
+### Features
+
+* **@node/pkg2:** major new feature ([72f962d](https://www.github.com/fake/repo/commit/72f962d44ba0bcee15594ea6bdc67d8b))
+
+filename: node/pkg2/package.json
+{
+  "name": "@node/pkg2",
+  "version": "2.0.0"
+}
+
+filename: CHANGELOG.md
+# Changelog
+
+## [3.0.0](https://www.github.com/fake/repo/compare/googleapis-v2.0.0...googleapis-v3.0.0) (1983-10-10)
+
+
+### ⚠ BREAKING CHANGES
+
+* **@node/pkg2:** major new feature
+
+### Features
+
+* **@node/pkg2:** major new feature ([72f962d](https://www.github.com/fake/repo/commit/72f962d44ba0bcee15594ea6bdc67d8b))
+
+filename: package.json
+{
+  "name": "googleapis",
+  "version": "3.0.0"
+}
+
+filename: .release-please-manifest.json
+{
+  ".": "3.0.0",
+  "node/pkg1": "4.0.0",
+  "node/pkg2": "2.0.0"
+}
+
+`
+
+exports['Manifest pullRequest allows root module to be published, via special "." path: options'] = `
+
+upstreamOwner: fake
+upstreamRepo: repo
+title: chore: release
+branch: release-please/branches/main
+description: :robot: I have created a release \\*beep\\* \\*boop\\*
+
+---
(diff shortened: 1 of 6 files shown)


test/manifest.ts
```diff
@@ -283,6 +286,77 @@ describe('Manifest', () => {
...
+        },
+      });
+      const commits = [
+        buildMockCommit('feat(@node/pkg1)!: major new feature', [
+          'node/pkg1/src/foo.ts',
+        ]),
+        buildMockCommit('feat(@node/pkg2)!: major new feature', [
+          'node/pkg2/src/bar.ts',
+        ]),
+      ];
```
joeldodge79, line 309: the `googleapis/google-api-nodejs-client` use case ("aggregate" package) makes the most sense to me but I wonder if other folks might have a setup where by non-submodule changes could be important to the `'.'` package? I'm not saying we need to extend existing functionality, but it would be good to understand what the current behavior would be if say we added a commit to this list like

```typescript
        buildMockCommit('fix(root): root only change', [
          'src/foo.ts',
        ])
```

in this particular case the version bumping would be shadowed by the `!` commits on submodules but I'd expect at least an entry in the root changelog?

__snapshots__/manifest.js
```diff
@@ -1,3 +1,134 @@
...
+
+filename: package.json
+{
+  "name": "googleapis",
+  "version": "3.0.0"
+}
+
+filename: .release-please-manifest.json
+{
+  ".": "3.0.0",
```
joeldodge79, line 65: wow, so close out of the box! I wonder why `@node/pkg1`'s change didn't get picked up in the root changelog?

joeldodge79 (changes requested):
I'm a little surprised this worked as smoothly as it did! but it does seem like we're missing a changelog entry for `@node/pkg1` in the root changelog ?

joeldodge79:
random question: how do you setup interactive debugging on this project? I've been cherry-picking [this commit](https://github.com/joeldodge79/release-please/commit/18ab3556561a71981544a238106c969fc1bd6595) to use vscode debugging

test/manifest.ts
```diff
@@ -283,6 +286,77 @@ describe('Manifest', () => {
       ]);
     });
 
+    it('allows root module to be published, via special "." path', async function () {
```
joeldodge79, line 289: also, I think we should add a test under the `describe('githubRelease'` section. I'd be curious to see what the tag creation looks like for the root package.

also maybe a note somewhere in `docs/manifest-releaser.md` about this root pkg ability?

__snapshots__/manifest.js
bcoe: I was accidentally overwriting the `commits` variable, unless it was first in the list.

test/manifest.ts
bcoe: Added a test case 👍

test/manifest.ts
bcoe: Done 👍

bcoe: Done 👍

test/manifest.ts
```diff
@@ -1750,6 +1825,86 @@ describe('Manifest', () => {
...
+          ['release-please-config.json', config],
+          ['.release-please-manifest.json', manifest],
+          ['CHANGELOG.md', '#Changelog\n\n## v3.2.1\n\n* entry'],
+        ],
+      });
+      expectLabelAndComment(mock, {
+        addLabel: 'autorelease: tagged',
+        removeLabel: 'autorelease: pending',
+        prComments: [
+          ':robot: Release for googleapis is at https://pkg1@3.2.1:html :sunflower:',
```
joeldodge79, line 1867:
```suggestion
          ':robot: Release for googleapis is at https://googleapis@3.2.1:html :sunflower:',
```

this value comes from the mock `.resolves()` values on L1885-6 where I'll leave suggested changes too. It's just cosmetic but slightly confusing.

test/manifest.ts
```diff
@@ -1750,6 +1825,86 @@ describe('Manifest', () => {
...
+          false
+        )
+        .once()
+        .resolves({
+          name: 'googleapis googleapis-v3.2.1',
+          tag_name: 'googleapis-v3.2.1',
+          draft: false,
+          body: '',
+          html_url: 'https://pkg1@3.2.1:html',
+          upload_url: 'https://pkg1@3.2.1:upload',
```
joeldodge79, line 1886:
```suggestion
          html_url: 'https://googleapis@3.2.1:html',
          upload_url: 'https://googleapis@3.2.1:upload',
```

test/manifest.ts
```diff
@@ -1750,6 +1825,86 @@ describe('Manifest', () => {
...
+        '.': {
+          version: '3.2.1',
+          major: 3,
+          minor: 2,
+          patch: 1,
+          pr: 22,
+          draft: false,
+          body: '',
+          sha: 'abc123',
+          html_url: 'https://pkg1@3.2.1:html',
```
joeldodge79, line 1901:
```suggestion
          html_url: 'https://googleapis@3.2.1:html',
```

test/manifest.ts
```diff
@@ -1750,6 +1825,86 @@ describe('Manifest', () => {
...
+          tag_name: 'googleapis-v3.2.1',
+          name: 'googleapis googleapis-v3.2.1',
+          upload_url: 'https://pkg1@3.2.1:upload',
```
joeldodge79, line 1904:
```suggestion
          upload_url: 'https://googleapis@3.2.1:upload',
```

joeldodge79 (changes requested):
minor test typo/nits but phew! glad the overwritten changelog was just a test bug

chingor13 (approved):

PR #835, opened 2021-03-22
refactor(manifest): typing cleanup

joeldodge79:
- `Package` type is really a limited collection of ReleasePROptions and
  GitHubReleaseOptions.
- Renamed some internal interfaces used to join package config to
  commit/last-version data for ReleasePR instances (`PackageForReleaser`)
  and to data for opening a pr (`PackageWithPRData`)
- DRYed out code to instantiate a ReleasePR instance
- reverted `"release-draft"` config option to the original
  GitHubReleaseFactoryOptions.draft to remove option parsing complexity:
  if we ever need to support a "draft PR" then we'll make sure to
  namespace that config option appropriately (e.g. `"draft-pr"`)

### DIFF
--- a/test/manifest.ts
+++ b/test/manifest.ts
@@ -1085,7 +1085,7 @@ describe('Manifest', () => {
           python: {
             'release-type': 'python',
             'package-name': 'foolib',
-            'release-draft': true,
+            draft: true,
           },
         },
       });
--- a/src/index.ts
+++ b/src/index.ts
@@ -30,12 +30,12 @@ interface GitHubOptions {
 }
 
 // Used by GitHubRelease: Factory and Constructor
-interface GitHubReleaseOptions {
+export interface GitHubReleaseOptions {
   draft?: boolean;
 }
 
 // Used by ReleasePR: Factory and Constructor
-interface ReleasePROptions {
+export interface ReleasePROptions {
   path?: string;
   packageName?: string;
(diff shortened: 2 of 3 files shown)


sofisl (approved):

test/manifest.ts
```diff
@@ -1085,7 +1085,7 @@ describe('Manifest', () => {
           python: {
             'release-type': 'python',
             'package-name': 'foolib',
-            'release-draft': true,
+            draft: true,
```
bcoe, line 1088: I believe this will technically be breaking right? If any folks happened to have found the `release-draft` config already?

bcoe (commented):
Do we expect that this will break an upstream library using the types? We might want to test in _repo-automation-bots_ before we merge.

test/manifest.ts
joeldodge79: you are correct. when I first wrote/submitted this it was very shortly after the factory/cli wireup so I felt ok at that time that no one was using it yet. but now that it's been a minute and this is setup in maybe a few places it could break people who are using this knob... I could just revert it and keep `release-draft`?

joeldodge79:
> an upstream library using the types

The actual type changes are on types/interfaces that are not exported so that shouldn't break anybody. The only potential breaking change is the `'release-draft'` -> `'draft'` rename

bcoe (approved):

PR #841, opened 2021-03-25
fix(manifest): do not group changes under "."

bcoe:
If "." was anything but the last path in the list, it would result in only receiving changes for the special "." path.

### DIFF
--- a/src/commit-split.ts
+++ b/src/commit-split.ts
@@ -91,7 +91,11 @@ export class CommitSplit {
         let pkgName;
         if (this.packagePaths) {
           // only track paths under this.packagePaths
-          pkgName = this.packagePaths.find(p => file.indexOf(p) >= 0);
+          pkgName = this.packagePaths.find(p => {
+            // The special "." path, representing the root of the module
+            // should be ignored by commit-split:
+            return file.indexOf(p) >= 0 && p !== '.';
+          });
         } else {
           // track paths by top level folder
           pkgName = splitPath[0];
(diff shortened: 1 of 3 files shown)


SurferJeffAtGoogle (approved):

src/commit-split.ts
```diff
@@ -91,7 +91,11 @@ export class CommitSplit {
         let pkgName;
         if (this.packagePaths) {
           // only track paths under this.packagePaths
-          pkgName = this.packagePaths.find(p => file.indexOf(p) >= 0);
+          pkgName = this.packagePaths.find(p => {
+            // The special "." path, representing the root of the module
+            // should be ignored by commit-split:
+            return file.indexOf(p) >= 0 && p !== '.';
```
joeldodge79, line 97: hm, it seems like the `'.'` packagePath will never get any commits unless I'm reading this wrong? how will it get changelog entries and version bumps?

joeldodge79: but it seems to work. I confess I don't feel like I totally understand really how we're assigning commits to packages. I have a case were a commit got assigned to a package but none of the files for the commit are in the package path... and now that I wrote that I know exactly what happened in my case:

package path: "python"
commit files all under "examples/python/..."
but `"examples/python/foo.py".indexOf("python") >= 0` is true! I think we want a stricter check here? probably

```
return file.indexOf(p) === 0 && p !== '.';
```

I'll submit a patch for this bug, but I still don't understand how the `"."` package is being assign the appropriate commits as it seems to be...

Issue #597, opened 2020-10-16
Please include Version.java in VERSION_REPLACEMENT_FILENAMES


igorbernstein2:
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

 **Is your feature request related to a problem? Please describe.**
I would like to have a robust way to embed my client's version in the client. This is necessary to enable DirectPath despite customer applications shading client artifacts.

 **Describe the solution you'd like**
I would like to add a Version.java file to bigtable client and it be auto updated during release. Currently I either have rely on resource filtering gymnastics or hijack the existing allowlist entry for GoogleUtils.java

 **Describe alternatives you've considered**
1. Using resource filtering gymnastics to populate files with a version from a maven property. This is pretty clunky and will probably cause issues when importing in g3
2. Using the already allowlisted GoogleUtils.java. This is a bit hacky and intended for a different audience.

 **Additional context**
https://github.com/googleapis/java-bigtable/pull/451

Issue #305, opened 2019-11-04
develop more generic "update" class for long-tail of version replacements


bcoe:
Looking at `google-api-java-client`, there was a long tail of files we failed to update the version in:

```bash
./google-api-client/src/test/java/com/google/api/client/googleapis/GoogleUtilsTest.java:    String version = "1.30.3";
./google-api-client/src/test/java/com/google/api/client/googleapis/GoogleUtilsTest.java:    String version = "1.30.3-SNAPSHOT";
./google-api-client/src/main/java/com/google/api/client/googleapis/GoogleUtils.java:  public static final String VERSION = "1.30.3".toString();
./google-api-client-bom/README.md:      <version>1.30.3</version>
```

We continue to add one off fixes, like https://github.com/googleapis/release-please/pull/304, but it might be nice to have a slightly more generic update class that handles a variety of simple string replacements.

e.g., `<version>0.3.0</version>`, `VERSION = "0.20.4"`, maybe the class could support a variety of regexes like releasetool.

PR #850, opened 2021-04-09
feat: add ability to specify extra files for releasers to consider

chingor13:
Currently, only Java utilizes this option, but this could be used by
any release strategy.

Towards #305 
Towards #597

### DIFF
--- a/src/releasers/java-yoshi.ts
+++ b/src/releasers/java-yoshi.ts
@@ -299,6 +299,18 @@ export class JavaYoshi extends ReleasePR {
         })
       );
     });
+
+    this.extraFiles.forEach(path => {
+      updates.push(
+        new JavaUpdate({
+          path,
+          changelogEntry,
+          versions: candidateVersions,
+          version: candidate.version,
+          packageName: packageName.name,
+        })
+      );
+    });
     return updates;
   }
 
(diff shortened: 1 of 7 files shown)


chingor13:
cc @igorbernstein2 This would be a more generalized solution for being able to make replacements is one-off files (without hard-coding a list)

src/releasers/java-yoshi.ts
```diff
@@ -299,6 +299,18 @@ export class JavaYoshi extends ReleasePR {
         })
       );
     });
+
+    this.extraFiles.forEach(path => {
```
bcoe, line 303: my slight concern here is behavior that's one-language specific, perhaps add a TODO to flesh out this behavior. I'm imagining you'd eventually pass a file plus the updater to apply?

chingor13: I think that will be too much config. The general use case for extra files would be one-offs that update the version strings in a file. Each language will want a generic way to do this - we don't want a bunch of one-off updaters for this.

bcoe (approved):

bcoe (approved):
👍 with the note that I'd like to minimize language specific features if we can make them more generic.

viv718918 (commented):
ok

PR #857, opened 2021-04-13
fix: adjust patch rule for Java LTS strategy

chingor13:
The LTS branch will not bump the minor version - it will preserve the
major/minor/patch of the version it was patched from.

### DIFF
--- a/src/releasers/java/version.ts
+++ b/src/releasers/java/version.ts
@@ -94,14 +94,21 @@ export class Version {
         break;
       case 'lts':
         if (this.lts) {
-          this.lts += 1;
-        } else {
           if (!this.snapshot) {
-            this.patch += 1;
+            this.lts += 1;
           }
+        } else {
           this.lts = 1;
         }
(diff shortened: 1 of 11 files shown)


src/releasers/java/version.ts
```diff
@@ -94,15 +94,22 @@ export class Version {
         break;
       case 'lts':
         if (this.lts) {
-          this.lts += 1;
-        } else {
           if (!this.snapshot) {
```
bcoe, line 97: mind adding a couple comments to `snapshot` vs., `lts-snapshot` explaining the nuances between how we bump the two types of releases?

bcoe (changes requested):
I think it would be good to explain the nuance that requires the new `lts-snapshot`.

chingor13:
> I think it would be good to explain the nuance that requires the new `lts-snapshot`.

Added a note in the docs/java-release.md and in the java-lts releaser

bcoe (approved):

PR #884, opened 2021-04-24
feat: add custom logger interface

### LINKED ISSUE #883: Provide injectable logger interface
Similar to https://github.com/googleapis/code-suggester/issues/178, we should be able to provide a custom logger interface.

This would allow an integration to inject a structured logger for metrics or other log processing.

chingor13:
* Add a CheckpointLogger which outputs similar output to the checkpoint function.

Fixes #883


bcoe (approved):
This looks good to me, as soon as you're ready to turn it into a PR.

PR #894, opened 2021-04-30
feat(python): support pyproject.toml

### LINKED ISSUE #515: Add support for python repositories with a pyproject.toml
I have a python repository managed with [poetry](https://github.com/python-poetry/poetry) that I'd like to use release-please with.
Support for pyproject.toml defined python repositories.
Dropping poetry support for my project, which isn't really something I'd like to do in order to use release-please.
 **Additional context**
None

Mesteery:

Fixes #515.
Fixes https://github.com/google-github-actions/release-please-action/issues/287.

Added support for `pyproject.toml` (`[project]` or `[tool.poetry]`): the version is changed in the `pyproject.toml` if present and valid. If a name is specified in this file, `__version__` (if present) in `<name>/__init__.py` is also updated.
To be more generic, the `VersionPy` updater has been renamed to `PyBasedFile` (used for `version.py` and `__init__.py`).

*I didn't really do extensive unit tests due to lack of time (and the lack of knowledge of the test structure in this project).*


Mesteery:
@googlebot I signed it!

bcoe:
@Mesteery thank you for the contribution, I will need to get someone who knows Python better than me to take a look 😆

Julien00859:
Hello @bcoe, first I would like to thank you to give this pull request a chance. Sorry to drag you into the complicated world of python packaging. I'm working with @Mesteery on a project where we plan to use flit, a packager for python. We love release-please that we are using in other projects and we would like to use it in our new project too. Sadly, release-please is not yet compatible with flit, this pull-request aims at bringing the compatibility not only for flit but for a bunch of other packager for python.

The `pyproject.toml` is a *new* (2016) file that was originally designed to store generic build-system configuration (under the `[build-system]` section). Many tools quickly adopted that file to store their own configuration under a `[tools.x]` section where `x` is the tool's name, quickly the pep got revised and the `[tool.x]` section got standardized. You can read more here [pep-518](https://www.python.org/dev/peps/pep-0518/).

Fast forwarding a few years later, the packaging environment is evolving and people are dragging away from the de-facto standard [setuptools](https://setuptools.readthedocs.io/en/latest/userguide/index.html) to give others tools like [poetry](https://python-poetry.org/), [flit](https://pypi.org/project/flit/) or [pdm](https://pdm.fming.dev/) a chance. Each of these tools was using the `[tool.x]` section to add various project metadata. Just last year, a pep standardized a `[project]` section of that very `pyproject.toml` file to store the various project metadata a standardized way. You can read more here [pep-621](https://www.python.org/dev/peps/pep-0621)

Obviously none of these systems read the version from the same place. Setuptools gets the version either from `setup.py` or via `setup.cfg`. Poetry gets the version from the `version` option of the `[tool.poetry]` section of the `pyproject.toml` file. Both flit and pdm gets the version from the standardized `version` option of the `[project]` section of the `pyproject.toml` file. Do not trust the documentation of flit, it is documented that it gets the configuration only from a `__version__` variable defined in the project source code, this is wrong, it is capable to retrieve the version from `pyproject.toml` too since https://github.com/takluyver/flit/pull/393. You can read more here [pep-621](https://www.python.org/dev/peps/pep-0621/#version)

--

When using **setuptools**, it is possible to configure either setup.py or setup.cfg to read the version from a file. This makes your `release-type: simple` compatible out-of-the box. While it would be handy release-please updates the configuration file right away, the work around is really easy, I wouldn't suggest you try to support it. The python community is dragging away from setuptools and that there are ongoing discussion whether or not to deprecate `setup.cfg` in favor of `pyproject.toml`.

I am not an expert of **poetry** as I never used the project myself outside of a sandbox to test various build systems. I don't know if they are willing to use the standardized `[project] version=` over `[tool.poetry] version=` any time soon. For the moment, dedicated code to support poetry is the only way.

When using either **flit** or **pdm**, the standard `[project] version=` is supported. This is the important piece and I would suggest to take care it is correctly tested in Mesteery's work.

--

I'm not a typescript developer myself nor do I know release-please internals. Hopefully this message helps you in the reviewing process. If you want to see what's a `pyproject.toml` file looks like, we have it all documented in this [wiki](https://github.com/NotANameServer/incipyt/wiki/Build-systems)

Regards,

src/updaters/python/py-based-file.ts
```diff
@@ -15,7 +15,10 @@
 import {Update, UpdateOptions, VersionsMap} from '../update';
 import {GitHubFileContents} from '../../github';
 
-export class VersionPy implements Update {
+/**
+ * Python based files (for version.py and __init__.py)
```
tmatsuo, line 19: Python file with a `__version__` property (or attribute, or whatever).

src/releasers/python.ts
```diff
@@ -20,8 +20,16 @@ import {Changelog} from '../updaters/changelog';
 // Python specific.
 import {SetupPy} from '../updaters/python/setup-py';
 import {SetupCfg} from '../updaters/python/setup-cfg';
-import {VersionPy} from '../updaters/python/version-py';
+import {PyBasedFile} from '../updaters/python/py-based-file';
```
tmatsuo, line 23: I'm not a big fan of the new name `PyBasedFile`.
It makes some sense in a context of `pyproject.toml` with the `name` field, but it doesn't make sense for the original `version.py` case.

Can we keep the original name and add a comment about `__init__.py` case?
Or maybe `python-file-with-version.ts` and `PythonFileWithVersion`?

tmatsuo (commented):
LGTM

I'm planning to approve around noon, currently I'm asking internally around if there's anyone else wants to also review.

src/releasers/python.ts
```diff
@@ -76,6 +84,37 @@ export class Python extends ReleasePR {
...
+      updates.push(
+        new PyProjectToml({
+          path: this.addPath('pyproject.toml'),
+          changelogEntry,
+          version: candidate.version,
+          packageName: packageName.name,
+        })
+      );
+
+      if (pyProject.name) {
```
tmatsuo, line 99: It seems like this code path is not covered by the test. Can you add the test case for it?

tmatsuo (approved):
LGTM

tmatsuo:
@Mesteery Thanks for the contribution. I'm going to merge this.

Let us know how it goes with your repository with `pyproject.toml`.

LucasRoesler:
Hi @Mesteery I am sorry for commenting on something that is a bit old, but do you have a public repo that is using this. I have run into an issue where release-please doesn't seem to recognize or even search for the pyproject.toml.  I was hoping to look at an example repo that is using it before trying to compare with before creating a bug

Mesteery:
Hi! Yes, there is for example https://github.com/NotANameServer/incipyt/.

LucasRoesler:
Thanks that was a big help. I found two issues, one with my structure and the other was a github permissions issue 

Again, thanks. :pray:

efebeydogan01:
Hello, I am sorry to bump a 2 year old issue but I am running into problems configuring release-please in a monorepo with typescript and python packages. Basically, my current configuration works for my typescript packages, and release-please creates a changelog for each package, as well as creating release PR's for each of them when I new push code. However, I couldn't get it to work with my python packages, each package has its own pyproject.toml file. You can view the question I asked on StackOverflow here as well: https://stackoverflow.com/questions/77391427/how-to-configure-google-release-please-with-pyproject-toml-files-in-monorepo. Any help would be greatly appreciated, thank you.

PR #903, opened 2021-05-11
feat: custom version increment strategies

joeldodge79:
Allow for version incrementing strategies other than conventional
commits. The first example is "always-patch" which completely ignores
conventional commits and hard-codes a "patch" level version bump. This
strategy is useful in cases where the "major" and "minor" numbers are
set by conditions outside the package codebase but the "patch" number
is still available for representing changes as a new version.


joeldodge79:
@bcoe and I discussed this possibility a while back: the basic problem is I have several packages that have their major/minor version numbers dictated by outside factors but we do want to increment the patch for **any** user facing change. We've been working around this with a lot of constant adding/removing/tweaking of `"release-as"` keys in the `release-please-config.json` but that is getting old quickly.

While we'll still have to set a `"release-as"` key on a ~monthly basis (for resetting the major/minor versions of our non-semver-standard packages), having this `"always-patch"` version bumping strategy will reduce a lot of mundane yet confusing config file churn.

I'm open to a different approach if anyone has suggestions: it would be awesome to have some kind of mapping like the `ChangelogSection` stuff. Ideally we'd only be exposing that functionality of `conventional-changelog` but it doesn't exist.

joeldodge79:
@chingor13 where would the `interface VersioningStrategy` suggestion get called?

I've taken a quick look through the java-bom and java-lts releasers and it looks like in each case they completely override `ReleasePR.coerceReleaseCandidate` - so in addition to their custom versioning, they don't follow the [`release-as` footer / `releaseAs` config](https://github.com/googleapis/release-please/blob/d8ab1bbcc61776b80e7978a008d70631a62e5dc1/src/release-pr.ts#L304) logic the parent provides (incidentally it looks like the `preRelease` argument is currently unused outside of test code)

Maybe we can meet and talk about how to introduce this feature so it works both for the Java releasers as well as others that still rely on the `ReleasePR.coerceReleaseCandidate` logic?

joeldodge79:
sorry, got distracted - been meaning to follow up out of band with @chingor13 to discuss a more generic solution for my needs + Java use cases.

bcoe:
@joeldodge79 👍 will leave it to you and @chingor13 to discuss, let me know when/if my review is needed.

bcoe:
I'd love to figure out how to unblock this work next quarter, perhaps we can have a catch up soonish? Apologies for leaving this to stagnate.

bcoe:
@joeldodge79 okay, @chingor13 has landed his refactor that pulls versioning into its own class.

I think it should be a good starting point for you to finally implement this functionality.

chingor13:
The always-bump-patch versioning strategy is built-in to the new refactored version. You can mix and match any language strategy with a versioning strategy. Previously the always-bump-patch semantics were bundled with the `java-backport` ReleasePR type, but now any component can specify it.

joeldodge79:
@chingor13 awesome, thanks! could you point me to the code or examples of using `always-bump-patch` - we'd want to apply it to a subset of packages in the [release-please-config.json](/docs/manifest-releaser.md#configfile) of [our monorepo](https://github.com/looker-open-source/sdk-codegen/blob/main/release-please-config.json)

bcoe:
@joeldodge79 okay if we bring this conversation to chat? I think the functionality is likely close to where you need it, and @chingor13 just needs to point you towards it.

PR #905, opened 2021-05-11
fix: correctly find Java LTS latestTag version strings

### LINKED ISSUE #904: Java LTS reproposes same release
https://github.com/googleapis/java-bigquery/pull/1293 is being proposed, but that version was already released.

It looks like `latestTag()` is skipping the `-sp.1` release version.

chingor13:
Fixes #904

### DIFF
--- a/src/release-pr.ts
+++ b/src/release-pr.ts
@@ -548,6 +548,25 @@ export class ReleasePR {
   }
 
   /**
+   * Normalize version parsing when searching for a latest release.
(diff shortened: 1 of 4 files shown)


chingor13:
>  looks reasonable to me, would be nice to have test for the `release-pr` `normalizeVersion` behavior (if there isn't one).

These should be handled by the `ReleasePR` `latestTag` tests.

PR #923, opened 2021-06-02
feat: use original json's indentation and style

### LINKED ISSUE #453: release-please rewrites indentation on package.json
#### Environment details

Tested with Github action bcoe/release-please-action@1.2.1 (uses release-please@5.2.0).

#### Steps to reproduce

1. Setup release-please-action on a repository with tab indented package.json
2. Wait for release to be requested
3. Note that PR changes all indentation of package.json to 2 spaces, example cfware/fastify-babel#45.

Other tools like `standard-version` seem to detect the existing indentation and leave it alone.

Mesteery:

Fixes #453 🦕

*(I don't understand why the tests don't pass on windows)*

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,8 @@
     "typescript": "^3.8.3",
     "unist-util-visit": "^2.0.3",
     "unist-util-visit-parents": "^3.1.1",
-    "yargs": "^16.0.0"
+    "yargs": "^16.0.0",
+    "detect-indent": "^6.1.0"
   },
   "engines": {
     "node": ">=10.12.0"
--- a/src/plugins/node-workspace.ts
+++ b/src/plugins/node-workspace.ts
@@ -14,33 +14,45 @@
 
 import * as semver from 'semver';
 import cu = require('@lerna/collect-updates');
-import {Package, PackageJson} from '@lerna/package';
+import {Package as LernaPackage, PackageJson} from '@lerna/package';
 import {PackageGraph, PackageGraphNode} from '@lerna/package-graph';
 import {runTopologically} from '@lerna/run-topologically';
 import {ManifestPlugin} from './plugin';
 import {ManifestPackageWithPRData, ManifestPackage} from '..';
 import {VersionsMap} from '../updaters/update';
-import {packageJsonStringify} from '../util/package-json-stringify';
+import {jsonStringify} from '../util/json-stringify';
 import {CheckpointType} from '../util/checkpoint';
 import {RELEASE_PLEASE} from '../constants';
 import {Changes} from 'code-suggester';
 import {ConventionalCommits} from '../conventional-commits';
 import {Changelog} from '../updaters/changelog';
 
-type PathPkgJson = Map<string, PackageJson>;
+class Package extends LernaPackage {
+  constructor(
+    public readonly rawContent: string,
+    location: string,
+    pkg?: PackageJson
+  ) {
+    super(pkg ?? JSON.parse(rawContent), location);
+  }
+
+  clone() {
+    return new Package(this.rawContent, this.location, this.toJSON());
+  }
+}
 
 export default class NodeWorkspaceDependencyUpdates extends ManifestPlugin {
   // package.json contents already updated by the node releasers.
   private filterPackages(
     pkgsWithPRData: ManifestPackageWithPRData[]
-  ): PathPkgJson {
-    const pathPkgs = new Map();
+  ): Map<string, string> {
+    const pathPkgs = new Map<string, string>();
     for (const pkg of pkgsWithPRData) {
       if (pkg.config.releaseType === 'node' && pkg.config.path !== '.') {
         for (const [path, fileData] of pkg.prData.changes) {
           if (path === `${pkg.config.path}/package.json`) {
             this.log(`found ${path} in changes`, CheckpointType.Success);
-            pathPkgs.set(path, JSON.parse(fileData.content!) as PackageJson);
+            pathPkgs.set(path, fileData.content!);
           }
         }
       }
@@ -51,21 +63,21 @@ export default class NodeWorkspaceDependencyUpdates extends ManifestPlugin {
   // all packages' package.json content - both updated by this run as well as
   // those that did not update (no user facing commits).
   private async getAllWorkspacePackages(
-    rpUpdatedPkgs: PathPkgJson
+    rpUpdatedPkgs: Map<string, string>
   ): Promise<Map<string, Package>> {
-    const nodePkgs = new Map();
+    const nodePkgs = new Map<string, Package>();
     for (const pkg of this.config.parsedPackages) {
       if (pkg.releaseType !== 'node' || pkg.path === '.') {
         continue;
       }
       const path = `${pkg.path}/package.json`;
-      let contents: PackageJson;
+      let contents: string;
       const alreadyUpdated = rpUpdatedPkgs.get(path);
       if (alreadyUpdated) {
         contents = alreadyUpdated;
       } else {
-        const fileContents = await this.gh.getFileContents(path);
-        contents = JSON.parse(fileContents.parsedContent);
+        const {parsedContent} = await this.gh.getFileContents(path);
+        contents = parsedContent;
       }
       this.log(
         `loaded ${path} from ${alreadyUpdated ? 'existing changes' : 'github'}`,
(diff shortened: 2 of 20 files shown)


bcoe (approved):

bcoe (approved):
Great approach 👌  thank you for the contribution.

bcoe:
@Mesteery unfortunately, it looks like some tests are having issues on Windows, likely due to line endings. Any thoughts?

Mesteery:
I don't understand why the tests don't pass on Windows.

src/plugins/node-workspace.ts
```diff
@@ -52,39 +55,39 @@ export default class NodeWorkspaceDependencyUpdates extends ManifestPlugin {
...
     }
     return nodePkgs;
   }
 
   private async runLernaVersion(
     rpUpdatedPkgs: PathPkgJson,
-    allPkgs: Map<string, Package>
-  ): Promise<Map<string, Package>> {
+    allPkgs: Map<string, [Package, string]>
+  ): Promise<Map<string, [Package, string]>> {
```
bcoe, line 85: I'm a little worried about the implicit behavior we've introduced here, where someone needs to know that the first argument is the package object, and the second argument is the raw string.

I think I'd rather that we create a new interface that included the Package type and the raw content type.

@joeldodge79 does this sound reasonable to you?

bcoe (changes requested):
Got windows tests passing 👍 

did see one slight refactor that I think would be worthwhile in the manifest class however.

Let me know what you think?

src/plugins/node-workspace.ts
Mesteery: We can give a name to the elements of the tuple like `[package: Package, rawContent: string]`?

bcoe: The line that concerns me the most is this:

`allUpdated.map(p => [p.location, [p, allPkgs.get(p.location)![1][1]]])`

Because you need to know that `1` represents the original content.

It would be nicer to be able to do something like:

```js
allUpdated.map(p => [p.location, [p, allPkgs.get(p.location)![1][1]]])
```

With the triple nested arrays, I find it a bit hard to know what's going on, I think named properties could help clean this up a bit.

Would be curious to know @joeldodge79's feedback too, who implemented most of this class.

joeldodge79: I agree that [L207](https://github.com/googleapis/release-please/pull/923/files?file-filters%5B%5D=.js&file-filters%5B%5D=.ts#diff-811e4913774d18713311ccab03d320c30f1a04286989067b28003404fd88197dR207) is hard on my brain. Does TS supported namedtuples?

What if we made an assumption that all files under the control of a plugin had the same indentation style? Then we could set a `indentContent` instance property on the plugin using the first raw content we come across and reference that for every call rather than passing around the raw content for each file. Alternatively we could make `jsonStringify` a simple class that takes `indentedContent: string` as a constructor and initialize it once for the plugin.

But if we really want the flexibility of supporting different indentation styles per plugin-scope in a monorepo then this seems like the only solution.

bcoe: > What if we made an assumption that all files under the control of a plugin had the same indentation style?

If we think this is a reasonable assumption, I like this idea as it cuts down on the amount of information we need to pass around.

bcoe:
@Mesteery does @joeldodge79's idea of tracking the indentation at the outset seem reasonable? Still have the cycles to get this over finish line? I under stand it can be a hassle when there are multiple back and forths on design.

Mesteery:
> idea of tracking the indentation at the outset seem reasonable

Sorry, I forgot to answer. I didn't quite understand this idea.

src/plugins/node-workspace.ts
```diff
@@ -40,7 +40,10 @@ export default class NodeWorkspaceDependencyUpdates extends ManifestPlugin {
         for (const [path, fileData] of pkg.prData.changes) {
           if (path === `${pkg.config.path}/package.json`) {
             this.log(`found ${path} in changes`, CheckpointType.Success);
-            pathPkgs.set(path, JSON.parse(fileData.content!) as PackageJson);
```
joeldodge79, line 43: maybe

```typescript 
if (this.indentContent === undefined) {
  this.indentContent = fileData.content;
}
pathPkgs.set(path, JSON.parse(fileData.content!) as PackageJson);
```

src/plugins/node-workspace.ts
```diff
@@ -218,14 +223,14 @@ export default class NodeWorkspaceDependencyUpdates extends ManifestPlugin {
       const filePath = `${data.config.path}/package.json`;
       const updated = allUpdated.get(filePath)!; // bug if not defined
       data.prData.changes.set(filePath, {
-        content: packageJsonStringify(updated.toJSON()),
```
joeldodge79, line 221: and then something like this here:

```typescript
content: jsonStringify(updated.toJSON(), this.indentContent),
```

joeldodge79 (commented):
inline samples for what this might look like (first suggestion):
> set a indentContent instance property on the plugin using the first raw content we come across and reference that for every call rather than passing around the raw content for each file

src/plugins/node-workspace.ts
Mesteery: I understood, thank you! I'll make the changes soon.

src/plugins/node-workspace.ts
```diff
@@ -14,33 +14,48 @@
...
-import {packageJsonStringify} from '../util/package-json-stringify';
+import {jsonStringify} from '../util/json-stringify';
 import {CheckpointType} from '../util/checkpoint';
 import {RELEASE_PLEASE} from '../constants';
 import {Changes} from 'code-suggester';
 import {ConventionalCommits} from '../conventional-commits';
 import {Changelog} from '../updaters/changelog';
 
-type PathPkgJson = Map<string, PackageJson>;
+class Package extends LernaPackage {
```
Mesteery, line 30: What do you think about something like this?

src/plugins/node-workspace.ts
```diff
@@ -14,33 +14,48 @@
...
-  ): PathPkgJson {
-    const pathPkgs = new Map();
+  ): Map<string, string> {
+    const pathPkgs = new Map<string, string>();
     for (const pkg of pkgsWithPRData) {
       if (pkg.config.releaseType === 'node' && pkg.config.path !== '.') {
         for (const [path, fileData] of pkg.prData.changes) {
-          if (path === `${pkg.config.path}/package.json`) {
+          if (
+            fileData.content !== null &&
```
Mesteery, line 54: I added this because `content` can be `null`, but is this really necessary?

package.json
```diff
@@ -81,7 +81,8 @@
     "typescript": "^3.8.3",
     "unist-util-visit": "^2.0.3",
     "unist-util-visit-parents": "^3.1.1",
-    "yargs": "^16.0.0"
+    "yargs": "^16.0.0",
+    "detect-indent": "^7.0.0"
```
Mesteery, line 85: `detect-indent 7.x` requires at least Node.js 12.20, what can we do?

src/plugins/node-workspace.ts
joeldodge79: sorry for the lag - I think I get this approach and it seems good to me. but I need to spend another minute reading it all the way through

package.json
joeldodge79: does typescript have a way of conditionally importing and otherwise stubbing (such that pre-12.20 users just don't get this feature?)

src/plugins/node-workspace.ts
joeldodge79: I'd expect it to always be non-null (which I was just cheating the typing with `fileData.content!`

src/plugins/node-workspace.ts
```diff
@@ -14,33 +14,48 @@
...
+    public readonly rawContent: string,
+    location: string,
+    pkg?: PackageJson
+  ) {
+    super(pkg ?? JSON.parse(rawContent), location);
+  }
+
+  clone() {
+    return new Package(this.rawContent, this.location, this.toJSON());
+  }
```
joeldodge79, line 41: 👍

joeldodge79 (approved):
yeah, I like this `class Package extends LernaPackage` approach

package.json
Mesteery: TypeScript directly, no. But it is always possible to do an if/else. But I think that using the second last version (6.1.0) is a better solution. The last version (7.0.0) only brings some minor and syntax changes and the use of ESM.

bcoe (approved):
This reads a lot better to me, thank you for this refactor 👏

PR #949, opened 2021-07-02
feat: wrap GitHub errors

### LINKED ISSUE #932: Throw better exceptions
Downstream consumers of this library should be able to catch different kinds of exceptions that occur. Doing so will allow developers to react differently to different types of exceptions.

For example, a developer should be able to distinguish between a missing expected file to update from an error like trying to create a GitHub release for a tag that already exists.

- [ ] `HttpError: Validation Failed: {"resource":"Release","code":"already_exists","field":"tag_name"}` when trying to tag a release and the tag already exists

### LINKED ISSUE #945: HttpError: Validation Failed: {"resource":"Release","code":"already_exists","field":"tag_name"}
```
HttpError: Validation Failed: {"resource":"Release","code":"already_exists","field":"tag_name"}
undefined. ( /workspace/node_modules/@octokit/request/dist-node/index.js:86 )
```

From bot logs.

chingor13:
* Adds a `wrapAsync` helper to the `github` module that catches Octokit's `RequestError` and allows us to specify a error handler callback. The fall-through error handler wraps the caught exception in our own Error type so that we control the interface of the errors throw.
* Adds a handler for creating releases that throws a `DuplicateReleaseError` if the release failed because the release tag already exists
* Force updates some octokit dependencies to ensure we get certain error response fields set

Fixes #932
Fixes #945


chingor13:
The alternative is to try to parse the octokit errors in the calling program, but that's exposing the types of a transitive dependency and error prone if for whatever reason the underlying implementation changes.

src/github.ts
```diff
@@ -266,23 +267,25 @@ export class GitHub {
...
-    },
-    maxRetries = 1
-  ) {
-    while (maxRetries >= 0) {
-      try {
-        return await this.makeGraphqlRequest(opts);
-      } catch (err) {
-        if (err.status !== 502) {
-          throw err;
+  private graphqlRequest = wrapAsync(
```
bcoe, line 270: Did you confirm that `wrapAsync` retains stack trace information from the original error?

You might be able to copy the `.stack` property from the Error you're wrapping on the new error object you're creating. Or, alternatively, fiddle with the [captureStackStrace](https://nodejs.org/api/errors.html#errors_error_capturestacktrace_targetobject_constructoropt) method.

test/github.ts
```diff
@@ -809,6 +811,61 @@ describe('GitHub', () => {
...
+        'abc123',
+        'Some release notes',
+        false
+      );
+      await assert.rejects(promise, error => {
+        return error instanceof DuplicateReleaseError;
+      });
+    });
+
+    it('should raise a RequestError for other validation errors', async () => {
```
bcoe, line 846: I would add a unit test that confirms the original stack trace information wasn't lost.

bcoe (changes requested):
A couple nits, looking good overall.

src/github.ts
chingor13: We're now capturing the original error in the wrapped error as `cause?: Error`.

test/github.ts
chingor13: The assertion now ensures that the calling method in the GitHub class is in the stacktrace. In this case, `GitHub.createRelease`.

bcoe (approved):

PR #954, opened 2021-07-07
feat(rust): Support manifest releaser

### LINKED ISSUE #953: Add manifest support Rust releaser

I've noticed release-please got this new manifest releaser and it sounds _exactly_ like what we need.

The docs say only the `node` and `python` releasers are supported (off the top of my head).

I'd like to use the `rust` releaser but.. _I contributed the rust releaser_ about six months ago.

So the problem is I now want to add support for `manifest-pr` / `manifest-release` to the Rust releaser.

I'd like to open a PR to support that in the near future and eventually get it merged.

Merging individual PRs as usual. I don't like that.

 **Additional context**

I've already started work, this can serve as a tracking issue and I'll open the PR when I get it working.

fasterthanlime:

Fixes #953 🦕


fasterthanlime:
Current status: all the right updates are generated (and less code in the Rust releaser, yay! The `ReleasePR` base is really solid).

Unfortunately, there's two things getting in the way:

  * `GitHub.getChangeSet` does not handle multiple "conflicting" updates: if multiple updates touch the same file, the last will overwrite the others. This is easy to fix since updates are not just "here's the new content" but essentially a function of the old content, so we can chain them.
  * The manifest releaser (`manifest.ts`) concatenates together a map of `Change`s instead of `Update`s, so even teaching `getChangeSet` about multiple-updates-per-file won't solve this

My planned changes:

  * [ ] Have `GitHub.getChangeSet` look up in the current changeset when applying updates, so that the "old content" may in fact be the output of a previous `Update`
  * [ ] Have the manifest releaser gather all `Update`s (manifest and package updates) and then do a single call to `GitHub.getChangeSet`, turning them into a `Change` map.

fasterthanlime:
> * Have the manifest releaser gather all `Update`s (manifest and package updates) and then do a single call to `GitHub.getChangeSet`, turning them into a `Change` map.

Mh this is complicated by the fact that the `node-workspaces` plugin is relying on `pkgsWithPRData` having `Update` entries to.. do what the Rust releaser does out of the box. 😢  (upgrade dependency versions for bumped packages)

Not sure how to proceed there. Changing manifest's interface would break `node-workspaces` (and potentially other third-party plugins), but _not_ changing it means making a new plugin for cargo workspaces and that'd be a breaking change since the non-manifest variant of the Rust releaser would no longer update dependencies.

fasterthanlime:
I opted for adding a `forManifestReleaser` option to ReleasePR's constructor so that the rust releaser can know _not_ to update dependencies and the Cargo lockfile, and then I'll write up a `cargo-workspaces` plugin that does what it used to do, but the same way `node-workspaces` does it.

fasterthanlime:
Alright, this is now working pretty well!

When `forManifestReleaser` is set to true, the Rust releaser doesn't touch other packages' cargo manifests or `Cargo.lock`. This is done instead by the `cargo-workspace` plugin, and it even bumps packages when one of its monorepo dependencies updated, just like the `node-workspace` plugin does.

There's a lot of logic in there, so it's going to need good tests, but I'm going to test my fork on a repo of mine for a bit and add tests as I go along.

joeldodge79:
awesome, thanks for tackling this! I like the `forManifestReleaser` approach to continue to support the individual rust releaser functionality. But I still have a question about the [complete correctness](https://github.com/googleapis/release-please/pull/684#issuecomment-772043399) of that approach. Maybe we can consider deprecating the dependency updating in the individual releaser and eventually removing it?

would be great to see tests:
1. modifications to test/manifest.ts adding rust packages to a couple of the tests
2. a test/plugins/cargo-workspace.ts test

fasterthanlime:
> awesome, thanks for tackling this! I like the `forManifestReleaser` approach to continue to support the individual rust releaser functionality. But I still have a question about the [complete correctness](https://github.com/googleapis/release-please/pull/684#issuecomment-772043399) of that approach. Maybe we can consider deprecating the dependency updating in the individual releaser and eventually removing it?

I'd also be in favor of deprecating that in the future (although with a comfortable window for users to upgrade to the manifest releaser).

I migrated one of our less critical monorepos to this (using [my published fork](https://www.npmjs.com/package/ftl-release-please)) and one of the more annoying bits was to create the config *and* manifest myself by searching for package versions through our private crate registry. Before that I was able to keep things DRY by running the single releaser on all packages, iterating through them using `cargo metadata`. But that resulted in a _ton_ of GitHub API calls and things got slow quickly - I can see why the manifest approach is much faster.

---

I have a pretty good idea of the tests / polish / docs needed for this to land, I'll start working on this now.

fasterthanlime:
> But I still have a question about the [complete correctness](https://github.com/googleapis/release-please/pull/684#issuecomment-772043399) of that approach

Ah also, yeah, we definitely noticed that when using it internally. We've definitely encountered cases with an A->B->C dependency graph where C was bumped so B had its "C" dependency updated, but neither A nor B were bumped/released, and we've just been doing dummy commits to poke release-please into releasing them.

I was puzzled by the node-workspaces approach at first but it makes a lot of sense for Rust workspaces too. I still wish it was working on "Updates" (which are trivial to chain) rather than "Changes" (which are text blobs) but I'm happy with the solution I found!

joeldodge79:
> working on "Updates" (which are trivial to chain) rather than "Changes" (which are text blobs)

I used some low-level [lerna](https://github.com/lerna/lerna) packages to manage dependency updates. In particular there's an API for building a "package graph" and it wants the complete contents of the `package.json` (which includes lists of dependencies). In retrospect I could have supplied `Update`s to the plugin and just done something similar to `GitHub.getChangeSet` inside the plugin before building the package graph... If that would work (and you have cycles/interest) I think that would be better (the low level `Changes` is not great to work with). If that is of interest to you then my preference would be to see a separate PR for it (either before or after this one)

fasterthanlime:
> If that is of interest to you then my preference would be to see a separate PR for it (either before or after this one)

I'm interested in this in that I would like to see it happen, but I can't guarantee I have the bandwidth to make it personally happen!

I realized as I was adding tests that for the correct behavior I also had to build a complete dependency graph, hence my low-tech `GraphNode` / `Graph` business in this PR, so, thanks for the tests in `node-workspace`, they were helpful!

> @fasterthanlime thanks for the patch 👍 will work with @chingor13 to review soon.

Thanks! I want to add at least one more test (discontigous dependency graph) and some docs.

fasterthanlime:
I believe this is ready for a first review.

Main changes:

  * The Rust releaser is now leveraging `ReleasePR`'s base methods a lot more: it mostly implements `buildUpdates`, `getPackageName` and uses the defaults for everything else.
    * It's still using its own `commits()` method, unclear if the main one is now suitable and the custom method has become unnecessary — I took a quick look but couldn't tell.
    * It's using a new `ReleasePR` construct option, `forManifestReleaser`, to avoid doing duplicate work with the `cargo-workspace` plugin
  * release-please's second plugin, `cargo-workspace`, is in! It's (mostly) documented in `manifest-releaser.md`.

Main regrets:

  * The graph traversal thing is very adhoc and naive. However I believe other parts of the code (GitHub API calls) would become a bottleneck long before the dependency graph building/traversal does. I have no objection to ripping it out and replacing it with something different (maybe in a subsequent PR?) as I expect a lot of other languages need similar logic. I was not able to use the lerna stuff out of the box, since it expects everything to look like a JS package, which, crates do not.
  * The changelog generation is probably a bit underwhelming, `node-workspace` does it better. This PR could use a second pass on that, or we could decide it's good to merge the initial functionality in first and worry about improving changelogs later.

joeldodge79 (commented):
I'm out on vacation for a week and I haven't read the rust specific logic in-depth but it looks good and seems to be well tested. I'll trust @fasterthanlime to verify it behaves as desired in the wild :-) I can also dig into it more when I get back if desired.

I have reviewed the small touch points on non-rust releaser/plugin files and those look good so I'm going to give my 👍  on this as-is but defer to @bcoe to provide official approval.

fasterthanlime:
Okay, been running this on large monorepos today.

"leaving changelog entries for later" appears to not be an option since "bumps because dependencies were bumped" results in release-please failing to create tags with:

```
✖ Failed to create release for Rust(redacted)@1.6.1: could not find changelog entry corresponding to release PR
```

I'm gonna take another look at that.

fasterthanlime:
> "leaving changelog entries for later" appears to not be an option since "bumps because dependencies were bumped" results in release-please failing to create tags with (changelog entry not found)

Those last changes fixed it! I adjusted the tests (and snapshots) appropriately.

bcoe (approved):
This work looks really solid to me, I like that much of the new logic lives in the plugin, and that our code outside of the plugin seems to shrink a bit (_I guess by virtue of you using more of the methods in the base class_).

👍  from me, will wait on @chingor13's perspective too.

src/index.ts
```diff
@@ -117,6 +117,7 @@ export interface ReleasePRConstructorOptions
   extends ReleasePROptions,
     ReleaserConstructorOptions {
   labels?: string[];
+  forManifestReleaser?: boolean;
```
chingor13, line 120: Can we rename this to what the option is actually trying to do (like `skipDependencyUpdates` or something like that)?

In the future, we (I) would like to make all releasers manifest releasers with a single releaser as an expected use-case. It would also make it a bit clearer what the option is trying to do.

test/plugins/cargo-workspace.ts
```diff
@@ -0,0 +1,823 @@
...
+          .never();
+      }
+    }
+  }
+
+  function makeGraph(nodes: GraphNode[]): Map<string, GraphNode> {
+    return new Map(nodes.map(n => [n.name, n]));
+  }
+
+  describe('run', () => {
```
chingor13, line 129: nit: can we separate the `postOrder` tests into a separate `describe` block?

chingor13 (changes requested):
Thanks this is looking good so far. A tiny nit, but otherwise LGTM.

fasterthanlime:
@chingor13 I addressed both your nits!

chingor13 (approved):

fasterthanlime:
So happy this landed, thanks for the reviews all!

PR #958, opened 2021-07-08
fix: debug logging for node-workspaces plugin

joeldodge79:
When running as a GH action, there are times when the plugin does not
set the version correctly on a package. Specifically, if a package is
getting a version bump only because at least one dependency is version
bumping then we see that it only gets a "patch" bump even if there is a
specific releaseAs config key set for that package.

When I try to reproduce this running the cli locally on my mac laptop
the problem disappears.

### DIFF
--- a/__snapshots__/node-workspace.js
+++ b/__snapshots__/node-workspace.js
@@ -148,12 +148,40 @@ exports['NodeWorkspaceDependencyUpdates run does not update dependencies on preM
     "success"
   ],
   [
+    "node-workspace: @here/pkgA collected for update (dependency-only = false)",
+    "success"
+  ],
+  [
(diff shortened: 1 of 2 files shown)


joeldodge79:
There's some Heisenberg business going on here. With this change in place running as an Action (from https://github.com/google-github-actions/release-please-action/tree/joeldodge-debug-node-workspaces) the issue doesn't occur. Observing it seems to be affecting it :-( (more likely I'm missing some other subtle change).

@bcoe I think it would be good to get this one in - it provides more visibility in general to the behavior of the node-workspace plugin. If the problem crops up again I'll have better info to troubleshoot it

chingor13 (approved):

PR #964, opened 2021-07-12
feat: add new ConfigurationError types so consumers can handle errors

### LINKED ISSUE #960: HttpError: Not Found for missing Java versions.txt file
If a Yoshi Java repo is missing the `versions.txt` file, then it won't be able to run. We should throw a better exception so the caller can decide how to handle this error rather than a generic 404.

chingor13:
Fixes #960

fix: JavaYoshi releasers now throw `MissingRequiredFileError` if missing required `versions.txt` file.


bcoe (approved):

bcoe:
I've been doing a tiny bit of hacking in Node.js this week, noticed they're using a similar pattern to what you're building out:

https://github.com/nodejs/node/blob/master/lib/internal/errors.js

Enumerating all the common errors in an `errors.js` and using errors as a way to control flow throughout the application.

PR #969, opened 2021-07-14
fix(go-yoshi): Release-As should only apply to the correct file scope

### LINKED ISSUE #878: release-please `Release-As` edits multiple releases in mono-repos
## Issue

When working in a mono-repo with several releasable packages the `Release-As` commit feature will apply to multiple releases when it should not.

## Example

A sub-package in google-cloud-go tried to change it's release version via a dummy commit. It did but it also changed the main packages release version as well.

Storage changing release version: https://github.com/googleapis/google-cloud-go/pull/3982
Storage release PR: https://github.com/googleapis/google-cloud-go/pull/3984
Root package release PR: https://github.com/googleapis/google-cloud-go/pull/3983

chingor13:
Fixes #878

### DIFF
--- a/test/releasers/go-yoshi.ts
+++ b/test/releasers/go-yoshi.ts
@@ -132,6 +132,58 @@ describe('GoYoshi', () => {
         const prNumber = await releasePR.run();
         assert.ok(prNumber);
       });
+
+      it('should ignore Release-As for scoped commits', async function () {
+        const releasePR = new GoYoshi({
+          github: new GitHub({owner: 'googleapis', repo: 'google-cloud-go'}),
+        });
+
+        // Indicates that there are no PRs currently waiting to be released:
+        sandbox
+          .stub(releasePR.gh, 'findMergedReleasePR')
+          .returns(Promise.resolve(undefined));
+
+        // Return latest tag used to determine next version #:
+        sandbox.stub(releasePR, 'latestTag').returns(
+          Promise.resolve({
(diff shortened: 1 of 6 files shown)


chingor13:
cc @codyoss

bcoe (commented):
This looks good to me, if I'm understanding properly:

1. we now check for release as after the filter has occurred (which is what yoshi-go uses to reduce the set of commits).
2. we look for release-as in the parsed commit, rather than the original commit object (this has a slightly different structure).

test/releasers/go-yoshi.ts
```diff
@@ -0,0 +1,245 @@
...
+describe('GoYoshi', () => {
+  afterEach(() => {
+    sandbox.restore();
+  });
+  describe('run', () => {
+    it('creates a release PR', async function () {
+      const releasePR = new GoYoshi({
+        github: new GitHub({owner: 'googleapis', repo: 'google-cloud-go'}),
+      });
+
```
bcoe, line 35: I'm glad we now have tests for the yoshi releaser 👏

chingor13:
> 1. we now check for release as after the filter has occurred (which is what yoshi-go uses to reduce the set of commits).
> 2. we look for release-as in the parsed commit, rather than the original commit object (this has a slightly different structure).

Yes, we analyze the commits after they have been filtered and post-processed. We look in the body and footer of the parsed commit.

bcoe (approved):

PR #976, opened 2021-07-19
feat: add krm-blueprint releaser

### LINKED ISSUE #975: Add support for KRM Blueprints
Add support for [KRM Blueprint](https://github.com/GoogleCloudPlatform/blueprints/tree/main/catalog) release type with updater for [attribution annotation](https://github.com/GoogleCloudPlatform/blueprints/blob/9f7aee7dff152d6f87e851d08e68c7092fec6f6c/catalog/project/project.yaml#L22)

bharathkkb:

Fixes #975 🦕

### DIFF
--- a/src/updaters/krm/krm-blueprint-version.ts
+++ b/src/updaters/krm/krm-blueprint-version.ts
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC
+// Copyright 2021 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/updaters/krm/krm-blueprint-version.ts
+++ b/src/updaters/krm/krm-blueprint-version.ts
@@ -35,17 +35,19 @@ export class KRMBlueprintVersion implements Update {
   updateContent(content: string): string {
     // js-yaml(and kpt TS SDK) does not preserve comments hence regex match
     // match starting cnrm/ ending with semver to prevent wrong updates like pinned config.kubernetes.io/function
-    const oldVersion = content.match(/(cnrm\/.*\/)(v[0-9]+\.[0-9]+\.[0-9])+(-\w+)?/)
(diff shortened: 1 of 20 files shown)


src/updaters/krm/krm-blueprint-version.ts
```diff
@@ -0,0 +1,53 @@
...
+    this.create = false;
+    this.path = options.path;
+    this.changelogEntry = options.changelogEntry;
+    this.version = options.version;
+    this.packageName = options.packageName;
+  }
+  updateContent(content: string): string {
+    // js-yaml(and kpt TS SDK) does not preserve comments hence regex match
+    // match starting cnrm/ ending with semver to prevent wrong updates like pinned config.kubernetes.io/function
+    const oldVersion = content.match(
```
morgante, line 38: We might need to be a bit more careful here. In particular, if a package contains a subpackage pulled from another repository (at an older version), we *don't* want to rewrite that version number. Instead I think we might need to just search for the exact previous version and replace with that.

chingor13: It looks like the released package is also on the line to replace, so you might be able to match on that as well.

chingor13 (approved):
Thanks for this - it's looking good. We can tighten up the version replacements in this PR or in a follow up.

src/updaters/krm/krm-blueprint-version.ts
bharathkkb: @morgante @chingor13 I added some logic to use previous tag if available. We still fallback to default matching if no tag available (IIUC, there would be a bootstrapping problem otherwise on first release).

> released package is also on the line to replace

I will investigate this in a follow up. We need to align some of our existing attributions with pkg names in the upstream repo.

morgante: Thanks, this new approach looks workable.

chingor13 (approved):

PR #989, opened 2021-07-30
feat(manifest): support hard-coded last-release-sha

joeldodge79:
allow specifying an arbitrary sha to represent the last release. This is
helpful to use when trying to revert the last release PR. While you can
revert the changes from a bad release PR and delete associated tags/releases,
GitHub does not allow you to delete the actual PR so release-please will
still find it and consider it the last release.

worked for us in the wild: https://github.com/looker-open-source/sdk-codegen/pull/772

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -61,6 +61,7 @@ export interface Config extends ReleaserConfigJson {
   packages: Record<string, ReleaserPackageConfig>;
   parsedPackages: ManifestPackage[];
   'bootstrap-sha'?: string;
+  'last-release-sha'?: string;
   plugins?: PluginType[];
 }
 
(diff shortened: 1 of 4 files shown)


joeldodge79:
just had another thought about this: I suppose a more specific solution to my problem would be supporting a PR label that disqualifies a PR from being considered as a release PR. thoughts?

src/manifest.ts
```diff
@@ -337,11 +338,12 @@ export class Manifest {
     getFileMethod: 'getConfigJson' | 'getManifestJson',
     fileName: string
   ): Promise<{valid: true; obj: object} | {valid: false; obj: undefined}> {
-    let response: {valid: true; obj: object} | {valid: false; obj: undefined} =
-      {
-        valid: false,
-        obj: undefined,
-      };
+    let response:
+      | {valid: true; obj: object}
```
joeldodge79, line 342: something weird going on here. This annotation claims it should be back in its original form. but in its original form gts complains and says that prettier wants it new-lined. ♾️

joeldodge79: btw this is how `npm run fix` leaves it

sofisl: You can try `rm -rf package-lock.json` and `rm -rf node_modules`, then `npm i` and `npm run fix`, see if that fixes your linting issue.

sofisl (approved):

src/manifest.ts
joeldodge79: oh interesting, after doing that `npm run fix` puts it back in the original form and `npm run lint` no longer complains about it. thanks for the tip!

bcoe (approved):

PR #1000, opened 2021-08-04
feat(prerelease): support "normal" form for python pre-releases

bcoe:
This PR adds partial support for Python's "normal" form pre-release tags, by coercing them into the type of pre-release currently supported by release-please (`-1.0.0b2` becomes `-1.0.0-b2`).

Python now also has the ability to handle pre-releases enabled by default, we likely might want to enable this feature for all languages but thought I'd start with one.

We do not fully support the form `1.0.0b2` for release tags, but using `Release-As:` can still be used.

### DIFF
--- a/src/release-pr.ts
+++ b/src/release-pr.ts
@@ -87,6 +87,7 @@ export class ReleasePR {
   pullRequestTitlePattern?: string;
   extraFiles: string[];
   forManifestReleaser: boolean;
+  preRelease = false;
 
   constructor(options: ReleasePRConstructorOptions) {
     this.bumpMinorPreMajor = options.bumpMinorPreMajor || false;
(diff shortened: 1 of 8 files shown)


src/release-pr.ts
```diff
@@ -572,6 +574,10 @@ export class ReleasePR {
     if (!preRelease && version.indexOf('-') >= 0) {
       return null;
     }
+    // Allow the '-' separator to be omitted.
+    if (preRelease && !version.includes('-') && version.match(/[a-zB-Z]/)) {
+      version = version.replace(/([a-zA-Z])/, '-$1');
+    }
```
chingor13, line 580: Won't this also turn `1.0.0-b1` into `1.0.0--b1`?

src/releasers/python.ts
```diff
@@ -48,6 +48,8 @@ const CHANGELOG_SECTIONS = [
 ];
 
 export class Python extends ReleasePR {
+  preRelease = true;
```
chingor13, line 51: It's still unclear to me what this member variable is meant to convey.

chingor13 (changes requested):

src/release-pr.ts
chingor13: Oh, missed the second conditional

chingor13 (approved):

src/releasers/python.ts
bcoe: what if we go with the name `enableSimplePrereleaseParsing`, when we undertake your refactor that divides version bumping logic out into its own methods, we can make this one of the strategies.

PR #1003, opened 2021-08-05
Revert "fix(github-release): private repo not found error"

bcoe:
Refs: https://github.com/googleapis/repo-automation-bots/pull/2334 this change appears to cause issues for some of the releasers.

Reverts googleapis/release-please#982

### DIFF
--- a/src/errors/index.ts
+++ b/src/errors/index.ts
@@ -41,15 +41,6 @@ export class MissingRequiredFileError extends ConfigurationError {
   }
 }
 
-export class PathNotFoundError extends Error {
-  path: string;
-  constructor(path: string) {
-    super(`Could not find requested path: ${path}`);
-    this.path = path;
-    this.name = PathNotFoundError.name;
-  }
-}
-
 export class GitHubAPIError extends Error {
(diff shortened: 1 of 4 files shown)


chingor13 (approved):

bcoe:
@abeMedia with this update the PathNotFound error started throwing for files that are considered optional, such as `package-lock.json`.

bcoe:
@abeMedia I believe the problem is that PathNotFound does not itself have a `code = 404`.

If it were to have a code of `404`, we would not need to update that test adding an additional `nock` test, I don't believe.

abeMedia:
Ah damn! Before it was throwing an error at the same place so I guess the error handling just needs to be updated in some places to deal with the new one. Gonna get on it right away.

abeMedia:
> @abeMedia I believe the problem is that PathNotFound does not itself have a `code = 404`.
> 
> If it were to have a code of `404`, we would not need to update that test adding an additional `nock` test, I don't believe.

Thanks for the heads-up. Will test that!

abeMedia:
Sorry something came up yesterday. Will do another PR over the weekend though.

PR #1013, opened 2021-08-11
fix: stop invalid pages api calls

### LINKED ISSUE #1012: [BUG] keep making api calls even when commits end
#### Environment details

  - OS: ubuntu
  - Node.js version: v14.17.3
  - npm version: 6.14.13
  - `release-please` version: 
     - 11.11.0 from https://github.com/google-github-actions/release-please-action
     - tested locally on current version

#### Steps to reproduce

  1. setup release-please-action on a repo with manifest mode
  2. don't set neither `bootstrap-sha` or `last-release-sha` in the manifest config

#### Related
https://github.com/google-github-actions/release-please-action/issues/311

rabraghib:

Fixes #1012

### DIFF
--- a/src/github.ts
+++ b/src/github.ts
@@ -337,6 +337,7 @@ export class GitHub {
           }
           commits.push([commit.sha, commit.commit.message]);
         }
+        found = response.data.length === 0 || found;
         page++;
       }
       const ret = [];


rabraghib:
@googlebot I signed it!

src/github.ts
```diff
@@ -337,6 +337,7 @@ export class GitHub {
           }
           commits.push([commit.sha, commit.commit.message]);
         }
+        found = response.data.length < per_page || found;
```
bcoe, line 340: I'm not 100% sure that we wouldn't sometimes get smaller page sizes when we haven't iterated over the whole dataset. Perhaps alternatively we could pick a reasonable upper-bound on the response size?

N pages back in time?

bcoe: @abraghib I'd like to test the `response.data.length === 0` idea, but it seems like a reasonable approach.

PR #1019, opened 2021-08-13
feat: handle single quotes in version-rb

### LINKED ISSUE #1018: Support single quoted strings for ruby version.rb

In Ruby, string constants commonly use single (`'`) instead of double (`"`) quotes. This also applies to the version string in `version.rb`: 

```ruby
module Google
  module Cloud
    module Bigtable
      VERSION = '0.5.0'.freeze
    end
  end
end
```

Right now, having a `version.rb` like the above will result in the version file not being updated by `release-please`. The root cause is the regex in `updaters/version-rb.ts` not matching on single quotes: 
```js
/"[0-9]+\.[0-9]+\.[0-9](-\w+)?"/,
 ^
```

`updaters/version-rb.ts` should be updated to also handle single-quoted strings. A simple implementation would look like this: 

```js
return content.replace(
  /(["'])[0-9]+\.[0-9]+\.[0-9](-\w+)?["']/,
  `$1${this.version}$1`
);
```

One could use double quotes in the `version.rb`, even if every other file in the codebase uses single quotes. Apart from having to make exceptions _just_ to be able to use `release-please`, the default configuration for the [Rubocop](https://rubocop.org/) linter demands single quotes in this instance - resulting in the user having to ignore the version file in the configuration.

 **Additional context**
 
This is a port of an issue [I created in the release-please-github-action project](https://github.com/google-github-actions/release-please-action/issues/331). I managed to identify the root cause, and thus I'm recreating it here.

hschne:
This fixes the linked issue by applying a small fix to `updaters/version-rb.ts`. 

Fixes #1018 🦕

### DIFF
--- a/__snapshots__/version-rb.js
+++ b/__snapshots__/version-rb.js
@@ -1,3 +1,29 @@
+exports['version.rb updateContent updates content with single quotes in version.rb 1'] = `
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the 'License');
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an 'AS IS' BASIS,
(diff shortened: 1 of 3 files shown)


bcoe (approved):

bcoe:
@hschne thanks for the patch 👌

PR #1026, opened 2021-08-29
feat(python): support packagename/__init__.py without pyproject.toml

### LINKED ISSUE #1025: Support packagename/__init__.py irrespective of pyproject.toml

Supporting `<projectname>/__init__.py` is nice, but it only happens if there is a `pyproject.toml` (and it's "valid" -- the "validity" is something debatable but let's not get to that here). However the two are not really related/dependent, so it'd be great if the mentioned replacement would happen unconditionally.

Version in `<projectname>/__init__.py` always updated, if there is one.

-

 **Additional context**

-

scop:

Fixes #1025 (eventually)

Some first pseudocodelike steps to fixing #1025. Untested.

### DIFF
--- a/src/releasers/python.ts
+++ b/src/releasers/python.ts
@@ -89,6 +89,7 @@ export class Python extends ReleasePR {
 
     const parsedPyProject = await this.getPyProject();
     const pyProject = parsedPyProject?.project || parsedPyProject?.tool?.poetry;
+    let projectName = packageName.name;
     if (pyProject) {
       updates.push(
         new PyProjectToml({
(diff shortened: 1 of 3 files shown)


chingor13 (changes requested):
This looks like a good start, but we need to adjust the tests accordingly and make sure we have a test case for the new functionality so we don't regress in the future.

scop:
Sure. I'm not sure when/if I'll be taking this further, would appreciate a lot if someone who knows their way around release-please internals would finish it.

bcoe:
@Mesteery if you have the cycles, could you help with this review, what was the reasoning for not updating `__init__.py` if there's no `pyproject.toml` file?

Mesteery:
There is no reason for it. I added it with pyproject and didn't think of generalizing it for the rest.

chingor13 (approved):

PR #1030, opened 2021-08-30
feat: add elixir updater and releaser

### LINKED ISSUE #1029: Add Elixir Support
mix.exs files for elixir projects should be supported to be updated

thepiwo:
Fixes #1029  🦕


thepiwo:
hey @chingor13 @bcoe do you have any feedback for this?

bcoe:
Hey @thepiwo I'll make an effort to give some feedback soon, thank you for the contribution 👏

chingor13 (approved):
Thanks for the contribution!

My only concern is that because the `mix.exs` file is source, you can programatically set that value. For example, in our generated Elixir clients, we use a module attribute: https://github.com/googleapis/elixir-google-api/blob/bb1ee161f6abadbffbd42db1a48f29c6c75ce7d1/clients/analytics/mix.exs#L21-L27

This would be handled by smarter parsing or being opinionated about how the version number must be specified (and documented).

thepiwo:
@chingor13 thanks I understand your concern, there are many ways this version field can be populated, I think for now there is one supported method, maybe others can contribute more ways in the future.

PR #1036, opened 2021-09-07
feat(manifest)!: force local linking in node-workspaces w/ config to disable

### LINKED ISSUE #1035: Plugin node-workspace doesn't update versions falling outside of range
When using the `node-workspace` plugin inside a "manifest"-type project, package versions aren't updated when they fall outside the semver range in a package. This seems to be the desired "safe" behavior, according to the tests. However, some projects need this behavior, as things can be very intertwined.

Ideally, this would just be a configuration settings that can be set to always link all local dependencies before building the dependency graph. I have a patch that I'll submit in a minute.

I considered always setting this value and then filtering dependencies based on their updated version, but that implementation seemed wrought with edge cases.

 **Additional context**
Debugged a bit with @bcoe - this is what I'm trying to achieve: https://github.com/phated/grain/pull/7/files

phated:

Fixes #1035 🦕

### DIFF
--- a/__snapshots__/node-workspace.js
+++ b/__snapshots__/node-workspace.js
@@ -1,4 +1,208 @@
-exports['NodeWorkspaceDependencyUpdates run does not update dependencies on preMajor versions with minor bump changes'] = `
+exports['NodeWorkspaceDependencyUpdates forces local package linking with config changes'] = `
+====================
+{
+  "config": {
+    "releaseType": "node",
+    "packageName": "@here/pkgA",
+    "path": "packages/pkgA"
+  },
+  "prData": {
+    "version": "1.1.2",
+    "changes": {}
+  }
+}
+
+filename: packages/pkgA/package.json
+{
+  "name": "@here/pkgA",
+  "version": "1.1.2",
+  "dependencies": {
+    "@there/foo": "^4.1.7"
+  }
+}
+
+filename: packages/pkgA/CHANGELOG.md
+# Changelog
+
+All notable changes to this project will be documented in this file.### [1.1.2](https://www.github.com/fake/repo/compare/pkgA-v1.1.1...pkgA-v1.1.2) (1983-10-10)
+
+
+### Bug Fixes
+
+* We fixed a bug!
+====================
+{
+  "config": {
+    "releaseType": "node",
+    "packageName": "@here/pkgB",
+    "path": "packages/pkgB"
+  },
+  "prData": {
+    "version": "0.3.0",
+    "changes": {}
+  }
+}
+
+filename: packages/pkgB/package.json
+{
+  "name": "@here/pkgB",
+  "version": "0.3.0",
+  "dependencies": {
+    "@here/pkgA": "^1.1.2",
+    "someExternal": "^9.2.3"
+  }
+}
+
+filename: packages/pkgB/CHANGELOG.md
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+### [0.3.0](https://www.github.com/fake/repo/compare/pkgB-v0.2.1...pkgB-v0.3.0) (1983-10-10)
+
+
+### Features
+
+* We added a feature
+
+
+### Dependencies
+
+* The following workspace dependencies were updated
+  * dependencies
+    * @here/pkgA bumped from ^1.1.1 to ^1.1.2
+
+### [0.2.1](https://www.github.com/fake/repo/compare/pkgB-v0.2.0...pkgB-v0.2.1) (1983-10-10)
+
+
+### Bug Fixes
+
+* We fixed a bug
+====================
+{
+  "config": {
+    "path": "packages/pkgC",
+    "releaseType": "node",
+    "packageName": "@here/pkgC"
+  },
+  "prData": {
+    "version": "3.3.4",
+    "changes": {}
+  }
+}
+
+filename: packages/pkgC/package.json
+{
+  "name": "@here/pkgC",
+  "version": "3.3.4",
+  "dependencies": {
+    "@here/pkgA": "^1.1.2",
+    "@here/pkgB": "^0.3.0",
+    "anotherExternal": "^4.3.1"
+  }
+}
+
+filename: packages/pkgC/CHANGELOG.md
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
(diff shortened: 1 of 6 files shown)


phated:
cc @joeldodge79 - This is the PR that I was talking about in that inline thread. It allows you to force the local linking to avoid the semver range issue. I personally think this should be the default, but understand that it might be too heavy handed.

docs/manifest-releaser.md
```diff
@@ -153,6 +153,12 @@ documented in comments)
   // absence defaults to false and Releases are created as already Published.
   "draft": true
 
+  // when using the `node-workspace` plugin, force the package discovery
+  // to always link local dependencies, even if the SemVer ranges don't match.
+  // can be used to allow breaking version bumps to update during a release.
+  // see Breaking Changes section below
+  // absence defaults to false
+  "always-link-local": true
```
joeldodge79, line 161: cool, I didn't know this feature existed. I think you're right that it should be the default because whenever I have local pkgA crossing a major version boundary, not only does pkgB's dependency version not update, the build fails because it can't find the old version of pkgA.

test/plugins/node-workspace.ts
```diff
@@ -1115,4 +1115,131 @@ describe('NodeWorkspaceDependencyUpdates', () => {
...
+          '/pkgC-v3.3.2...pkgC-v3.3.3) (1983-10-10)' +
+          '\n\n\n### Bug Fixes' +
+          '\n\n* We fixed a bug',
+      ],
+    ]);
+
+    // pkgA had a patch bump and pkgB had a minor bump from
+    // manifest.runReleasers()
+    const newManifestVersions = new Map([
+      ['packages/pkgA', '1.1.2'],
```
joeldodge79, line 1170: maybe it's over-kill but just for my clarification: if you made this change:
```suggestion
      ['packages/pkgA', '2.0.0'],
```
then B's `@here/pkgA` would update to `^2.0.0` as well, correct? In otherwords, it's not just the "pre-major, breaking change version bump" that's handled but also the "post-major, breaking change version bump" (and I suppose the "pre-pre-major, major version bump" but I don't think we need to vet that here)?

joeldodge79 (approved):
this is great, thanks! I've already had to manually fix breaking-change dependency pointers and it's not fun

a/ I agree `always-link-local` should default to true. i don't know if there's a good way to estimate whether we'd really cause any damage just going for it - maybe an "ask for forgiveness" situation.
b/ minor suggestion on the test: I'll leave it up to you if you want to incorporate it.

test/plugins/node-workspace.ts
phated: I believe you are correct here, but the node-workspace plugin makes an assumption that the version update for dependencies should always be `"patch"` here: https://github.com/googleapis/release-please/blob/master/src/plugins/node-workspace.ts#L165

Do you have any ideas how we'd determine the correct bump to make?

test/plugins/node-workspace.ts
```diff
@@ -1115,4 +1115,131 @@ describe('NodeWorkspaceDependencyUpdates', () => {
...
+          '\n\nAll notable changes to this project will be ' +
+          'documented in this file.' +
+          '\n\n### [3.3.3](https://www.github.com/fake/repo/compare' +
+          '/pkgC-v3.3.2...pkgC-v3.3.3) (1983-10-10)' +
+          '\n\n\n### Bug Fixes' +
+          '\n\n* We fixed a bug',
+      ],
+    ]);
+
+    // pkgA had a patch bump and pkgB had a minor bump from
```
joeldodge79, line 1167:
```suggestion
    // pkgA had a patch bump and pkgB had a "breaking change" minor bump from
```

test/plugins/node-workspace.ts
joeldodge79: That particular assumption is the following: unless otherwise bumped by actual conventional commits, a package should only bump by a `patch` since the only thing that's changing about it is dependency versions - even when a dependency version crosses a breaking change boundary. I think that still makes sense: just because a library you're using has some breaking change doesn't imply that the functionality you expose to users of your library has a breaking change. If it did, then you'd most likely have an actual conventional commit against your package with a `feat/fix!`

What I was driving at here was simply verifying that if conventional commits had pushed pkgA to `2.0.0`, then pkgB and pkgC would now point to to `'@here/pkgA': '^2.0.0'`. I'd still expect pkgB to be at `0.3.0` and pkgC to be at `3.3.4` at L1236-40. You'd need to copy/tweak the data for `pkgAdata` from L49 to use on L1174.

But, since you've already verified that crossing the "breaking change" boundary for `0.2.1` -> `0.3.0` for a dependency works then it's probably not necessary so I'll let you decide if you want to include it.

__snapshots__/node-workspace.js
```diff
@@ -1747,6 +1687,125 @@ exports['NodeWorkspaceDependencyUpdates run prefers release-as configuration ove
   ]
 ]
 
+exports['NodeWorkspaceDependencyUpdates run updates dependency to pre-release version changes'] = `
```
phated, line 1690: The upstream diff is really confusing, so I'm commenting on the commit itself.

@joeldodge79 this is the breaking behavior that might be the most confusing and I'd like to get your take. What's happening with "always-link-local" default to `true`, the pre-release versions are being bumped into non-pre-release packages and those are getting a patch bump. Do you agree with this behavior?

phated:
@joeldodge79 I've updated this with the breaking changes. Can you take a look to make sure you agree with the new assumptions if we default this to `true`? 🙏

__snapshots__/node-workspace.js
joeldodge79: Hm, this piece seems less appealing on the surface. But again, at least given how we currently build/publish our packages from looker-open-source/sdk-codegen we'd fail to find `@here/pkgA 1.1.1` since it was bumped to `1.1.2-alpha.0` so we'd need this change. It could be we're doing something wrong with our build/publish where it should still be able to use the previous version even though there's no longer a local representation of it. 

anyway, I think I'm still good with this change!

joeldodge79 (approved):
sorry for the lag and slightly muddle response but I say yes, go for it!

phated:
Thanks @joeldodge79! I'll get this updated and rebased.

PR #1047, opened 2021-09-10
feat: bumpMinorPreMajor works even if > v1.0.0

ilya-lesikov:


ilya-lesikov:
Sorry, not intended to open PR here.

PR #1048, opened 2021-09-12
feat(manifest): add option to skip creating github release

### LINKED ISSUE #1042: Allow skipping release part for a specific package for `manifest` command
To have a config option in `release-please-config.json` for that.
e.g:
```json
{
  "packages": {
    "path/to/pkg": {
      "skip-github-release": true
    }
  }
}
```

rabraghib:

Fixes #1042 🦕

Here is a simple repo that shows the functionality: https://github.com/rabraghib/release-please-tests

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -49,6 +49,7 @@ interface ReleaserConfigJson {
   'bump-patch-for-minor-pre-major'?: boolean;
   'changelog-sections'?: ChangelogSection[];
   'release-as'?: string;
+  'skip-github-release'?: boolean;
   draft?: boolean;
 }
 
@@ -240,6 +241,10 @@ export class Manifest {
             config['release-as']
           ),
           draft: pkgCfg['draft'] ?? config['draft'],
+          skipGithubRelease:
+            pkgCfg['skip-github-release'] ??
+            config['skip-github-release'] ??
+            false,
         };
         packages.push(pkg);
       }
@@ -697,6 +702,14 @@ export class Manifest {
         this.checkpoint(
           `Unable to find last version for ${pkgLogDisp}.`,
           CheckpointType.Failure
+        );
+        releases[pkg.config.path] = undefined;
+        continue;
+      }
+      if (pkg.config.skipGithubRelease) {
+        this.gh.commentOnIssue(
+          `:robot: ${pkgName} not configured for release :no_entry_sign:`,
+          lastMergedPR.number
         );
         releases[pkg.config.path] = undefined;
         continue;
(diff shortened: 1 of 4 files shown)


bcoe:
@rabraghib this looks quite reasonable to me, will let @joeldodge79 chime in too.

rabraghib:
thanks @bcoe for your feedback!!

rabraghib:
@joeldodge79

src/manifest.ts
```diff
@@ -676,15 +681,17 @@ export class Manifest {
...
+      await this.getPackagesToRelease(
+        // use the lastMergedPR.sha as a Commit: lastMergedPR.files will inform
+        // getPackagesToRelease() what packages had changes (i.e. at least one
+        // file under their path changed in the lastMergedPR such as
+        // "packages/mypkg/package.json"). These are exactly the packages we want
+        // to create releases/tags for.
+        [{sha: lastMergedPR.sha, message: '', files: lastMergedPR.files}],
+        lastMergedPR.sha
+      )
+    ).filter(pkg => !pkg.config.skipGithubRelease);
```
joeldodge79, line 694: What do you think about a slightly different approach: rather than just exclude them what if we checked this flag further down and did a `this.gh.commentOnIssue("`:robot: ${pkgName} not configured for release :no_entry_sign:`"). Easier for after-the-fact troubleshooting (rather than looking at git history on `release-please-config.json` and tying the timeline to release creation runs)

I also think it handles this edge case better: all packages are marked as `skip-github-release` - that way we still "release" the PR (remove the `autorelease: pending` label and add the `autorelease: tagged` label). That doesn't really affect functionality (we only ever look at the `lastMergedPR`) but for historical viewing/reporting I think it makes more sense to see a "released" PR and it just so happens that no package was actually released.

joeldodge79 (changes requested):
sorry for the lag (missed the notification). Thanks for the contribution, I like the feature! I have a suggestion (inline) for slightly different behavior, let me know what you think.

Also, could you add a [test or two](https://github.com/googleapis/release-please/blob/bd96093fd7c2d6fc655a6ab43f58a7a3fd4356f3/test/manifest.ts#L1226)?

src/manifest.ts
rabraghib: Yp totally agreed.
I will update it & I will try adding some tests

joeldodge79 (approved):
Looks good! seems like there's a file conflict that needs fixing though

rabraghib:
@joeldodge79 fixed.

bcoe (approved):

bcoe:
@rabraghib thank you for the contribution.

PR #1057, opened 2021-09-16
chore: correct step output key in readme

willarmiros:
Just a quick doc update - seems like the output keyname for a release being created has changed and it wasn't reflected in docs. See this output from `release-please` for evidence: https://github.com/willarmiros/test-package-lock-repo/runs/3617929254?check_suite_focus=true#step:3:4

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,19 +173,19 @@ jobs:
       - uses: actions/checkout@v2
         # these if statements ensure that a publication only occurs when
         # a new release is created:
-        if: ${{ steps.release.outputs.release_created }}
+        if: ${{ steps.release.outputs.releases_created }}
       - uses: actions/setup-node@v1
         with:
           node-version: 12
           registry-url: 'https://registry.npmjs.org'
-        if: ${{ steps.release.outputs.release_created }}
+        if: ${{ steps.release.outputs.releases_created }}
       # if you are using Yarn, substitute the command below with `yarn install --frozen-lockfile`
       - run: npm ci
-        if: ${{ steps.release.outputs.release_created }}
+        if: ${{ steps.release.outputs.releases_created }}
       - run: npm publish
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
-        if: ${{ steps.release.outputs.release_created }}
+        if: ${{ steps.release.outputs.releases_created }}
 ```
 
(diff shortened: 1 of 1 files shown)


willarmiros:
@googlebot I signed it!

README.md
```diff
@@ -173,19 +173,19 @@ jobs:
       - uses: actions/checkout@v2
         # these if statements ensure that a publication only occurs when
         # a new release is created:
-        if: ${{ steps.release.outputs.release_created }}
+        if: ${{ steps.release.outputs.releases_created }}
```
chingor13, line 176: I don't think this is correct: https://github.com/google-github-actions/release-please-action/blob/main/index.js#L56

chingor13 (changes requested):

README.md
willarmiros: I think we were both mistaken - it looks like `releases_created` is actually set as output always: https://github.com/google-github-actions/release-please-action/blob/973b63d3d58b9a5a9fed0dd512501ad5a9ab319b/index.js#L49

Whereas `release_created` wasn't removed as I initially thought, but is only set if you are releasing a package at the root of the repo(?) Regardless, IMO it would be more correct to use the output that is always set in the example code to cover more use cases.

bcoe:
@willarmiros are you using the manifest releaser, and this is why the `releases_created` is the variable you needed?

What about a compromise and updating the documents for the manifest  functionality to include an example of the GitHub action?

willarmiros:
@bcoe ah yes I am using the manifest releaser, I guess the behavior is pretty different from the single-package option. I can try to add a sample action workflow to this PR in https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md.

Given manifest releases would typically be for monorepos, would it be ok to include lerna in the example for running release-please and publishing to NPM?

bcoe:
> Given manifest releases would typically be for monorepos, would it be ok to include lerna in the example for running release-please and publishing to NPM?

@willarmiros for sure, I'd be interested in seeing how you recommend doing this for my own benefit.

bcoe:
@willarmiros please feel free to dust this off if you'd like to add docs about lerna (_I think this would be valuable for folks_).

willarmiros:
@bcoe see #1100

PR #1058, opened 2021-09-16
feat: github release notes optional header/footer

### LINKED ISSUE #1039: Add the input for GitHub action to specify release message pattern
Currently, there is no way to change the release message.
Thus, the user cannot add static text before/after changelog or use bumped version anyhow.

It seems to me a template with variables could help many users. E.g.:
```
- uses: GoogleCloudPlatform/release-please-action@v2
  with:
    release-type: node
    package-name: release-please-action
    release-message: |
      The community keeps growing, and we'd love to see you there!

      - Link to slack
      - Link to tg
      - Link to twitter

      ## Installation

      Download [here](https://storage.com/release/v${VERSION}/app.tar.gz).

      ## Changelog

      ${CHANGELOG}
```

ilya-lesikov:
Add `--notes-header` and `--notes-footer` options for `github-release`
command (adds them to the manifest too), so that the user can pass
text templated with handlebars.js to be added before/after the resulting
GitHub release notes.

This allows customizing GitHub release notes by the user, e.g. adding
release installation instructions at the end of the changelog displayed
on the release page.

Partials exposed to the templating engine to be used by the user:
```
{{> version | tag | githubRepo | githubOwner | changelogPath | PRNumber | PRSha | PRTitle }}
```

Usage:
```shell
release-please.js github-release --notes-header "Hi." --notes-footer "## Documentation:\nhttp://example.org/docs/v{{> version }}/"
```

Resulting GitHub release notes:
```markdown
Hi.

## Features:
...release changelog here...

## Documentation:
http://example.org/docs/v1.0.0/
```

Fixes #1039

Signed-off-by: Ilya Lesikov <ilya@lesikov.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,6 +232,8 @@ release-please github-release --repo-url=googleapis/nodejs-firestore \
 | `--repo-url`      | is the URL of the repository on GitHub.                 |
 | `--token`         | a token with write access to `--repo-url`.              |
 | `--path`          | create a release from a path other than the repository's root |
+| `--notes-header`  | prepend release notes with templated block of text. |
+| `--notes-footer`  | prepend release notes with templated block of text. |
 
 ### Running as a GitHub App
 
--- a/__snapshots__/cli.js
+++ b/__snapshots__/cli.js
@@ -62,6 +62,20 @@ Options:
                                                       [boolean] [default: false]
   --release-label                   set a pull request label other than
                                     "autorelease: tagged"               [string]
+  --notes-header                    optionally add block of text before github
+                                    release notes. Text can be templated with
+                                    handlebars.js syntax. The template will have
+                                    these partials exposed: {{> version | tag |
+                                    githubRepo | githubOwner | changelogPath |
+                                    PRNumber | PRSha | PRTitle }}. Literal line
+                                    feeds replaced with newlines.       [string]
+  --notes-footer                    optionally add block of text after github
+                                    release notes. Text can be templated with
+                                    handlebars.js syntax. The template will have
+                                    these partials exposed: {{> version | tag |
+                                    githubRepo | githubOwner | changelogPath |
+                                    PRNumber | PRSha | PRTitle }}. Literal line
+                                    feeds replaced with newlines.       [string]
 `
 
 exports['CLI flags latest-tag flags 1'] = `
--- a/__snapshots__/release-notes.js
+++ b/__snapshots__/release-notes.js
@@ -98,3 +98,24 @@ exports['extractReleaseNotes php-yoshi extracts appropriate release notes, when 
 </details>
 
 `
+
+exports['generateReleaseNotes renders notes with header and footer 1'] = `
+
+HEADER
+1|999999|pr_title|CHANGELOG.md|googleapis|foo|v1.2.0|1.2.0
+
+
+### Bug Fixes
+
+* candidate issue should only be updated every 15 minutes. ([#70](https://www.github.com/googleapis/release-please/issues/70)) ([edcd1f7](https://www.github.com/googleapis/release-please/commit/edcd1f7))
+
+
+### Features
+
+* add GitHub action for generating candidate issue ([#69](https://www.github.com/googleapis/release-please/issues/69)) ([6373aed](https://www.github.com/googleapis/release-please/commit/6373aed))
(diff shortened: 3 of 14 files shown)


bcoe:
@ilya-lesikov bother you to fix the conflict, we'll work on giving you a review.

ilya-lesikov:
@bcoe fixed.

bcoe:
@ilya-lesikov I discussed this PR with @chingor13, I don't love pulling in templating support for a nice to have feature like this.

Two thoughts:

* could we simplify the approach to not use the handlebars templating engine.
* alternatively, perhaps we could introduce something resembling hooks to `release-please`, which would allow for added behavior like this without adding it to the library itself.

ilya-lesikov:
@bcoe I can render the variables by replacing some static keywords like `{{ $VERSION }}` with the data, would this suffice? I used the handlebars since it's already used by conventional-commits JS libraries under the hood, so I decided it wouldn't hurt to allow people to use the fully-featured templating. I can change it though.

bcoe:
@ilya-lesikov that's a good point (the fact that we already have handlebars as a dependency). The other thing on my mind, is that this adds quite a bit of logic to handle the template expansion.

What if instead we added conditional logic [here](https://github.com/googleapis/release-please/blob/main/templates/template.hbs) and were somewhat opinionated about the format of the footer message?

For your use case, what would you like the footer to look like?

ilya-lesikov:
@bcoe sorry, I didn't have much time recently. I skimmed through the code, and isn't this file https://github.com/googleapis/release-please/blob/main/templates/template.hbs used only in ruby-yoshi? I believe [these are](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-writer/templates) the templates that are actually used.

And I'm afraid this would change the changelog itself too, which is not needed in our particular case. We only need to add a header/footer to the GitHub release message, not the changelog. We'll use it to add short installation instructions and also to provide links to the actual binaries since we do not attach them to the GitHub release, instead we store them in a secure TUF-repo. So we would like not to alter the changelog while changing the GitHub release message, I believe if people need to add some extra info to the GitHub release, it most likely would not be useful in CHANGELOG and vice versa.

ilya-lesikov:
@bcoe Hello, I rebased again. Would the current design suffice? If not, how do you wish to proceed? I still can go for replacing keywords like `{{ $VERSION }}` with data instead of handlebars templating, although we would really prefer to have adequate templating available instead of this. Also, there is a room for features like templating/extending the CHANGELOG.md and PR/commit messages done in the same manner. It would be nicer to have them rendered with handlebars.js instead of plain text. We do not need it right now, but we might in the future.

ilya-lesikov:
ref https://github.com/googleapis/release-please/pull/1087
ref github/feedback#6417

bcoe:
@ilya-lesikov I'll dig into this again next week, I apologize that I've been dragging my feet a little bit on getting this over the finish line. What I think I'm finding challenging, is that it could feel slightly odd that footers and headers are only added to the GitHub release (_not to the PR body itself, or to the CHANGELOG_).  _I'm not sure what the right design is._

Having said this, I understand why you want this functionality. _I've been thinking I'd like similar functionality for adding a footer to PR bodies._

ilya-lesikov:
@bcoe I'm not sure about the design too, but for me it looks like the Release page is for the user, PR is for the developers and changelog is for a subset of both? I mean things that might be useful in Release descriptions:
* short installation instructions, docs
* links to hosted somewhere else or third-party packages/binaries/files
* disclaimers, notes (like short license notes)

usually have no value in a PR and especially in a changelog. Still, I see the value in customizing PR/changelog template, but doing so independently of Release template.

@chingor13 I'm fine with that. I guess I should've opened the issue first and discussed the design instead of making the PR straight away, but whatever.

bcoe:
@ilya-lesikov okay, the big refactor we were working on has finally been finished, appologies that this work got stuck behind our migration to `v13`.

Would you like to dust off this work? @chingor13 where do you think an extension like @ilya-lesikov is proposing would best live now?

ilya-lesikov:
@bcoe Hello, I see there is a major refactor will be required on my part, but I'm afraid I won't have time for this soon, sorry. For now we'll have to stick to the fork, hope I can return to this PR someday.

bcoe:
@ilya-lesikov thanks for your patience, and sorry this forced you onto a fork.

We'll keep your feature request in mind, and if possible try to implement it for you (I know @chingor13 was doing some refactoring specifically to support your use case).

PR #1062, opened 2021-09-20
feat(python): support src/packagename/__init__.py

scop:
For the popular src directory layout,
https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure

Refs #1026 

Basically just monkeying what was done with earlier related python changes recently and git grepping what might need updating without really understanding it all, could very well be that I've missed something.


src/releasers/python.ts
```diff
@@ -115,6 +115,14 @@ export class Python extends ReleasePR {
         packageName: packageName.name,
       })
     );
+    updates.push(
```
bcoe, line 118: Mind adding a `// TODO: figure out refactor that makes logic for updating __init__.py, etc., configurable`

I'm a little worried about the number of optional paths we've added at this point.

bcoe (changes requested):

src/releasers/python.ts
scop: Added.

FWIW after this one's in, pretty much the last one re filenames I'm aware some projects could appreciate would be updating `__version__.py` files anywhere in the tree. That could be handled just like `version.py` currently is.

But at that point I think all the "standard" ways to refer to the version would be addressed and a reasonable alternative to configurability might be to nudge projects to use a more common filename to place their `__version__` in.

Then again while the `__version__` _symbol_ is probably the most prominent, some projects use just `version` or `VERSION`...

bcoe (approved):

PR #1074, opened 2021-09-28
feat: add java-backport releaser

chingor13:
This release strategy always bumps the patch version. We will use it for backporting fixes to `<major>.<minor>.x` branches.


test/releasers/java-backport.ts
```diff
@@ -0,0 +1,630 @@
...
+        readFileSync(
+          resolve('./test/fixtures', 'latest-tag-sp-version.json'),
+          'utf8'
+        )
+      );
+      req.post('/graphql').reply(200, {
+        data: graphql,
+      });
+      const latestTag = await releasePR.latestTag(undefined, true);
+      expect(latestTag!.version).to.equal('1.127.1-sp.1-SNAPSHOT');
```
suztomo, line 596: Can we use a version without "-sp.1" for test data?

test/releasers/java-backport.ts
```diff
@@ -0,0 +1,630 @@
...
+        readFileSync(
+          resolve('./test/fixtures', 'latest-tag-sp-version.json'),
+          'utf8'
+        )
+      );
+      req.post('/graphql').reply(200, {
+        data: graphql,
+      });
+      const latestTag = await releasePR.latestTag();
+      expect(latestTag!.version).to.equal('1.127.0-sp.1');
```
suztomo, line 581: Can we use a version without "-sp.1" for test data?

test/releasers/java-backport.ts
```diff
@@ -0,0 +1,630 @@
...
+    // Indicates that there are no PRs currently waiting to be released:
+    sandbox
+      .stub(releasePR.gh, 'findMergedReleasePR')
+      .returns(Promise.resolve(undefined));
+
+    sandbox.stub(releasePR, 'latestTag').returns(
+      Promise.resolve({
+        name: 'v0.20.3',
+        sha: 'abc123',
+        version: '0.20.3',
```
suztomo, line 67: Can I expect an assertion in which this "0.20.3" becomes "0.20.4" after applying JavaBackport strategy? (If this is not the point of this test case, then never mind)

bcoe (approved):
Can we potentially retire this once we split out bumping logic from release logic this coming quarter?

test/releasers/java-backport.ts
chingor13: I will just remove this test - it was copied from the LTS test. The purpose of this one was to detect the special -sp.1 version scheme which is irrelevant to this releaser.

test/releasers/java-backport.ts
chingor13: For these tests, it's captured in the snapshot file (which records the data sent to the api calls).

chingor13:
> Can we potentially retire this once we split out bumping logic from release logic this coming quarter?

Yes, Java will be the strategy for knowing what files to update and the versioning strategy would be something like "Always bump patch"

suztomo (approved):

__snapshots__/java-backport.js
```diff
@@ -0,0 +1,2338 @@
+exports['JavaBackport creates a release PR against a feature branch: changes'] = `
+
+filename: CHANGELOG.md
+# Changelog
+
+### [0.20.4](https://www.github.com/googleapis/java-trace/compare/0.20.3...v0.20.4) (1983-10-10)
```
suztomo, line 6: I see 0.20.4 here. Thank you.

ralphstone1 (commented):
- Okay

PR #1095, opened 2021-10-26
fix(rust): Don't update dev-dependencies lacking a version key

### LINKED ISSUE #1094: cargo-toml updater should not attempt to update non-existent version field
#### Environment details

  - OS: linux
  - Node.js version: 17.0.1
  - npm version: whichever ships with node 17.0.1
  - `release-please` version: not sure but I looked at the relevant code and it didn't change

#### Steps to reproduce

  1. Have a Cargo.toml with a dev-dependency that has a `path =` but not a `version =
  2. release-please fails with:

```
✔ updating crates/redacted1/Cargo.toml dev-dependencies.redacted2 from undefined to 0.12.3
```

---

That stems from a misunderstanding on my part when doing the initial implementation. dev-dependencies don't need to be published for a crate to be published, and when they're part of a workspace, a path-dependency is good enough, there's no need to specify a version number.

In fact, if a dev-dependency is also a path dependency, it probably shouldn't have a version number, see this discussion in cargo-workspaces: https://github.com/pksunkara/cargo-workspaces/pull/53/files#r735178500 - putting in a version number makes `cargo publish` fail because cargo-workspaces (rightfully) doesn't consider dev-dependencies when building its DAG.

The PR for this will follow shortly.

fasterthanlime:
This is a non-breaking change. If you have a version key for a
dev-dependency that has a path key, release-please will bump it
as before. If you don't have a version key (which is recommended,
and required for cargo-workspaces publish to work, see this discussion:
https://github.com/pksunkara/cargo-workspaces/pull/53/files#r735178500 )
then release-please will no longer fail when trying to process your
workspace.

Closes https://github.com/googleapis/release-please/issues/1094

---

- [x] Appropriate docs were updated (no updates necessary)

Fixes #1094 🦕


bcoe:
@fasterthanlime, @chingor13 was undertaking a major refactor to make release-please more flexible for some advanced features we need to build on top of it (_we're moving towards adding better support for multiple release lines._).

1. Sorry that this review fell on the ground.
2. Would you be up for submitting your work using the new structure of the codebase, the major difference should be that releasers are now moved into "strategies".

chingor13:
Looks like I can't push the commit to resolve the conflicts here. We can change this PR to target the 12.6 branch as a backport and I can apply the diff to the v13 version in a separate PR

bcoe (approved):

PR #1100, opened 2021-11-05
docs: add docs for releasing a monorepo with lerna

willarmiros:
Follow up to #1057, this is a proper example GH Action for publishing several packages in a monorepo using lerna and release-please. It works for us over here: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/.github/workflows/release-please.yml

Here's an example of a successful release: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/696

One feature request while I'm here: if there was an option to disable the release comments per-package to make releases less noisy that would be great!


docs/manifest-releaser.md
```diff
@@ -230,6 +230,57 @@ it fails to find a package's version in the last released manifest content. It
...
+      - name: Checkout Repository
+        if: ${{ steps.release.outputs.releases_created }}
+        uses: actions/checkout@v2
+      - name: Setup Node
+        uses: actions/setup-node@v1
+        if: ${{ steps.release.outputs.releases_created }}
+        with:
+          node-version: 14
+          registry-url: 'https://registry.npmjs.org'
+      - name: Build Packages
```
bcoe, line 268: Is the only requirement that someone have their packages in `packages/`? And lerna bootstrap will link the `file://` paths appropriately?

bcoe (changes requested):
Thank you for adding this, one question for you 👏

docs/manifest-releaser.md
willarmiros: Yes and yes!

https://github.com/lerna/lerna/tree/main/commands/bootstrap#usage

bcoe: That's way simpler than I expected, thank you for the docs.

bcoe (approved):

PR #1104, opened 2021-11-22
refactor!: merge Manifest and standard PR paths

chingor13:
refactor!: releasers are now "strategies", more logic moved into base classes
refactor!: Node 12 is now required
refactor!: manifest is now main entrypoint for release please, and logic is shared between mono-repo/split-repo flow
refactor!: versioning straregy now handled by VersionStrategies rather than regexes.

Summary of what's changed:

1. No more passing around version strings. We introduce a Version class that is basically a data class
2. VersioningStrategies are now responsible for bumping a version given a list of commits
3. Strategies are the previous versions of Releasers and are responsible for building a candidate pull request for that language.
4. Manifest is the main entrypoint and we treat every repository as a Manifest. For the single library repos, we have an adapter that makes a single-component manifest.
5. For multi-component manifests, we split by path and have each configured strategy build a release pull request candidate. Plugins then post-process these candidate pull requests into one or more pull requests. Merging into a single aggregate pull request is handled by a plugin as well.


bcoe (approved):

src/releasers/python.ts
```diff
@@ -1,159 +0,0 @@
...
-import {ReleasePRConstructorOptions} from '..';
-import {
-  parsePyProject,
-  PyProject,
-  PyProjectToml,
-} from '../updaters/python/pyproject-toml';
-import {GitHubFileContents} from '../github';
-import {logger} from '../util/logger';
-
-const CHANGELOG_SECTIONS = [
```
tswast, line 35: I'm curious where the default `CHANGELOG_SECTIONS` are defined now for Python? I used to refer to this page often to know which conventional commits refer to which sections.

tlaziuk: I also was curious about that, I found out that the default sections are defined in another package: https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-conventionalcommits/writer-opts.js#L179-L192

PR #1114, opened 2021-11-24
feat: return path along with created release

chingor13:
fix: return created pull request rather than just the pull request number

This returns the path of the component releases along with the release tag information when building/creating a GitHub release.


bcoe (approved):
Thanks for adding this behavior back.

PR #1120, opened 2021-11-25
feat: add GitHub changelog notes generator

### LINKED ISSUE #1106: Add support for GitHub's automaticly generated release notes
It's possible to hit an API and generate what would be populated for GitHub's automatically generated release notes.

It would be nice to support this, so that folks can use release-please in conjuntion with the GitHub feature.

chingor13:
Fixes #1106 

Allows configuration in the manifest config as `changelog-type: 'github'` or via the CLI `--changelog-type=github`.


src/changelog-notes/github.ts
```diff
@@ -0,0 +1,34 @@
...
+export class GitHubChangelogNotes implements ChangelogNotes {
+  private github: GitHub;
+  constructor(github: GitHub) {
+    this.github = github;
+  }
+  async buildNotes(
+    _commits: ConventionalCommit[],
+    options: BuildNotesOptions
+  ): Promise<string> {
+    return await this.github.generateReleaseNotes(
```
bcoe, line 28: Neat, so this will just grab the release notes generated by the GitHub API?

src/factory.ts
```diff
@@ -111,7 +118,15 @@ export async function buildStrategy(
     bumpMinorPreMajor: options.bumpMinorPreMajor,
     bumpPatchForMinorPreMajor: options.bumpPatchForMinorPreMajor,
   });
-  const strategyOptions: StrategyOptions = {
+  const changelogNotes = buildChangelogNotes({
+    type: options.changelogType || 'default',
+    github: options.github,
+    changelogSections: options.changelogSections,
+    // commitPartial: options.commitPartial,
```
bcoe, line 125: Do we need these commented out, can they just be deleted?

src/changelog-notes/github.ts
chingor13: Yes I added the GitHub doc link to the associated issue.

src/factory.ts
chingor13: We can probably delete. Those are options used by ruby to customize the changelog templates and we considering exposing them to the configuration options but did not.

src/factory.ts
```diff
@@ -111,7 +118,15 @@ export async function buildStrategy(
...
+    // headerPartial: options.headerPartial,
+    // mainTemplate: options.mainTemplate,
```
bcoe, line 127:
```suggestion
```

bcoe (approved):

PR #1123, opened 2021-11-25
fix: Manifest.fromConfig can find latest release version without component

### LINKED ISSUE #1121: Running on non-manifest repository fails to find prior releases
## Problem

Node.js' releaser, when not in manifest mode, uses the old style branch names `release-v1.0.0`, `release-v1.1.0`, etc.

I ran release-please against the yargs repo, [and this was the output](https://github.com/yargs/yargs/pull/2072), due to it failing to find the prior release branches.

## Reproducing

```bash
node ./build/src/bin/release-please release-pr --dry-run  --release-type=node --token=$GITHUB_TOKEN --repo-url=yargs/yargs
```

## Expected Behavior

Historical branch names continue to work.

chingor13:
For strategies that fetch a default component (i.e. node reading from `package.json`), it was trying to find a merged PR matching the component. In this case, it failed to find the merged release PR. In the case of Java, which does not detect a default component, it was working because it would fall back to not having a component.

To fix this, we set the component to compare against to the empty string if configured to not `includeComponentInTag`

Fixes #1121

### DIFF
--- a/src/strategy.ts
+++ b/src/strategy.ts
@@ -115,6 +115,9 @@ export abstract class Strategy {
   ): Promise<Update[]>;
 
   async getComponent(): Promise<string | undefined> {
+    if (!this.includeComponentInTag) {
+      return '';
+    }
     return this.component || (await this.getDefaultComponent());
   }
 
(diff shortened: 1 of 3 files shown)


src/strategy.ts
```diff
@@ -115,6 +115,9 @@ export abstract class Strategy {
   ): Promise<Update[]>;
 
   async getComponent(): Promise<string | undefined> {
+    if (!this.includeComponentInTag) {
```
bcoe, line 118: I'm glad this can be fixed at the `Strategy` level, and that we avoid custom logic in the Node strategy.

bcoe (approved):

PR #1129, opened 2021-11-30
fix: merge manifest release PRs unless separatePullRequests is configured

chingor13:
Also, skip merging if there is only a single configured package for the manifest.

### DIFF
--- a/test/plugins/merge.ts
+++ b/test/plugins/merge.ts
@@ -55,9 +55,9 @@ describe('Merge plugin', () => {
       const plugin = new Merge(github, 'main', {});
       const newCandidates = await plugin.run(candidates);
       expect(newCandidates).lengthOf(1);
-      console.log(newCandidates[0].pullRequest.title);
-      console.log(newCandidates[0].pullRequest.title.toString());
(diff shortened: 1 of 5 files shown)


test/plugins/merge.ts
```diff
@@ -48,13 +48,16 @@ describe('Merge plugin', () => {
...
-    it('ignores a single pull request', async () => {
+    it('merges a single pull request', async () => {
       const candidates: CandidateReleasePullRequest[] = [
         buildMockCandidatePullRequest('python', 'python', '1.0.0'),
       ];
       const plugin = new Merge(github, 'main', {});
       const newCandidates = await plugin.run(candidates);
-      expect(newCandidates).to.eql(candidates);
+      expect(newCandidates).lengthOf(1);
+      console.log(newCandidates[0].pullRequest.title);
```
bcoe, line 58: Looks like some debug logic was left in.

bcoe (changes requested):
Looks like perhaps some debug logic was left in, and I think you just need to run `npm run fix`.

bcoe (approved):

Issue #1088, opened 2021-10-18
Document base classes and interfaces


codyoss:
The model of this project is such that when adding new components you need to extend a base implementation or implement an interface. It would be nice if these foundational classes could be documented so that when a person is adding new components they could read docs on how fields are used instead of having to jump around to different areas in the source to find out this information.

- [x] Plugin
- [x] Strategy
- [x] Updater
- [x] Version
- [x] VersioningStrategy

PR #1146, opened 2021-12-20
fix: fallback to look at releases when looking for latest release

### LINKED ISSUE #1132: Releasing from alternate path fails to find prior release
When releasing from alternate path, with `--monorepo-tags`, `v13` of `release-please` fails to find prior release, resulting in bad release proposal, see:

https://github.com/conventional-changelog/conventional-changelog/pull/874

## Reproducing

```
node ./build/src/bin/release-please.js release-pr --release-type=node --token=$GITHUB_TOKEN --path=packages/conventional-changelog-cli --monorepo-tags --package-name=conventional-changelog-cli --repo-url=conventional-changelog/conventional-changelog --dry-run
```

## Expected Output

Ideally we'd be able to bootstrap the library based on the prior tags, or based on a manifest?

chingor13:
The current method of finding the latest release for non-manifest repositories is to look through latest commits to find a merged PR that looks like a release-please PR for that component. This adds a fallback to consider manually tagged releases if that commit is on the target branch.

Fixes #1132
Towards #1088

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -1007,7 +1007,8 @@ async function latestReleaseVersion(
     candidateReleaseVersions
   );
 
-  return candidateReleaseVersions.sort((a, b) => a.compare(b))[0];
+  // Find largest release number (sort descending then return first)
+  return candidateReleaseVersions.sort((a, b) => b.compare(a))[0];
 }
 
 function mergeReleaserConfig(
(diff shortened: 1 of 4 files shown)


bcoe (commented):
Do we still fallback to `bootstrap-sha` as well, if all methods for looking up a prior release fail?

chingor13:
> Do we still fallback to `bootstrap-sha` as well, if all methods for looking up a prior release fail?

This path is not using the manifest at all -- this is using the CLI with `release-please release-pr --release-type=<type>`.

bcoe (approved):

PR #1151, opened 2021-12-21
fix: only backfill files if requested

### LINKED ISSUE #1147: Make pull request files backfill optional
When fetching the list of commits via GraphQL, we return the list of files touched in the commit. The list of files is only available if there is an associated pull request. We are currently always backfilling the list of commits via the REST API.

In many circumstances the list of files touched is not needed and thus we are potentially making a lot of unnecessary API calls. We should make the helper that fetches the list of commits accept an optional parameter to fill in the list of files touched.

chingor13:
Fixes #1147

### DIFF
--- a/test/github.ts
+++ b/test/github.ts
@@ -292,7 +292,6 @@ describe('GitHub', () => {
 
   describe('commitsSince', () => {
     it('finds commits up until a condition', async () => {
-      sandbox.stub(github, 'getCommitFiles').resolves([]);
       const graphql = JSON.parse(
         readFileSync(resolve(fixturesPath, 'commits-since.json'), 'utf8')
       );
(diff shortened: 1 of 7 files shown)


test/github.ts
```diff
@@ -434,7 +429,9 @@ describe('GitHub', () => {
         commit => {
           // this commit is the 2nd most recent
           return commit.sha === 'b29149f890e6f76ee31ed128585744d4c598924c';
-        }
+        },
+        Number.MAX_SAFE_INTEGER,
+        true
```
bcoe, line 434: nit: this method takes enough arguments it might worth considering an interface.

bcoe (approved):

test/github.ts
chingor13: I agree, it's a code smell to have to provide the same default value. I'll follow up with a refactor PR.

PR #1152, opened 2021-12-21
fix(rust): Don't update dev-dependencies lacking a version key (#1095)

### LINKED ISSUE #1094: cargo-toml updater should not attempt to update non-existent version field

chingor13:
Cherry pick #1095 into v13 and resolve conflicts

Fixes #1094

### DIFF
--- a/__snapshots__/cargo-toml.js
+++ b/__snapshots__/cargo-toml.js
@@ -10,6 +10,7 @@ normal-dep = "1.2.3"
 
 [dev-dependencies]
 dev-dep = { version = "1.2.3" }
+dev-dep-2 = { path = "../dev-dep-2" }
 
 [build-dependencies]
 # this is using a private registry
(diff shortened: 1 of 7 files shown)


bcoe:
CC: @fasterthanlime

fasterthanlime:
LGTM! Thanks for doing the work to get this merged, I've been really busy!

bcoe (approved):

PR #1157, opened 2021-12-21
feat: introduce generic updater

### LINKED ISSUE #305: develop more generic "update" class for long-tail of version replacements

### LINKED ISSUE #435: Have the ability to insert major, minor and revision constants into source files.
It should be possible to access the version of a library in code. Currently this is (afaik) only possible in java with https://googleapis.dev/java/gax/latest/com/google/api/gax/core/GaxProperties.html#getLibraryVersion-java.lang.Class-, which reaches into the maven metadata to do so.

It would be useful to be able to paste MAJOR_VERSION = x MINOR_VERSION = y, REVISION = z constants in each language, or even just the numbers representing them by regex substitution into source files upon release.

### LINKED ISSUE #1139: Add a generic script-based release strategy
It would be useful to be able to launch a custom script from the GitHub repo to update specific files without having to create a new type of release strategy in release-please.

Something like:

```
 jobs:
   release-please:
     runs-on: ubuntu-latest
     steps:
       - uses: GoogleCloudPlatform/release-please-action@v2
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           release-type: exec
           command: update-my-files.sh
```

The script would then receive the version number as parameter and be able to perform actions on the repository before release-please commits the files to the branch.

### LINKED ISSUE #1174: document generic releaser
Add docs in customizing.md for how to update long-tail files.

chingor13:
This introduces a new `Generic` updater class that scans the file line by line for annotations.

You can annotate a line (inline) via:
* `x-release-please-version`
* `x-release-please-major`
* `x-release-please-minor`
* `x-release-please-patch`

For these annotations, we will try to replace the value on that line only.

You can annotate a block by starting with a line containing:
* `x-release-please-start-version`
* `x-release-please-start-major`
* `x-release-please-start-minor`
* `x-release-please-start-patch`

and close the block with a line containing `x-release-please-end`. Within the block, we will attempt to replace version values.

Additionally, all basic strategies now support the `extra-files` option which will apply this generic updater class.

Fixes #435
Fixes #305
Fixes #1139
Fixes #1174

### DIFF
--- a/__snapshots__/generic.js
+++ b/__snapshots__/generic.js
@@ -1,0 +1,42 @@
+exports['Generic updateContent updates generic version markers 1'] = `
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.example;
+
+public final class Version {
+  // {x-release-please-start-version}
+  public static String VERSION = "2.3.4
+  // {x-release-please-end}
+
+  // {x-release-please-start-major}
+  public static String MAJOR = "2";
+  // {x-release-please-end}
+
+  // {x-release-please-start-minor}
+  public static String MINOR = "3";
+  // {x-release-please-end}
+
+  // {x-release-please-start-patch}
+  public static String PATCH = "4";
+  // {x-release-please-end}
+
+  public static String INLINE_VERSION = "2.3.4
+  public static String INLINE_MAJOR = "2"; // {x-release-please-major}
+  public static String INLINE_MINOR = "3"; // {x-release-please-minor}
+  public static String INLINE_PATCH = "4"; // {x-release-please-patch}
+}
+
+`
--- a/__snapshots__/generic.js
+++ b/__snapshots__/generic.js
@@ -18,7 +18,7 @@ package com.google.cloud.example;
 
 public final class Version {
   // {x-release-please-start-version}
-  public static String VERSION = "2.3.4
+  public static String VERSION = "2.3.4";
   // {x-release-please-end}
 
   // {x-release-please-start-major}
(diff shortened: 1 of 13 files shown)


src/updaters/composite.ts
```diff
@@ -40,3 +40,27 @@ export class CompositeUpdater implements Updater {
     return content || '';
   }
 }
+
+export function mergeUpdates(updates: Update[]): Update[] {
+  const updatesByPath: Record<string, Update[]> = {};
```
kalosisz, line 45: Would it make sense to have `Record<Update['path'], Update[]>`?

__snapshots__/generic.js
```diff
@@ -0,0 +1,42 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.example;
+
+public final class Version {
+  // {x-release-please-start-version}
```
bcoe, line 20: will this work for other style comments too, e.g., `#`, `/*`?

bcoe (commented):
I like the idea of annotations. In practice, has is the list of files provided to update?

__snapshots__/generic.js
chingor13: Yes, we only look for the x-release-please part.

chingor13:
> I like the idea of annotations. In practice, has is the list of files provided to update?

For the generic updater, this will be provided adhoc via the `extraFiles` factory option or `extra-files` CLI/manifest option.

src/updaters/composite.ts
chingor13: That would probably work in case update's path's type changes, but I think for readability, `string` might be easier to read.

bcoe (changes requested):
I like this idea, but I think it would be good to ad a `docs/customizing-releases.md` or something along those lines, where you speak to how someone can configure and use this functionality.

chingor13:
> I like this idea, but I think it would be good to ad a `docs/customizing-releases.md` or something along those lines, where you speak to how someone can configure and use this functionality.

Opened: #1174 to follow up with

chingor13:
@bcoe Added documentation in docs/customizing.md

bcoe (approved):
Seems like a great addition to functionality, and we can evolve the feature over time if we find supporting JSON files is a requirement.

PR #1160, opened 2021-12-21
fix: fallback to look at tags when looking for latest release

chingor13:
Similar to #1146

If we don't find a recent release PR or release, fallback to parsing tags. This should only be necessary for non-manifest releases that are being bootstrapped to release-please.


bcoe (commented):
Will it check releases and then tags, what's the order of resolution?

chingor13:
To find the last released version, it checks recent merged commits/PRs, then releases, then tags. After a release-please release has been created, it will continue to find the recent merged release PR.

bcoe (approved):

PR #1172, opened 2021-12-23
feat: add arbitrary file updater

### LINKED ISSUE #1171: Add ability to update arbitrary JSON files

I would like to have an arbitrary file updater.

This updater should be able to replace content from a version to another, without any extra regex.

For example, replace all 0.1.0 to 0.2.0 in a file

goatwu1993:

Fixes #1171 🦕

### DIFF
--- a/src/updaters/arbitrary-file.ts
+++ b/src/updaters/arbitrary-file.ts
@@ -1,0 +1,43 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


chingor13:
Thanks for building this.

The reason we didn't build a version replacer like this is that it can replace content in unintended places. If you happen to have another string in the file that matches the previous version (perhaps a version for a different dependency), it will update that version as well. This is probably an ok risk for most files, but the edge cases will catch people off guard.

goatwu1993:
Yeah i totally agree, and I just found the pr which provide adhoc annotation, which will probably solve my use cases.

However, a json file, AFAIU, does not allow annotating or commenting. 

I believe arbitrary updater make sense in these worst case senario? Please correct me if I am wrong

bcoe:
@goatwu1993 @chingor13 it would be interesting to figure out how to solve the `.json` file use case.

goatwu1993:
I guess JMESPath, jq or similar JSON Path parsing library should work for json.

I will close this PR since I get the idea that an updater should have the least parsing logic to avoid unintentional content change.

PR #1181, opened 2021-12-30
feat: allow prerelease releases on Github

### LINKED ISSUE #1180: Allow to release prerelease on Github
I would like *release-please* to be able to release prerelease on Github next to draft and full releases. 

*release-please* should be able to release prerelease for those versions that have the prerelease part (e.g. beta) present if a `prerelease` option is specified. 

*release-please* could release all releaseses as prerelease if a `prerelease` option is specified.

 **Additional context**
This feature would allow more flexibility on Github side since one could create Github Actions triggered specifilly by prerelease releases.

kalosisz:

Fixes #1180  🦕


bcoe:
Thank you for the PR @kalosisz 👏 might take a few days to review due to NY but excited for this.

src/manifest.ts
```diff
@@ -229,6 +234,7 @@ export class Manifest {
     this.bootstrapSha = manifestOptions?.bootstrapSha;
     this.lastReleaseSha = manifestOptions?.lastReleaseSha;
     this.draft = manifestOptions?.draft;
+    this.prerelease = manifestOptions?.prerelease;
```
chingor13, line 237: I don't think we need to capture this one because it's configured on the per-component config (the default is pulled from the root of the manifest config on new line 1078)

src/manifest.ts
```diff
@@ -715,6 +721,9 @@ export class Manifest {
             path,
             pullRequest,
             draft: config.draft ?? this.draft,
+            prerelease:
+              (config.prerelease ?? this.prerelease) &&
```
chingor13, line 725: same as above, we shouldn't need to have `this.prerelease`

chingor13 (changes requested):
Thanks and this is looking great!

If I'm reading this right, it's not affecting the commits we consider since the last release. Are you intending to use this on a separate release branch? If not, when it comes time to release a stable version, it won't grab all the commits since the last stable version - it will grab since last release of any type on the release branch. (FWIW, we are probably going to be maintaining a `preview` branch in the near future which runs in parallel to the `main` branch and would probably use this `--prerelease` option).

src/manifest.ts
kalosisz: it was a bit confusing for me and for this part, I was trying to follow the logic of `draft`. In that case, should the line 236 also be removed?

src/manifest.ts
kalosisz: I agree. Would the same apply to line 723?

kalosisz:
@chingor13 you are right. it would always consider the last release if it's not releaseing from a different branch. 

I think it would be nice if one could specify to only consider stable releases (only storing those in the `manifest.json` or when looking for the latest release) as well. I could add such a feature if one's using the `--prerelease` option. I am a bit unsure of the builds though. Should they be considered stable?

src/manifest.ts
chingor13: Yes, we should clean that up, but maybe in a separate cleanup PR?

src/github.ts
```diff
@@ -628,27 +635,28 @@ export class GitHub {
...
+                  tag {
+                    name
+                  }
+                  tagCommit {
+                    oid
+                  }
+                  url
+                  description
+                  isDraft
+                  isPrerelease
```
bcoe, line 651: Is this the only addition to the graphql that you're asking for `isPrerelease`, are the rest of the updates reformating?

src/github.ts
```diff
@@ -109,6 +109,7 @@ interface GraphQLRelease {
   url: string;
   description: string;
   isDraft: boolean;
+  isPrerelease: boolean;
```
bcoe, line 112: What is this value used for? from the tests it looks like this feature createss pre-releases, but it didn't seem to do any filtering of results based on this variable, or return it to the user (I might just be missing where it's used).

src/manifest.ts
kalosisz: I looked into the case with having `draft` and `draftPullRequest` on `Manifest`. my conclusion is the order should be `this.draft ?? config.draft` since some `cli` commands can take `--draft` and `--draft-pull-request` in addition to a manifest file.  Eg. here https://github.com/googleapis/release-please/blob/580f6290e89da56f4e5afc97d8c25fdccbabca7a/src/bin/release-please.ts#L424-L430

I guess then the flag should override the behaviour from the config file (at least what I infered from the tests). I created a pr on my fork to have a better look on my mods there https://github.com/kalosisz/release-please/pull/7

Did I understend correctly the role of the two flags?

src/github.ts
kalosisz: I meant to make it consistent with having the draft in the query. I'm gonna better remove it

src/github.ts
kalosisz: Adding to the comment above, it's currently not used. It seems the `isDraft` of `GraphQLRelease` and `draft` of `GitHubRelease` is not used either. Maybe a cleanup for those in a separate PR?

kalosisz: I wanted to make it consistent with `isDraft`. I'm gonna remove `isPrerelease` from `GraphQLRelease` as well as `prerelease` from `GitHubRelease` as they aren't used downstream. Do you think the `isDraft` and `draft` could be also cleaned up? it's also not used anywhere.

src/github.ts
kalosisz: yes. I'm gonna revert the reformating.

src/manifest.ts
kalosisz: done

src/manifest.ts
kalosisz: I updated the logic to only consider the `config`

kalosisz:
hey @chingor13. I was thinking on having to generate the release notes on a prerelease from a stable. I reckon the library could have another flag for that for if one has a preview branch and only want to create release notes between prereleases, it 
wouldn't be possible. I would leave this `prerelese` flag as it is now, to only create a Github prerelease.

What do you think?

ChipWolf:
How would prerelease versioning work?

kalosisz:
Versionings are managed in `strategies` and `versioning-strategies`. They might need to be extended.

chingor13 (approved):
I might be ok with having it always create the release as `prerelease` on GitHub. A repo maintainer may want to turn on that flag for pre-1.0.0 versions.

kalosisz:
@chingor13 So far I was looking only at the prerelease part, but based on your last comment, I added releasing pre-major as prereleaes in https://github.com/googleapis/release-please/pull/1181/commits/6f3b7e5e901c852cca18a23244cca5bbfd4d6014

I see my new commits didn't remove your approval? Is this intentional?

chingor13 (approved):

PR #1184, opened 2021-12-31
feat: add grouped manifest pr title pattern option

### LINKED ISSUE #1072: Manifest releaser: Customize PR title
When using the Manifest releaser, the PR title is always `chore: release main`. It seems like the PR title is configurable for the non-Manifest releaser. We'd like to customize the title of our PRs in a similar fashion.

It'd be ideal to specify a PR title format in the release-please manifest file.

I've also considered writing a custom workflow that updates the PR title after release-please runs.

 **Additional context**
Here's a Grain release PR: https://github.com/grain-lang/grain/pull/872 - Notice that we have to remember to update the title to `chore: Release Grain vX.X.X` before we merge.

axieum:
This adds a new `group-pull-request-title-pattern` option to the manifest config to adjust the pull-request title when multiple packages are grouped/merged together into one manifest release pr.

The `${scope}`, `${component}`, `${version}` args for the pattern are inherited from the root package **if present,** i.e. `"path": "."`.

By giving this option a different name at the root of the config, it won't conflict with default `pull-request-title-pattern` for each package.

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/release-please/issues/new/choose) before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea

Closes #1072 🦕


docs/manifest-releaser.md
```diff
@@ -190,7 +190,19 @@ documented in comments)
   // setting to false will only bump your local dependencies within the SemVer range.
   // see Breaking Changes section below
   // absence defaults to true
-  "always-link-local": false
+  "always-link-local": false,
+
+  // if true, create separate pull requests for each package instead of a
+  // single manifest release pull request
+  // absence defaults to false and one pull request will be raised
+  "separate-pull-requests": false,
```
axieum, line 198: Add this option as it was missing.

docs/manifest-releaser.md
```diff
@@ -190,7 +190,19 @@ documented in comments)
...
+
+  // sets the manifest pull request title for when releasing multiple packages
+  // grouped together in the one pull request.
```
axieum, line 201:
```suggestion
  // grouped together in the one pull request.
  // this option has no effect when `separate-pull-requests` is `true`.
```

src/manifest.ts
```diff
@@ -115,6 +115,7 @@ export interface ManifestOptions {
   releaseLabels?: string[];
   draft?: boolean;
   draftPullRequest?: boolean;
+  groupPullRequestTitlePattern?: string;
```
axieum, line 118: I decided against naming it `mergedPullRequestTitlePattern` due to the keyword `merge` being ambiguous in this context

src/manifest.ts
```diff
@@ -230,6 +233,8 @@ export class Manifest {
     this.lastReleaseSha = manifestOptions?.lastReleaseSha;
     this.draft = manifestOptions?.draft;
     this.draftPullRequest = manifestOptions?.draftPullRequest;
+    this.groupPullRequestTitlePattern =
+      manifestOptions?.groupPullRequestTitlePattern;
```
axieum, line 237: I did not update jsdoc for this constructor as it already appears out-of-date.

src/manifest.ts
```diff
@@ -545,7 +550,12 @@ export class Manifest {
     // pull requests
     if (!this.separatePullRequests) {
       plugins.push(
-        new Merge(this.github, this.targetBranch, this.repositoryConfig)
+        new Merge(
+          this.github,
+          this.targetBranch,
+          this.repositoryConfig,
+          this.groupPullRequestTitlePattern
```
axieum, line 557: I think passing the `manifestOptions` to all plugins would be cleaner, but in this `Manifest` constructor we don't capture the `manifestOptions` arg on this instance.

src/strategies/base.ts
```diff
@@ -214,7 +214,7 @@ export abstract class BaseStrategy implements Strategy {
       this.includeComponentInTag ? component : undefined,
       this.tagSeparator
     );
-    logger.warn('pull request title pattern:', this.pullRequestTitlePattern);
+    logger.debug('pull request title pattern:', this.pullRequestTitlePattern);
```
axieum, line 217: I believe this was set to `warn` by mistake

test/manifest.ts
```diff
@@ -1248,6 +1248,117 @@ describe('Manifest', () => {
       expect(pullRequests[0].version?.toString()).to.eql('1.0.0');
     });
 
+    it('should allow customizing pull request title', async () => {
```
axieum, line 1251: Perhaps this test could be simplified a bit, it looks a bit too long 🤔

src/plugins/merge.ts
```diff
@@ -41,20 +56,26 @@ export class Merge extends ManifestPlugin {
...
+    let rootRelease: CandidateReleasePullRequest | null = null;
     for (const candidate of candidates) {
       const pullRequest = candidate.pullRequest;
       rawUpdates = rawUpdates.concat(...pullRequest.updates);
       for (const label of pullRequest.labels) {
         labels.add(label);
       }
       releaseData.push(...pullRequest.body.releaseData);
+      if (candidate.path === '.') {
+        rootRelease = candidate;
```
axieum, line 68: Since we're already iterating over each candidate pr, if we find the root package, we'll remember it for later when we build the pr title

src/plugins/merge.ts
```diff
@@ -41,20 +56,26 @@ export class Merge extends ManifestPlugin {
...
     }
     const updates = mergeUpdates(rawUpdates);
 
     const pullRequest = {
-      title: PullRequestTitle.ofTargetBranch(
+      title: PullRequestTitle.ofComponentTargetBranchVersion(
+        rootRelease?.pullRequest.title.component,
         this.targetBranch,
-        MANIFEST_PULL_REQUEST_TITLE_PATTERN
+        rootRelease?.pullRequest.title.version,
```
axieum, line 77: If we found a root package release pr, we'll "hoist" its component and version args to this title pattern

src/manifest.ts
axieum: Open to suggestions, e.g. `pull-request-group-title-pattern`, etc.

kalosisz:
how is the `pullRequestTitlePattern` set if the manifest is created using `Manifest.fromManifest` (ie from manifest config file)?

axieum:
> how is the `pullRequestTitlePattern` set if the manifest is created using `Manifest.fromManifest` (ie from manifest config file)?

Thanks, should be fixed in 6aae04ec769508bef240af99d6e1709c9fc9146d

src/manifest.ts
```diff
@@ -1068,6 +1079,9 @@ function mergeReleaserConfig(
     packageName: pathConfig.packageName ?? defaultConfig.packageName,
     versionFile: pathConfig.versionFile ?? defaultConfig.versionFile,
     extraFiles: pathConfig.extraFiles ?? defaultConfig.extraFiles,
+    pullRequestTitlePattern:
+      pathConfig.pullRequestTitlePattern ??
+      defaultConfig.pullRequestTitlePattern,
```
axieum, line 1084: This should fix the default pr title pattern not copying to each package

chingor13 (changes requested):
Thanks! This is looking very good.

One question: what happens when there is no root component to hoist? Can we add a test case for this scenario?

axieum:
> Thanks! This is looking very good.
> 
> One question: what happens when there is no root component to hoist? Can we add a test case for this scenario?

@chingor13 I've added commit d872f56be7fd3a0f2025c8b9728f62a04bd5fdb3 which adds a test case, but unsure of how it should behave. Right now it just replaces them with blank.

chingor13 (approved):

Issue #1141, opened 2021-12-12
How to release via Manifest config w/ additional files?


axieum:
Hi,

I have a Java (w/ Gradle) project acting as a mono-repo. I wish to increment the version of each submodule/package individually, which thus far appears to be "partially" working via the `manifest-pr` and `manifest-release` commands.

Please find the affected [.github/workflows/release.yml](https://github.com/axieum/fabric-release-please/blob/12479817b79c3ec3d05e06b454ef33d61109bc73/.github/workflows/release.yml), [release-please-config.json](https://github.com/axieum/fabric-release-please/blob/12479817b79c3ec3d05e06b454ef33d61109bc73/.github/release-please.json), and resulting pull request axieum/fabric-release-please#6.

1. The commit to be released does not appear in **its** package's section, I would expect it to be in both
2. It does not appear to know how to handle `extra-files: gradle.properties` argument, and hence any additional files are not included in the pull request, see [example-submod/gradle.properties#L5](https://github.com/axieum/fabric-release-please/blob/12479817b79c3ec3d05e06b454ef33d61109bc73/example-submod/gradle.properties#L5)
3. It fails to read the `release-type` from the manifest config when running `manifest-release`, but `manifest-pr` picks it up okay

If these are indeed three unique issues/bugs, I can split this issue up.

I understand this project is under active development for a major milestone, but some guidance would be greatly appreciated.

Thanks,

axieum:
Just seen that the `manifest-*` commands will be deprecated.

I will come back to this after [google-github-actions/release-please-action](https://github.com/google-github-actions/release-please-action) has been updated to reflect the `config-file` and `manifest-file` arguments.

chingor13:
In v3, manifest should still work, except the CLI command is rolled into the same `release-pr` and `github-release` commands. Also, in v13, the manifest config should support `extra-files` for each component in the manifest.

I think @bcoe has a candidate version of the GitHub action that is using the v13 release candidate version of this library if you're interested in trying it out.

chingor13:
Digging into your config, you are using the `java-yoshi` strategy which was built to support our custom repository structure and not necessarily meant for public consumption. It is documented [here](https://github.com/googleapis/release-please/blob/main/docs/java-releases.md)

The `java-yoshi` strategy already handles multiple artifact versions (at the same or different versions). You will need to create a `versions.txt` file at the root of the component ([example](https://github.com/googleapis/java-asset/blob/main/versions.txt)).

[...]

PR #1187, opened 2022-01-02
fix: extra file path relativity

axieum:
This makes any `extra-files: ['file', '../x']` relative to the package/strategy path, allowing shared files at the root of the manifest to be applied to each package, e.g.

```json
{
  "extra-files": ["gradle.properties"],
  "packages": [
    {"path": "."},
    {"path": "pkg/a"},
    {"path": "pkg/b"}
  ]
}
```

to result in the following files being updated:

```
gradle.properties
pkg/a/gradle.properties
pkg/b/gradle.properties
```

Discovered in #1141 🕵️‍♂️


src/strategies/base.ts
```diff
@@ -411,16 +410,14 @@ export abstract class BaseStrategy implements Strategy {
...
-    }
-    file = file.replace(/^[/\\]/, '');
-    if (this.path === undefined) {
-      return file;
-    } else {
-      const path = this.path.replace(/[/\\]$/, '');
-      return `${path}/${file}`;
-    }
+    return this.path && this.path !== ROOT_PROJECT_PATH
+      ? path.join(this.path, file)
```
axieum, line 420: Rather use Node's cross-platform path joining.

chingor13 (approved):
Thanks!

axieum:
Just seen that the tests are failing on Windows. 🙄 I'll try look into it tonight.

axieum:
Okay, all tests should be passing for Windows again in 12d9f80711b09a5e9d01a0d03aa03efc06c7b3bc.

Also, in the current state (and unchanged in this PR), it is possible for files to be edited outside of the git project folder, e.g. `extra-files: ["~/.m2/"]`. I assume GitHub Actions are secure, but should this maybe be looked at in a separate issue? Maybe I'm just being too cautious.

chingor13 (changes requested):
> Also, in the current state (and unchanged in this PR), it is possible for files to be edited outside of the git project folder, e.g. `extra-files: ["~/.m2/"]`. I assume GitHub Actions are secure, but should this maybe be looked at in a separate issue? Maybe I'm just being too cautious.

I think we always assumed paths were relative to the repository root, if we're allowing "absolute paths", they should probably be relative to the git root (`/path/to/foo` would actually translate to `<gitroot>/path/to/foo`. Can we restrict the absolute paths?

axieum:
Sure, I'll see what I can do to restrict the various path shortcuts like `~`, `../../../../`, etc.

axieum:
> 

So `/path` after this PR is merged will indeed point to the git root. I just need to account for and strip any leading `../`, `~/`, etc.

src/strategies/base.ts
```diff
@@ -416,12 +416,22 @@ export abstract class BaseStrategy implements Strategy {
    * @returns {string} The file relative to the strategy.
    */
   protected addPath(file: string) {
-    // There is no strategy path to join, or the strategy is at the root, or
-    // the file is at the root, as denoted by a leading slash
-    if (!this.path || this.path === ROOT_PROJECT_PATH || file.startsWith('/')) {
-      return path.posix.normalize(file.replace(/^\/+/, ''));
+    // Strip any trailing slashes - we can't update directories
```
kalosisz, line 419: isn't this stripping leading slashes?

src/strategies/base.ts
```diff
@@ -411,16 +410,22 @@ export abstract class BaseStrategy implements Strategy {
...
-      return file;
-    } else {
-      const path = this.path.replace(/[/\\]$/, '');
-      return `${path}/${file}`;
+    // Otherwise, the file is relative to the strategy path
+    else {
+      file = path.posix.join(this.path, file);
     }
+    // Strip any leading pathings (i.e. ~/, ../, /) and any trailing slashes and return
+    return file.replace(/^(\.{1,2}|~|\/)*\//, '').replace(/\/+$/, '');
```
axieum, line 429: At this point the `file` has been normalized (in both invocations of `path.posix.normalize` and `path.posix.join`, e.g.

* `foo/bar/../baz` -> `foo/baz`
* `foo/bar/../../baz` -> `baz`
* `foo/bar/../../../../baz` -> `../../baz`

We can then safely strip any leading `../`, `/`, or `~/`, for example `../../baz` -> `baz` which would be at the root of the git repository.

image

kalosisz: shouldn't such paths be skipped?

axieum: We could log a warning message and skip these files, but I think we could **only do this for extra files** updates - anywhere else `Strategy#addPath` is used (which appears to be a lot) probably aren't ready to receive `null` or catch an exception.

axieum: This was my main reasoning for just silently ignoring any leading pathing - at least on Linux,
```
[jonathan@Alpine /]$ cd ../../../..
[jonathan@Alpine /]$ pwd
/
[jonathan@Alpine /]$ cd ~/../../../../../../../.././././../../
[jonathan@Alpine /]$ pwd
/
```

chingor13: I feel like using the path module is introducing some extra complexity and cases we don't want to handle. Do we ever need to use relative paths like `../something`? Probably not.

We can probably simplify the logic to:
* if it starts with `/`, then drop the slash and take the rest as the path to the file from the repository root
* otherwise, add the base path of the component

axieum: I do think relative pathing can be useful, and in its current state, this works really well - as shown by the test cases I wrote. It supports jumping to the repo root, as well as using `../`, and as a side-effect of using this `path` module, it much more secure as it normalizes paths like `foo/../baz` to `baz`, meaning `foo/../../../../../etc/hosts` won't end up at `/etc/hosts`

axieum: We could maybe disallow `../` in paths (for security), and instead require any relative paths to be absolute from the project root, e.g. for component `pkgs/b` instead of `../b.txt` you'd have to write `/pkgs/b.txt`.

chingor13: Yes, let's disallow extra relativity. We should make sure that `/` actually is relative from the git root and that paths without the leading `/` are relative to the component's root.

axieum: @chingor13 removed the use of the `path` module in dc1f3adc4291f5721ffc3c9d022b320cf6c89124 - it now throws an `Error` if the resulting file path includes relativity, i.e. `../`

src/strategies/base.ts
```diff
@@ -450,16 +448,27 @@ export abstract class BaseStrategy implements Strategy {
...
+      file = `${this.path}/${file}`;
     }
-    file = file.replace(/^[/\\]/, '');
-    if (this.path === undefined) {
-      return file;
-    } else {
-      const path = this.path.replace(/[/\\]$/, '');
-      return `${path}/${file}`;
+    // Ensure the file path does not escape the workspace
+    if (/(\.{2}|^~|^\/*)+\//.test(file)) {
```
axieum, line 468: Any of the below sequences in a file path are now rejected -

image

test/strategies/base.ts
```diff
@@ -69,6 +69,74 @@ describe('Strategy', () => {
...
+        github,
+        component: 'google-cloud-automl',
+        extraFiles: [
+          '0',
+          'foo/1.~csv',
+          'foo/2.bak',
+          'foo/baz/bar/',
+          '/3.java',
+          '~/4.md',
+          '~/./5',
```
chingor13, line 84: I don't think there's much of a point of allowing `~` if we're just stripping it away. Let's remove the `~` handling and tests.

test/strategies/base.ts
```diff
@@ -69,6 +69,74 @@ describe('Strategy', () => {
...
+        undefined
+      );
+      expect(pullRequest).to.exist;
+      expect(pullRequest?.updates).to.be.an('array');
+      expect(pullRequest?.updates.map(update => update.path))
+        .to.include.members([
+          '0',
+          '3.java',
+          '4.md',
+          './5',
```
chingor13, line 98: AFAIK, this will end up making an invalid request to the GitHub API (fetching a file at path `./5`). Someone configuring this path should just use `5`.

chingor13 (changes requested):
Apologies for all the little nit-picky comments/requests for changes. The smaller we can keep the feature set, the easier this will be to debug/maintain.

test/strategies/base.ts
axieum: Agreed, dropped support in f9968f8025443b3aa062794e6a37b21d4e2e56b9

test/strategies/base.ts
axieum: It will now throw an error in f9968f8025443b3aa062794e6a37b21d4e2e56b9

src/strategies/base.ts
```diff
@@ -450,16 +448,27 @@ export abstract class BaseStrategy implements Strategy {
...
+      file = `${this.path.replace(/\/+$/, '')}/${file}`;
     }
-    file = file.replace(/^[/\\]/, '');
-    if (this.path === undefined) {
-      return file;
-    } else {
-      const path = this.path.replace(/[/\\]$/, '');
-      return `${path}/${file}`;
+    // Ensure the file path does not escape the workspace
+    if (/((^|\/)\.{1,2}|^~|^\/*)+\//.test(file)) {
```
axieum, line 468: image

chingor13 (approved):

PR #1230, opened 2022-01-18
feat: guess python package name by replacing dash to underscore

### LINKED ISSUE #1229: python release type not able to find correct python packages when using poetry

When using poetry and package name contains dash, python release type not able to find correct version files 

Have a better guess by replacing dash with an underscore.

For example, I have a pyproject.toml

```toml
[tool]

[tool.poetry]
name = "my-service-client"
version = "0.1.0"
description = ""
```

Release please warn me
```bash
file my-service-client/__init__.py did not exist
```

while is file path is actually `my_service_client/__init__.py`

Package name and extra files should be able to bump the file correctly, however guessing the package name better is always something nice to have.

goatwu1993:

Fixes #1229 🦕


src/strategies/python.ts
```diff
@@ -117,6 +117,22 @@ export class Python extends BaseStrategy {
           version,
         }),
       });
+      // In case the project name is splitted by -, replace - with _
+      const projectDirectoryName = projectName.replace('-', '_');
```
chingor13, line 121: I wonder if something weird will happen if `projectDirectoryName` and `projectName` are the same. E.g. will the later API code be weird if there are 2 updates to the same file.

Another approach we could do is to use a `Set`, add both `projectName` and `projectDirectoryName`, then iterate over the members and add the 2 paths as a way to de-dup.

chingor13 (changes requested):
Thanks for this.

Can we add a test for this?

bcoe:
@goatwu1993 just a friendly nudge, looks like @chingor13 has a couple outstanding questions.

goatwu1993:
I had updated with the set/iteration.

Test is lacking still. I checked the test and seems difficult to use sinom to stub a directory with name with dash replaced with underscore😖.

Any hint?

PR #1260, opened 2022-01-27
fix: cargo workspace cargo lock handling

### LINKED ISSUE #1250: cargo-workspace plugin does not handle Cargo.lock files correctly
#### Environment details

  - OS: Ubuntu 20.04
  - Node.js version: 16.13.2
  - npm version: 8.1.2
  - `release-please` version: 13.4.2

#### Steps to reproduce

  1. Create a rust workspace project (https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) containing a single package `crates/package`.
  2. Run release-pr and observe:
    1. release-please tries to add a patch for `crates/package/Cargo.lock`.
    2. release-please doesn't try to add a patch for `Cargo.lock`

When using a cargo workspace the individual crates within that workspace don't have their own Cargo.lock file. Rather, they share a single Cargo.lock file at the root of the workspace. Quoting from https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html:

> Notice that the workspace has only one Cargo.lock file at the top level of the workspace rather than having a Cargo.lock in each crate’s directory.

I'm happy to have a go at making this work correctly, I just need some pointers to make sure I head in the right direction.

gsrushton:

Fixes #1250  🦕

### DIFF
--- a/src/plugins/cargo-workspace.ts
+++ b/src/plugins/cargo-workspace.ts
@@ -282,6 +282,14 @@ export class CargoWorkspace extends WorkspacePlugin<CrateInfo> {
     };
   }
 
+  protected postProcessCandidates(
+    candidates: CandidateReleasePullRequest[],
+    _updatedVersions: VersionsMap
+  ): CandidateReleasePullRequest[] {
+    // NOP for now
+    return candidates;
+  }
+
   protected async buildGraph(
     allPackages: CrateInfo[]
   ): Promise<DependencyGraph<CrateInfo>> {
--- a/src/plugins/cargo-workspace.ts
+++ b/src/plugins/cargo-workspace.ts
@@ -201,10 +201,7 @@ export class CargoWorkspace extends WorkspacePlugin<CrateInfo> {
         } else if (
           update.path === addPath(existingCandidate.path, 'Cargo.lock')
         ) {
-          update.updater = new CargoLock({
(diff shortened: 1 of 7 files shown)


src/plugins/cargo-workspace.ts
```diff
@@ -285,6 +282,24 @@ export class CargoWorkspace extends WorkspacePlugin<CrateInfo> {
...
+    const root_candidate = candidates.find(c => c.path === ROOT_PROJECT_PATH);
+    if (!root_candidate) {
+      throw Error('Unable to find root candidate pull request');
+    }
+
+    root_candidate.pullRequest.updates.push({
+      path: 'Cargo.lock',
+      createIfMissing: false,
+      updater: new CargoLock(updatedVersions),
+    });
```
chingor13, line 298: nit: let's keep our variable names lowerCamelCased. `root_candidate` -> `rootCandidate`

chingor13 (changes requested):
Thanks! This looks good to me with the small nit of renaming a local variable.

src/plugins/cargo-workspace.ts
gsrushton: Done, obviously I was still in rust mode when I wrote this :)

chingor13 (approved):

PR #1267, opened 2022-02-01
fix: check recent commits and latest releases for latest version

### LINKED ISSUE #1266: Sometimes fails to find latest release when using Manifest.fromConfig
If you are using `Manifest.fromConfig` (e.g. not using a manifest file in the repository), we can fail to find the correct latest release.

This could be because the release was manually tagged and doesn't match with what release-please expects.

See https://github.com/googleapis/release-please/issues/1259#issuecomment-1022603167

chingor13:
It's possible that a manual release was created/tagged since the last release-please PR was merged/released. The logic currently prefers any merged release PR over manually created releases. Now, we stop iterating through commits after we find the first merged release PR, but still consider recent releases (picking the highest number).

Pro: This method is more robust and resilient
Con: We now always iterate over releases with GraphQL queries.

Fixes #1266

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -969,6 +969,7 @@ async function latestReleaseVersion(
   // is in the current branch
   const commitShas = new Set<string>();
 
+  const candidateReleaseVersions: Version[] = [];
   // only look at the last 250 or so commits to find the latest tag - we
   // don't want to scan the entire repository history if this repo has never
   // been released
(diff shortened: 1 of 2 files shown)


src/manifest.ts
```diff
@@ -1005,13 +1006,18 @@ async function latestReleaseVersion(
       continue;
     }
 
-    return version;
+    if (version) {
+      logger.debug(
+        `Found latest release pull request: ${mergedPullRequest.number} version: ${version}`
+      );
+      candidateReleaseVersions.push(version);
```
bcoe, line 1013: Will this behave appropriately with multiple branches, e.g., perhaps the tag is against a `preview` branch, but the release PR is against `main`?

bcoe (commented):
Left a question.

src/manifest.ts
chingor13: Yes, the first loop is iterating through commits on the target branch, so it should only find release PRs on that branch. Then we look at releases and only include that release if we see it's commit in the list of commits we've scanned (if we see one, then it's likely a later release than the one we found via PR). By cross-referencing against commit SHAs we've seen, we ensure we're only getting releases on the branch we are looking at.

bcoe (approved):

PR #1269, opened 2022-02-01
fix: krm blueprints should update yaml files

### LINKED ISSUE #1268: krm-blueprint releases are not update yamls
Example: https://github.com/GoogleCloudPlatform/blueprints/pull/130

chingor13:
Prior to v13, we were passing around version strings which included `v`. This regex replacement was assuming that the version included `v` which it did not in v13. Also, the tests did not cover the case when a `previousVersion` was provided.

Fixes #1268


chingor13:
cc @bharathkkb

bharathkkb (approved):
Thanks for fixing this!

bcoe (approved):

PR #1270, opened 2022-02-01
fix: read packageName from config

### LINKED ISSUE #1264: samples/package.json no longer updated during release
It appears that `samples/package.json` is no longer updated for Node.js release, see:

https://github.com/googleapis/google-api-nodejs-client/pull/2880
https://github.com/googleapis/nodejs-firestore/pull/1665

This results in the samples running against the wrong version of the library after the release is merged (noticed due to failing PRs on google-api-client).

bcoe:
Read packageName from config if not provided.

Fixes #1264

### DIFF
--- a/src/strategies/base.ts
+++ b/src/strategies/base.ts
@@ -135,7 +135,9 @@ export abstract class BaseStrategy implements Strategy {
   }
 
   async getDefaultComponent(): Promise<string | undefined> {
-    return this.normalizeComponent(await this.getDefaultPackageName());
+    return this.normalizeComponent(
+      this.packageName ?? (await this.getDefaultPackageName())
+    );
   }
 
   async getDefaultPackageName(): Promise<string | undefined> {
@@ -227,6 +229,8 @@ export abstract class BaseStrategy implements Strategy {
       await this.buildVersionsMap(conventionalCommits),
       conventionalCommits
     );
+    this.packageName = this.packageName ?? (await this.getDefaultPackageName());
+    logger.debug('packageName', this.packageName);
     const component = await this.getComponent();
     logger.debug('component:', component);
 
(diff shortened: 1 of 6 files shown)


SurferJeffAtGoogle (approved):

src/strategies/base.ts
```diff
@@ -227,6 +229,8 @@ export abstract class BaseStrategy implements Strategy {
       await this.buildVersionsMap(conventionalCommits),
       conventionalCommits
     );
+    this.packageName = this.packageName ?? (await this.getDefaultPackageName());
+    logger.debug('packageName', this.packageName);
```
chingor13, line 233: This is potentially an unnecessary call. We should probably make this a memoizing getter and only call it when necessary.

test/strategies/node.ts
```diff
@@ -132,6 +168,14 @@ describe('Node', () => {
       assertHasUpdate(updates, 'package-lock.json', PackageLockJson);
       assertHasUpdate(updates, 'npm-shrinkwrap.json', PackageLockJson);
       assertHasUpdate(updates, 'samples/package.json', SamplesPackageJson);
+      // Ensure that the packageName is used vs., the component
+      // when updating samples/package.json:
+      for (const update of updates) {
+        if (update.path === 'samples/package.json') {
```
chingor13, line 174: The `assertHasUpdate()` method (line 170) returns the `Update` if found. This would eliminate the need to fin

test/strategies/node.ts
```diff
@@ -113,13 +115,47 @@ describe('Node', () => {
...
+      expect(pullRequest!.version?.toString()).to.eql(expectedVersion);
+    });
   });
   describe('buildUpdates', () => {
     it('builds common files', async () => {
       const strategy = new Node({
         targetBranch: 'main',
         github,
         component: 'google-cloud-automl',
+        packageName: 'google-cloud-automl-pkg',
```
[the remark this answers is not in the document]
bcoe: Yes but because `component` was used in place of `packageName` this test case would always have passed, the issue is if the packageName and component diverge.

PR #1273, opened 2022-02-01
refactor: make packageName private

bcoe:
make pkgName property private.

### DIFF
--- a/src/strategies/base.ts
+++ b/src/strategies/base.ts
@@ -76,7 +76,7 @@ export abstract class BaseStrategy implements Strategy {
   readonly path: string;
   protected github: GitHub;
   protected component?: string;
-  protected packageName?: string;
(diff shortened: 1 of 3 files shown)


src/strategies/base.ts
```diff
@@ -141,7 +141,7 @@ export abstract class BaseStrategy implements Strategy {
   }
 
   async getDefaultPackageName(): Promise<string | undefined> {
-    return '';
+    return this.packageName ?? '';
```
chingor13, line 144: I think I had intended for `getDefaultPackageName()` to only contain the logic of fetching from another source (like `package.json`).

WDYT about making a `getPackageName()` accessor like we have for `getComponent()` that does the `this.packageName ||= await getDefaultPackageName()` logic.

chingor13 (approved):

PR #1285, opened 2022-02-04
fix: provide pull request to commit parser allow overrides

### LINKED ISSUE #1284: Commit message override is not working
https://github.com/googleapis/release-please#how-can-i-fix-release-notes appears to not be working

chingor13:
Fixes #1284

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -462,6 +462,7 @@ export class Manifest {
         sha: commit.sha,
         message: commit.message,
         files: commit.files,
+        pullRequest: commit.pullRequest,
       });
     }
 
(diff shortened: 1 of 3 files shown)


src/manifest.ts
```diff
@@ -462,6 +462,7 @@ export class Manifest {
         sha: commit.sha,
         message: commit.message,
         files: commit.files,
+        pullRequest: commit.pullRequest,
```
bcoe, line 465: Is this just the `pullRequest` #, curious why this was an optional parameter?

bcoe:
I'm not sure I'm fully understanding the cause of this bug.

src/manifest.ts
chingor13: This is the full pull request data. It's optional because not all commits on a branch have an associated pull request.

chingor13:
> I'm not sure I'm fully understanding the cause of this bug.

This happened because we weren't passing the pull request data to the strategies when we collected commits. The feature was unit tested assuming that the pull request data was set on the commits.

bcoe (approved):

PR #1287, opened 2022-02-04
fix(go-yoshi): allows using go-yoshi with manifest

chingor13:
fix: allow configuring includeComponentInTag and tagSeparator from manifest config
fix(go-yoshi): should not always skip modules
fix: correctly fetch full list of files
fix: Manifest should be able to find tagged versions without a release

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -141,6 +141,7 @@ export async function buildStrategy(
     changelogNotes,
     pullRequestTitlePattern: options.pullRequestTitlePattern,
     extraFiles: options.extraFiles,
+    tagSeparator: options.tagSeparator,
   };
   switch (options.releaseType) {
     case 'ruby': {
--- a/src/github.ts
+++ b/src/github.ts
@@ -95,6 +95,9 @@ interface GraphQLPullRequest {
     nodes: {
       path: string;
     }[];
+    pageInfo: {
+      hasNextPage: boolean;
+    };
   };
 }
 
@@ -387,7 +390,7 @@ export class GitHub {
       repo: this.repository.repo,
       num: 25,
       targetBranch,
-      maxFilesChanged: 64,
+      maxFilesChanged: 100, // max is 100
     });
 
     // if the branch does exist, return null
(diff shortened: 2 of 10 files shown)


codyoss:
LGTM me overall.

There is one tricky bit that I am not sure how to sort out still here though. The root module in this release should be basically just commits that are not associated with another release. Right now there is a hard coded list of what previously used to be all of the sub-moudles. But this list is closer to 70+ now and is dynamic with new modules that will be added for cloud APIs. I am not sure of a good way to get this dynamic list w/o using the go command or reading the file system for go.mod files. Maybe the best course is just to not release the root module /w release-please. Almost nothing is in it these days anyways besides some test helpers and adapters for cross-module compatibility.

Thoughts?

src/github.ts
```diff
@@ -387,7 +390,7 @@ export class GitHub {
       repo: this.repository.repo,
       num: 25,
       targetBranch,
-      maxFilesChanged: 64,
+      maxFilesChanged: 100, // max is 100
```
bcoe, line 393: This bug would have been an issue if more than 64 commits were merged since the last release?

src/manifest.ts
```diff
@@ -404,6 +405,43 @@ export class Manifest {
     }
 
     const needsBootstrap = releasesFound < expectedReleases;
+    if (releasesFound < expectedReleases) {
+      logger.warn(
+        `Expected ${expectedReleases} releases, only found ${releasesFound}`
```
bcoe, line 410: This adds quite a bit of logic, I think it would be worth pulling it into a helper method, just to encapsulate it.

What state does the repo need to be in for this additional lookup to be necessary, tags without releases?

test/fixtures/manifest/config/tag-separator.json
```diff
@@ -0,0 +1,13 @@
+{
+  "release-type": "simple",
+  "tag-separator": "/",
+  "packages": {
+    ".": {
+      "component": "root",
+      "tag-separator": "-"
```
bcoe, line 7: This answers my early question about seperators.

bcoe (changes requested):
one suggested refactor, a couple questions.

src/github.ts
chingor13: Previously, if there were >64 but less than 100 files in the PR, then we could possibly not attribute the PR to a specific path (the GraphQL for PR files is not paginatable). This minor improvement allows us to not have to make the extra backfill API call if there are between 64 and 100 files in the PR.

src/factory.ts
```diff
@@ -141,6 +141,7 @@ export async function buildStrategy(
     changelogNotes,
     pullRequestTitlePattern: options.pullRequestTitlePattern,
     extraFiles: options.extraFiles,
+    tagSeparator: options.tagSeparator,
```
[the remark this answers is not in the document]
chingor13: We default to `-`. Ruby uses `/` for the separator and go uses `/` for the yoshi monorepo

chingor13:
> The root module in this release should be basically just commits that are not associated with another release

From looking at https://github.com/googleapis/google-cloud-go/pull/5206, I had assumed it included all changes not in the subset of ignored modules (e.g. spanner, storage, pubsub)

bcoe (approved):

PR #1291, opened 2022-02-09
fix(go-yoshi): dynamically load list of ignored submodules for google-cloud-go

chingor13:
The root module was always intended to only include non-submodule components. We now dynamically load the list by finding the `go.mod` files in the repository.

This fixes the root component release PRs to look like: https://github.com/googleapis/google-cloud-go/pull/5498


chingor13:
cc @codyoss

codyoss (approved):
LGTM! Thank you!

bcoe (approved):

PR #1293, opened 2022-02-10
feat: add dotnet-yoshi strategy

chingor13:
This release strategy is meant for internal usage.

Updates apis.json and history.md files.

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -45,6 +45,7 @@ import {GitHubChangelogNotes} from './changelog-notes/github';
 import {DefaultChangelogNotes} from './changelog-notes/default';
 import {BaseStrategyOptions} from './strategies/base';
 import {LinkedVersions} from './plugins/linked-versions';
+import {Dotnet} from './strategies/dotnet';
 
 // Factory shared by GitHub Action and CLI for creating Release PRs
 // and GitHub Releases:
@@ -52,6 +53,7 @@ import {LinkedVersions} from './plugins/linked-versions';
 // object below.
 const allReleaseTypes = [
   'dart',
+  'dotnet',
   'elixir',
   'go',
   'go-yoshi',
@@ -76,6 +78,7 @@ export type ReleaseType = typeof allReleaseTypes[number];
 type ReleaseBuilder = (options: BaseStrategyOptions) => Strategy;
 type Releasers = Record<string, ReleaseBuilder>;
 const releasers: Releasers = {
+  dotnet: options => new Dotnet(options),
   go: options => new Go(options),
   'go-yoshi': options => new GoYoshi(options),
   'java-yoshi': options => new JavaYoshi(options),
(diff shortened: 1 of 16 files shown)


Tuneabq (approved):

src/factory.ts
```diff
@@ -45,13 +45,15 @@ import {GitHubChangelogNotes} from './changelog-notes/github';
...
 import {LinkedVersions} from './plugins/linked-versions';
+import {DotnetYoshi} from './strategies/dotnet-yoshi';
 
 // Factory shared by GitHub Action and CLI for creating Release PRs
 // and GitHub Releases:
 // add any new releasers you create to this type as well as the `releasers`
 // object below.
 const allReleaseTypes = [
   'dart',
+  'dotnet-yoshi',
```
bcoe, line 56: A good future refactor (_you've brought it up too_) would be making it so you only need to define a new strategy in one place.

src/strategies/dotnet-yoshi.ts
```diff
@@ -0,0 +1,142 @@
...
+    latestRelease?: Release
+  ): Promise<string> {
+    const notes = await super.buildReleaseNotes(
+      conventionalCommits,
+      newVersion,
+      newVersionTag,
+      latestRelease
+    );
+    return notes.replace(
+      RELEASE_NOTES_HEADER_PATTERN,
```
bcoe, line 78: We should double check with Martin V. that our release notes follow the correct pattern for the release notes on c.g.c design, once .NET is starting to move over.

src/strategies/dotnet-yoshi.ts
```diff
@@ -0,0 +1,142 @@
...
+    options: BuildUpdatesOptions
+  ): Promise<Update[]> {
+    const updates: Update[] = [];
+    const version = options.newVersion;
+    const component = await this.getComponent();
+
+    const api = await this.getApi();
+    if (api?.noVersionHistory) {
+      logger.info(
+        `Skipping changelog for ${component} via noVersionHistory configuration`
```
bcoe, line 114: Is `apis.json` similar to `release-please-manifest.json`?

Wondering out loud if an alternate implementation would have been treating .NET as being configured as a manifest based release, and supporting this alternate metadata file rather than `releasae-please-manifest.json`.

test/updaters/fixtures/dotnet/SecurityCenter.V1.csproj
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
```
bcoe, line 2: Do we still do anything with `.csproj` files, in the strategy I only saw apis.json being updated.

bcoe (approved):
Left a couple small nits and a few questions.

src/strategies/dotnet-yoshi.ts
chingor13: It will still be a manifest based release, but we don't need to have extra configs for this flag, unless it's a feature (skipping the changelog) we want to support for more strategies.

test/updaters/fixtures/dotnet/SecurityCenter.V1.csproj
chingor13: Do, but I somehow added the updater a long time ago without a test

PR #1296, opened 2022-02-14
fix: standalone releases should only be released by its matching component

chingor13:
If you have a component-less configuration, it should not try to handle a release for a standalone release PR that does have a component (it will incorrectly attempt to tag the release)

This check ensure that we only have the configured component releaser attempting to tag the release.


bcoe (approved):
Would it make sense to create a `component.js` class, similar to `version.js` that encapsulates some of this logic?

PR #1308, opened 2022-02-21
docs: add section on the format of a path

### LINKED ISSUE #1289: Document the "path" option as a segment of file's path
The `path` option is crucial for integrating Release Please in monorepos. However, there are no examples featured on how to use it. 

> For the context, I'm using `release-please-action` and pass the `path` option in my workflow YAML.

Naturally, I've followed the option's description from the [README](https://github.com/google-github-actions/release-please-action#configuration):

> create a release from a path other than the repository's root

And set it accordingly:

```yml
- uses: google-github-actions/release-please-action@v3
  id: release
  with:
    path: ./my-package
    release-type: node
    package-name: my-package
```

But Release Please couldn't find the existing commits at the given path. After looking into the source code, I can see how the `path` option is being treated:

https://github.com/googleapis/release-please/blob/a78993fea0058e1befaf8337034f2dc8e5e4eb33/src/util/commit-split.ts#L105-L106

This implies that that the value of the `path` option is, in fact, a _partial path_ to the nested repository:

```diff
- path: ./my-package
+ path: /my-package
```

The dot (`.`) in the path (which is the correct relative path format) will never match, as it's never present in the actual path upon comparison in `commit-split`. 

## Expected behavior

This must be documented in this repository. Although the issue originated from me using the action, this is really this library's behavior, and so it must be documented close to the source.

chingor13:
Fixes #1289

### DIFF
--- a/docs/customizing.md
+++ b/docs/customizing.md
@@ -44,6 +44,16 @@ version given a list of parsed commits.
 To add a new versioning strategy, create a new class that implements the
 [`VersioningStrategy` interface](/src/versioning-strategy.ts).
 
+## Subdirectories (paths) in a repository
+
+Release Please can operate on a subdirectory of a repository. If you configurate a `path`,
+Release Please will only consider commits that touch files on that path. The format of the
+`path` configuration option should be a simple path segment (`.`, `..`, etc are not allowed)
+relative to the repository root.
+
+To configure multiple components on different paths, configure a
+[manifest release](/docs/manifest-releaser).
+
 ## Changelog Types
 
 A changelog type's job is to build the CHANGELOG notes given a list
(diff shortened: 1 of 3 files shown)


docs/customizing.md
```diff
@@ -44,6 +44,16 @@ version given a list of parsed commits.
 To add a new versioning strategy, create a new class that implements the
 [`VersioningStrategy` interface](/src/versioning-strategy.ts).
 
+## Subdirectories (paths) in a repository
+
+Release Please can operate on a subdirectory of a repository. If you configurate a `path`,
```
axieum, line 49:
```suggestion
Release Please can operate on a subdirectory of a repository. If you configure a `path`,
```

bcoe (approved):

PR #1311, opened 2022-02-22
fix: address false-positive matches for autorelease branch naming

### LINKED ISSUE #1310: "Error parsing branch name" for branches like "v3.3.x"

Issue with release-please, used via release-please-action to automate releases of google/shaka-player.

  - Search the issues already opened: https://github.com/googleapis/release-please/issues
  - Search the issues on our "catch-all" repository: https://github.com/googleapis/google-cloud-node
  - Search or ask on StackOverflow (engineers monitor these tags): http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js

This is very similar to issue #1226 and related to the fix in PR #1227.  I think the fix could be improved. 
 I have an analysis of how it failed for my repo, and what could be done to make release-please a bit more resilient in this case.

No, but I'm a Googler working on an official Google project.

#### Environment details

  - OS: Ubuntu 20.04 via GitHub Actions (ubuntu-latest)
  - Node.js version: 16.14.0
  - npm version: 8.3.1
  - `release-please` version: 13.4.10 via release-please-action 3.1.2

#### Steps to reproduce

  1. Set up release-please or release-please-action on a repo with a branch named something like v3.3.x (as we have in google/shaka-player)
  2. Merge a release PR

#### Analysis

Pattern matching in src/util/branch-name.ts goes like this:

```ts
function getAllResourceNames(): BranchNameType[] {
  return [
    AutoreleaseBranchName,
    ComponentBranchName,
    DefaultBranchName,
    V12ComponentBranchName,
    V12DefaultBranchName,
  ];
}

// ...

      const branchNameClass = getAllResourceNames().find(clazz => {
        return clazz.matches(branchName);
      });
```

Since AutoreleaseBranchName appears before DefaultBranchName, that matches something like `release-please--branches--v3.3.x` based on:

```ts
const AUTORELEASE_PATTERN =
  /^release-?(?<component>[\w-.]*)?-v(?<version>[0-9].*)$/;
```

It thinks `please--branches-` is the component.  However, when constructed, AutoreleaseBranchName throws from the constructor because `Version.parse` can't parse `v3.3.x` (the branch name):

```ts
  constructor(branchName: string) {
    super(branchName);
    const match = branchName.match(AUTORELEASE_PATTERN);
    if (match?.groups) {
      this.component = match.groups['component'];
      this.version = Version.parse(match.groups['version']);
    }
  }
```

See https://github.com/google/shaka-player/issues/3968

#### Proposal

First, I would suggest we change the matching such that the component can't be `please--branches-`.  I think this alone would be sufficient for me and my case.  I'll go ahead and send a PR for that.

However, I would suggest that we also go a bit further.  If an exception is thrown in the constructor of one of these classes, I would suggest that we fall back to the next match in the list instead of returning undefined.  This would mean replacing `getAllResourceNames().find(...)` with a loop over `match()` and `new`, with try/catch inside the loop instead of outside.  If the maintainers are concerned about this step, I would be happy to drop it.

joeyparrish:
Autorelease branch naming can mistakenly match for any branch
beginning with a v.  A PR branch like
"release-please--branches--v3.3.x" will be treated as an autorelease
branch for a component called "please--branches-".

If the branch name is a valid semver, the wrong class will be chosen
to parse the branch name, and later steps will fail.  If it's not a
valid semver, then branch parsing will completely fail.

This excludes these false-positive matches by excluding branches which
start with "release-please--branches".

Fixes #1310 🦕

### DIFF
--- a/src/util/branch-name.ts
+++ b/src/util/branch-name.ts
@@ -101,8 +101,12 @@ export class BranchName {
  */
 const AUTORELEASE_PATTERN =
   /^release-?(?<component>[\w-.]*)?-v(?<version>[0-9].*)$/;
+const RELEASE_PLEASE_BRANCH_PREFIX = 'release-please--branches';
 class AutoreleaseBranchName extends BranchName {
   static matches(branchName: string): boolean {
+    if (branchName.startsWith(RELEASE_PLEASE_BRANCH_PREFIX)) {
+      return false;
+    }
     return !!branchName.match(AUTORELEASE_PATTERN);
   }
   constructor(branchName: string) {
--- a/test/util/branch-name.ts
+++ b/test/util/branch-name.ts
@@ -76,6 +76,26 @@ describe('BranchName', () => {
       expect(branchName?.toString()).to.eql(name);
     });
 
+    it('parses a target branch that starts with a v', () => {
+      const name = 'release-please--branches--v3.3.x';
+      const branchName = BranchName.parse(name);
+      expect(branchName).to.not.be.undefined;
+      expect(branchName?.getTargetBranch()).to.eql('v3.3.x');
+      expect(branchName?.getComponent()).to.be.undefined;
+      expect(branchName?.getVersion()).to.be.undefined;
+      expect(branchName?.toString()).to.eql(name);
+    });
+
+    it('parses a target branch named with a valid semver', () => {
+      const name = 'release-please--branches--v3.3.9';
+      const branchName = BranchName.parse(name);
+      expect(branchName).to.not.be.undefined;
+      expect(branchName?.getTargetBranch()).to.eql('v3.3.9');
+      expect(branchName?.getComponent()).to.be.undefined;
+      expect(branchName?.getVersion()).to.be.undefined;
+      expect(branchName?.toString()).to.eql(name);
+    });
+
     it('parses a target branch and component', () => {
       const name = 'release-please--branches--main--components--storage';
       const branchName = BranchName.parse(name);


src/util/branch-name.ts
```diff
@@ -103,7 +103,22 @@ const AUTORELEASE_PATTERN =
...
+    if (!match) {
+      return false;
+    }
+
+    // The regex can match release branches that should belong to
+    // DefaultBranchName (such as release-please--branches--v3.3.x). Make sure
+    // the component, if present, isn't something suspicious. See also
+    // https://github.com/googleapis/release-please/issues/1310
+    if (match?.groups) {
+      const component = match.groups['component'];
```
bcoe, line 116:
```suggestion
      const component = match.groups.component
```

stylistic-nit.

src/util/branch-name.ts
```diff
@@ -103,7 +103,22 @@ const AUTORELEASE_PATTERN =
   /^release-?(?<component>[\w-.]*)?-v(?<version>[0-9].*)$/;
 class AutoreleaseBranchName extends BranchName {
   static matches(branchName: string): boolean {
-    return !!branchName.match(AUTORELEASE_PATTERN);
+    const match = branchName.match(AUTORELEASE_PATTERN);
```
bcoe, line 106: If I'm understanding the problem properly, could this safely be simplified to:

```
const BRANCH_PREFIX= 'release-please--branches';
if (branchName.startsWith(BRANCH_PREFIX)) return false;
```

bcoe (changes requested):
Left a nit and a question.

src/util/branch-name.ts
joeyparrish: Thanks, I'm new to TS.  Still doing JS + Closure in my own projects.

src/util/branch-name.ts
joeyparrish: Ah, yes, that works.  I'll do that.

chingor13 (approved):

bcoe (approved):

PR #1315, opened 2022-02-24
fix: handle failures during multiple release creation

### LINKED ISSUE #1313: Manifest release can fail part way through multiple releases
Example: https://github.com/googleapis/google-cloud-go/pull/5682

Retrying will throw a `DuplicateReleaseError` which is correct, but it should still attempt each individual tag before throwing

chingor13:
Catch duplicate release errors when handling a manifest release. If all releases fail because of duplicate release errors, then it should throw the duplicate release error.

Fixes #1313


bcoe (approved):

codyoss:
@chingor13 Can we merge this 😄

PR #1321, opened 2022-03-01
fix: tagging many GitHub releases should not require a file touched check

### LINKED ISSUE #1320: Manifest release not tagging all releases
https://github.com/googleapis/google-cloud-go/pull/5682 only tagged 32 releases when it should be 106

chingor13:
Fixes #1320

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -761,30 +761,11 @@ export class Manifest {
     const generator = await this.findMergedReleasePullRequests();
     const releases: CandidateRelease[] = [];
     for await (const pullRequest of generator) {
-      logger.info('Looking at files touched by path');
-      const cs = new CommitSplit({
-        includeEmpty: true,
-        packagePaths: Object.keys(this.repositoryConfig),
-      });
-      const commits = [
-        {
-          sha: pullRequest.sha!,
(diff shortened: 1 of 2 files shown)


bcoe (approved):

src/manifest.ts
```diff
@@ -761,30 +761,11 @@ export class Manifest {
...
-          files: pullRequest.files,
-        },
-      ];
-      const commitsPerPath = cs.split(commits);
       for (const path in this.repositoryConfig) {
         const config = this.repositoryConfig[path];
         logger.info(`Building release for path: ${path}`);
         logger.debug(`type: ${config.releaseType}`);
         logger.debug(`targetBranch: ${this.targetBranch}`);
-        const pathCommits =
```
bcoe, line 782: I trust your debugging, but am a little confused by this problem. Shouldn't there have been some commits for each path that requires a release? Why did this result in 32 releases for Go rather than 100+?

chingor13: This missed because you need to paginate through the files list which you can't do from initial GraphQL query. 

The other option was to use the list files from a commit (with pagination) to backfill the full list, but if it doesn't actually matter if there were touched files, we should just remove the check.

PR #1326, opened 2022-03-06
feat: keep non-conforming commits as `undefined` type

mdvorak:
Conventional commits are great, but people make mistakes. Commits without conventional format got thrown away completely. This patch allows them to be displayed in Release Notes, and passed to a custom Release Notes generator as well. That way, information from those commits won't be lost during release.

This change should not be breaking,  despite its nature.

**Use-cases**
- Enable `undefined` type and simply append non-conforming commits to Release Notes
- Support "dirty" repo not following conventional commits spec completely (this can happen easily when for example forking repo and updating from upstream)
- Custom Release Notes generator, which might need all commits to work with.

Change itself is a bit dirty, but was easier way to add the support, and keep all the complex logic of commit processing in place.

Since this change might be a bit controversial, I'm open to a discussion :-)

### DIFF
--- a/src/commit.ts
+++ b/src/commit.ts
@@ -326,9 +326,24 @@ function hasExtendedContext(line: string) {
   return false;
 }
 
-function parseCommits(message: string): parser.ConventionalChangelogCommit[] {
+function parseOrUndefined(commit: Commit, message: string): parser.Message {
+  try {
+    return parser.parser(message);
+  } catch (_err) {
+    logger.debug(
+      `commit could not be parsed: ${commit.sha} ${message.split('\n')[0]}`
+    );
+
+    return parser.parser('undefined: ' + message);
+  }
+}
+
+function parseCommits(
+  commit: Commit,
+  message: string
+): parser.ConventionalChangelogCommit[] {
   return conventionalCommitsFilter(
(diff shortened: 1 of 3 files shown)


chingor13 (changes requested):
Can you open a separate issue with the use-case described? We can have the discussion there rather than in this pull request (keep the PR comments more around the implementation).

src/commit.ts
```diff
@@ -326,9 +326,24 @@ function hasExtendedContext(line: string) {
   return false;
 }
 
-function parseCommits(message: string): parser.ConventionalChangelogCommit[] {
+function parseOrUndefined(commit: Commit, message: string): parser.Message {
+  try {
+    return parser.parser(message);
+  } catch (_err) {
```
chingor13, line 332: Does this throw? If it does, then the existing code would have a bad time with non-conforming commits.

mdvorak: Yeah, parser throws an exception. I thought pretty long, whether change should be done in the parser itself, but I don't think it really belongs there (or it would have to be hidden by an optional flag).

Despite the hack-ish implementation, it seems to work pretty well here.

mdvorak:
I'm gonna close this for now, as I don't have ATM time to work on it. Hopefully I'll get around to it again.

PR #1332, opened 2022-03-09
feat: introduce generic json updater

### LINKED ISSUE #1171: Add ability to update arbitrary JSON files

chingor13:
For most release strategies, you can provide additional files to update
using the [GenericJson](/src/updaters/generic-json.ts) updater. You can
specify a configuration object in the `extra-files` option in the manifest
configuration.

```json
{
  "extra-files": [
    {
      "type": "json",
      "path": "path/to/file.json",
      "jsonpath": "$.json.path.to.field"
    }
  ]
}
```

Fixes #1171


chingor13:
An alternative configuration option is to add a new `extra-json-files` field rather than changing the `extraFiles` field to be a union of different options (string or object - perhaps with an XML version later)

bcoe (commented):
This is really slick, bother you to add some docs in the README?

mdvorak:
XML variant would be great, when combined with Java strategy, it would allow updating pom.xml files without inline comments.

bcoe (approved):

PR #1333, opened 2022-03-10
feat: added java strategy

### LINKED ISSUE #1334: Support for simple Java projects
There is a support via `java-yoshi` strategy, but that is tailored to Google needs, and it is quite confusing and error-prone for generic Maven/Gradle projects.

I propose creating basic `java` strategy, which does not require `versions.txt` file - this complex functionality is usually not needed, and easily can cause getting releases and actual artifact versions out-of-sync (by human error). `java-yoshi` should retain its current functionality.

Related PR draft: #1333

mdvorak:
There is a support via `java-yoshi` strategy, but that is tailored to Google needs, and it is quite confusing and error-prone for generic Maven/Gradle projects.

Closes #1334 

TBD

- [x] new strategies docs
- [x] PomXml updater tests
- [x] Additional Manifest tests
- [x] Java strategy tests
- [x] Maven strategy tests
- [x] Decide PomXml updater behavior


mdvorak:
@bcoe I've rebased it and cleaned it up, it is still missing tests, but I'd like to discuss it before moving further.

1. It does not work with manifest file. java-yoshi never did. It is related to #1340 - I would document it and leave it at that. It would be however better to make manifest json optional even when using config json. That would cover most use-cases.
2. It uses commit history to look for snapshot releases. So far my tests went well, but it needs to be battle tested in real world.
3. `JavaYoshi` now extends `Java` strategy - it is designed not to change its behavior, but one can never be 100% sure.
4. I've added another label `autorelease: snapshot` to mark snapshot PRs. This is needed, since when Manifest searches for open PRs, it would not found snapshot PR. That results in recreating and force-pushing to that PR on every run. There are other ways how to solve this, but this one seemed best. It however means adding `snapshotLabels` config everywhere, even thou it is used only by java strategy. I'm not happy with it, but I can't find better solution.
5. I'm not happy with `JavaUpdate` class behavior - I've kinda rigged it so it uses component/package name, but it does not really make sense. Generic updater however (and new Json/Xml updaters) does not differentiate between current and released versions. It works now, since released version is needed usually only in docs, and pom.xml etc are usually snapshots. But it won't cover all use-cases. Another way would be adding yet another updater for java, or extending updaters interface to differentiate between snapshots and releases. Neither is nice.

Please take a look. If current behavior is OK, I'll start on tests.

mdvorak:
After thinking about 5. `JavaUpdate`, I suggest renaming current class to `YoshiJavaUpdate` and creating new updater, based on `Generic`, using same markers. Thats because there is discrepancy between the patterns:

- `Generic` uses `x-release-please-version`
- `JavaUpdate` uses `x-version-update:<component>:<type>`

My idea is the new `JavaUpdate` with patterns `x-release-please-version:current`, `x-release-please-version:released` and compatibility with `Generic` pattern `x-release-please-version` equal to current version.

mdvorak:
I've cleaned it up a bit, by using new `JavaReleased` updater, I could remove versionMap hacks from the `Java` strategy.
I've also added `Maven` strategy, which does automatic `pom.xml` version updates.

It is still missing tests and docs, but I think it is getting very close to final version. I'll get on those ASAP.

mdvorak:
I had to add one more change: 
Support multiple updaters on the same file. Reuse previous changed content for second updater.

This issue is more visible with java strategy, but it would break with extraFiles listing already updated file.

chingor13:
> I suggest renaming current class to `YoshiJavaUpdate` and creating new updater, based on `Generic`, using same markers

This sounds reasonable to me.

> Support multiple updaters on the same file. Reuse previous changed content for second updater.

For this, we should probably use the `CompositeUpdater`. Alternatively, when we go to build the change set for suggesting the file change, de-dup on file path and apply both updates sequentially.

mdvorak:
> Thanks for this, it will take some time to wrap my head around all these proposed changes.

Yeah, it got a bit bigger than I expected at first. But I've tried to make systematic changes, not just hacking it to work. Therefore it affected more code.

In the mean time, I'll work on unit tests. There is quite some work left there anyway.

src/updaters/base-xml.ts
```diff
@@ -0,0 +1,45 @@
...
+    const updated = this.updateDocument(document);
+
+    if (updated) {
+      return new dom.XMLSerializer().serializeToString(document);
+    } else {
+      return content;
+    }
+  }
+
+  protected abstract updateDocument(document: Document): boolean;
```
chingor13, line 36: Let's make sure to document this, especially that the `document` is expected to be mutated.

src/updaters/generic.ts
```diff
@@ -51,6 +51,22 @@ type BlockScope = 'major' | 'minor' | 'patch' | 'version';
...
 export class Generic extends DefaultUpdater {
+  private readonly inlineUpdateRegex: RegExp;
+  private readonly blockStartRegex: RegExp;
+  private readonly blockEndRegex: RegExp;
+
+  constructor(
+    options: UpdateOptions,
+    inlineUpdateRegex = INLINE_UPDATE_REGEX,
+    blockStartRegex = BLOCK_START_REGEX,
+    blockEndRegex = BLOCK_END_REGEX
```
chingor13, line 62: Can we roll these options into an extended options class (like `GenericUpdateOptions`)? If for whatever reason, we need to add more optional params, then the caller doesn't have to specify the earlier default values unless they're being overridden.

src/strategy.ts
```diff
@@ -57,4 +58,11 @@ export interface Strategy {
    * @returns {string}
    */
   getComponent(): Promise<string | undefined>;
+
+  /**
+   * Validate whether version is a valid release.
+   * @param version Released version.
+   * @returns true of release is valid, false if it should be skipped.
+   */
+  isValidRelease(version: Version): boolean;
```
chingor13, line 67: How about `isPublishedVersion`?

src/strategies/java.ts
```diff
@@ -0,0 +1,280 @@
...
+    if (!extraFile) {
+      updates.push({
+        path: this.addPath(path),
+        createIfMissing: false,
+        updater: new Generic({
+          version: options.newVersion,
+          versionsMap: options.versionsMap,
+        }),
+      });
+    }
```
chingor13, line 267: Sorry, I'm not following why we're skipping the file if it's an `extraFile`

src/strategies/base.ts
```diff
@@ -69,6 +69,8 @@ export interface BaseStrategyOptions {
   includeComponentInTag?: boolean;
   pullRequestTitlePattern?: string;
   extraFiles?: ExtraFile[];
+  // Java-only
+  snapshotLabels?: string[];
```
chingor13, line 73: Can we move this to a `JavaStrategyOptions` similar to Ruby: https://github.com/googleapis/release-please/blob/d3c5be564e4b1034e35d0b4d87d0337b0370772c/src/strategies/ruby.ts#L26-L28

src/manifest.ts
```diff
@@ -1052,13 +1064,16 @@ async function parseReleasedVersions(
...
+ * @param github GitHub client instance.
+ * @param {string} targetBranch Name of the scanned branch.
+ * @param releaseFilter Validator function for release version. Used to filter-out SNAPSHOT releases for Java strategy.
+ * @param {string} prefix Limit the release to a specific component.
+ * @param pullRequestTitlePattern Configured PR title pattern.
  */
 async function latestReleaseVersion(
   github: GitHub,
   targetBranch: string,
+  releaseFilter: (version: Version) => boolean,
```
chingor13, line 1076: This is technically a breaking change to the interface (`Manifest` is exported from the index and this is a public method). Can we set a default implementation?

chingor13 (changes requested):
Overall, this is looking pretty solid.

I don't quite follow the difference between the `java` strategy and the `maven` strategy. I would think that `java` would be an abstract strategy (in charge of the snapshot logic). `maven` would be a strategy that updates only the `pom.xml` + `CHANGELOG.md` (right now I think it's still finding all the `build.gradle` and `dependencies.properties` files which are a `java-yoshi` thing).

src/updaters/java/pom-xml.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (projectVersionNodes) {
+      // If found update, detect actual change
+      return PomXml.updateNodes(projectVersionNodes, version);
+    }
+
+    // Try updating project.parent.version
+    const parentVersionNodes: Node[] = [
+      ...findNodes('/project/parent/version', document),
+    ];
+    return PomXml.updateNodes(parentVersionNodes, version);
```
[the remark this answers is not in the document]
mdvorak: Ok, it is matter of an opinion, but I can rework it so it will be handled by a plugin instead. I've just tried to provide out-of-the-box simple solution with minimal config, that would work with most common maven projects. You want something extra? Don't use maven strategy, use java strategy and with replaces driven by comments.

If we will rewrite this, I would suggest however to move this to separate PR, created after this is merged (that is, reducing the scope of this change and making it more managable).

src/updaters/java/pom-xml.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  private static updateNodes(nodes: Node[], value: string) {
```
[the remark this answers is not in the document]
mdvorak: Ah, me too, I've just missed this.

src/manifest.ts
mdvorak: This should not be breaking, this is actually a private method used only directly in the `manifest.ts` itself. 
But possible breaking change is done to `Strategy`, where this method is added.  That is, if anyone would implement the interface, and not extend the `BaseStrategy`. I highly doubt that. Still, I will make the `Strategy.isValidRelease()` optional to avoid that breaking change.

mdvorak: I've changed the interface `Strategy` in 7da1b08428eef15ff25fc281423e66a861cb3b3a - the `lastestReleaseVersion` signature is left intact however.

mdvorak:
> I don't quite follow the difference between the `java` strategy and the `maven` strategy.

Note taken, I've simplified it, which resulted in better cleaner code. Thanks for the kick in the right direction.

Strategies are now set as follows:

- `java` - generates SNAPSHOT version after each release, universal strategy that does not update any files on its own.
- `maven` - It generates SNAPSHOT version after each release, and updates all found pom.xml files automatically
- `java-yoshi` - As before.

`PomXml` updater still behaves the same (see discussion about `project.parent.version` updates) for now.

src/strategies/java.ts
mdvorak: Resolved by refactoring.

src/strategies/base.ts
mdvorak: Done

src/factory.ts
```diff
@@ -56,11 +58,13 @@ const allReleaseTypes = [
   'go',
   'go-yoshi',
   'helm',
+  'java',
```
chingor13, line 61: Do we expect folks to ever use this strategy directly if it doesn't update any files? It's possible to pair with the `extra-files`.

chingor13 (commented):
This is looking fantastic

src/factory.ts
mdvorak: I believe it makes sense to make it available, since it can handle some niche cases. where all replaces will be handled by `extra-files`.

src/updaters/java/pom-xml.ts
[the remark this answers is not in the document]
chingor13: Ok, I'm fine with this as it is. I can see this working nicely for a multi-module project where all the modules are on the same version and that version is defined in the parent.

If a maintainer need different versions, then you would need to configure a multi-path manifest with a java-workspace plugin to make sure cross dependency versions are updated appropriately.

src/factory.ts
chingor13: SGTM

mdvorak:
@chingor13 Looks like I'm done, can you please do final review? Thank you

chingor13 (approved):
Thank you!

chingor13 (approved):

PR #1337, opened 2022-03-14
feat: introduce generic xml updater

chingor13:
For most release strategies, you can provide additional files to update
using the [GenericXml](/src/updaters/generic-xml.ts) updater. You can
specify a configuration object in the `extra-files` option in the manifest
configuration.

```json
{
  "extra-files": [
    {
      "type": "xml",
      "path": "path/to/file.xml",
      "xpath": "//xpath/to/field"
    }
  ]
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "@types/pino": "^7.0.0",
     "@types/semver": "^7.0.0",
     "@types/sinon": "^10.0.0",
+    "@types/xmldom": "^0.1.31",
     "@types/yargs": "^17.0.0",
     "c8": "^7.0.0",
     "chai": "^4.2.0",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -85,6 +86,8 @@
     "typescript": "^3.8.3",
     "unist-util-visit": "^2.0.3",
     "unist-util-visit-parents": "^3.1.1",
+    "xmldom": "^0.6.0",
+    "xpath": "^0.0.32",
```
bcoe, line 90: I don't love that `xpath` feels a bit like abandonware,but this seems like the best option available.

bcoe (approved):
I wish xmldom/xpath were better maintained ecosystems, it jumped out at me that I believe `xpath` is a fork from the original?

Out of curiosity, did you try jsdom on XML files?

chingor13:
> did you try jsdom on XML files

I tried several alternatives. Most basic DOM based implementations are built to html specs and do no handle XML is a sufficient way (like dropping the declaration and reformatting in uncontrollable ways).

PR #1342, opened 2022-03-21
feat: extensible factories

### LINKED ISSUE #1341: Make release-please package extendable
Right now, release-please is customizable, but only by forking it. 
By making it possible to register new strategies (etc) in `factory.ts` by 3rd party, it would be extendable by using pre-built npm package instead. I did not find any reason why configuration should be static.

Here is rough PR: #1342

mdvorak:
Extend current code by registering new plugin/strategy before invoking release-please. This makes possible to extend this tool by composition, instead of inheritance (forking).

Closes #1341

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -117,25 +117,11 @@ export async function buildStrategy(
     changelogSections: options.changelogSections,
   });
   const strategyOptions: BaseStrategyOptions = {
-    github: options.github,
+    skipGitHubRelease: options.skipGithubRelease, // Note the case difference in GitHub
+    ...options,
     targetBranch,
-    path: options.path,
-    bumpMinorPreMajor: options.bumpMinorPreMajor,
-    bumpPatchForMinorPreMajor: options.bumpPatchForMinorPreMajor,
-    component: options.component,
-    packageName: options.packageName,
-    changelogPath: options.changelogPath,
-    changelogSections: options.changelogSections,
     versioningStrategy,
-    skipGitHubRelease: options.skipGithubRelease,
-    releaseAs: options.releaseAs,
-    includeComponentInTag: options.includeComponentInTag,
     changelogNotes,
-    pullRequestTitlePattern: options.pullRequestTitlePattern,
-    extraFiles: options.extraFiles,
-    tagSeparator: options.tagSeparator,
-    versionFile: options.versionFile,
-    snapshotLabels: options.snapshotLabels,
   };
 
   const builder = releasers[options.releaseType];
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -21,10 +21,10 @@ import {OCaml} from './strategies/ocaml';
 import {PHP} from './strategies/php';
 import {PHPYoshi} from './strategies/php-yoshi';
 import {Python} from './strategies/python';
-import {Ruby, RubyStrategyOptions} from './strategies/ruby';
-import {RubyYoshi, RubyYoshiStrategyOptions} from './strategies/ruby-yoshi';
+import {Ruby} from './strategies/ruby';
+import {RubyYoshi} from './strategies/ruby-yoshi';
 import {Rust} from './strategies/rust';
-import {Simple, SimpleStrategyOptions} from './strategies/simple';
+import {Simple} from './strategies/simple';
 import {TerraformModule} from './strategies/terraform-module';
 import {Helm} from './strategies/helm';
 import {Elixir} from './strategies/elixir';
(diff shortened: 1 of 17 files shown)


src/factory.ts
```diff
@@ -21,98 +21,89 @@ import {OCaml} from './strategies/ocaml';
...
-type ReleaseBuilder = (options: BaseStrategyOptions) => Strategy;
-type Releasers = Record<string, ReleaseBuilder>;
-const releasers: Releasers = {
+
+// TODO add JavaStrategyOptions
+export type AnyStrategyOptions =
+  | BaseStrategyOptions
+  | RubyStrategyOptions
+  | RubyYoshiStrategyOptions
+  | SimpleStrategyOptions;
```
chingor13, line 57: I'm wondering if I made a mistake in organizing these option types. Perhaps we should have gone with a general `StrategyOptions` which is a superset of all the possible strategy options and each strategy only picks and chooses what it uses.

We could rectify by turning `BaseStrategyOptions` into that superset of options (because it's currently part of the exported public interface) and remove the extra strategy option types. This would simplify the `ReleaseBuilder` type to always be `(options: BaseStrategyOptions) => Strategy` and we'd never have to add another type to the `AnyStrategyOptions` union type (and reduce the number of types we have to expose).

src/index.ts
```diff
@@ -13,14 +13,39 @@
...
+  VersioningStrategyFactoryOptions,
   VersioningStrategyType,
-  getVersioningStrategyTypes,
-  ChangelogNotesType,
   getChangelogTypes,
+  getPluginTypes,
+  getReleaserTypes,
+  getVersioningStrategyTypes,
+  registerChangelogNotes,
+  registerPlugin,
```
chingor13, line 41: A couple of these exports don't seem to be from the correct source file.

chingor13 (changes requested):
This is looking nice.

Can you update from main and resolve the minor conflicts from the java feature?

src/index.ts
mdvorak: I'll take a look.

mdvorak: Ok, so its matter of opinion, IDEA always suggests to use shortest import path possible (thus in case of a re-export in factory it suggests to use that). I actually agree with that, since its always better to use higher-level abstraction, even with node modules.
I'll modify it to always use source file directly.

src/factory.ts
mdvorak: I agree, specific options are in manifest configs anyway, so trying to split them in this specific place does not provide real loose coupling, only adds complexity.
Additionally, `SimpleStrategyOptions` is exactly same as `RubyStrategyOptions` and `RubyYoshiStrategyOptions` (all adds only `versionFile?: string` property).

So, idea is to add `versionFile` and `snapshotLabels` to `BaseStrategyOptions` and remove specific types altogether? For backward compatibility, I would just add type aliases for old names.

mdvorak:
Note: I've force-pushed rebased commit, it was safer to do than merge.

src/factory.ts
mdvorak: To be exact, all options are mushed together in `ReleaserConfigJson` atm.

src/factory.ts
```diff
@@ -131,7 +123,7 @@ export async function buildStrategy(
     github: options.github,
     changelogSections: options.changelogSections,
   });
-  const strategyOptions: BaseStrategyOptions = {
+  const strategyOptions: AnyStrategyOptions = {
```
mdvorak, line 126: I'm thinking about changing `buildStrategy` to use expansion instead of assign of each field manually.

Sidenote: There is case difference in one option:
`ReleaserConfig.skipGithubRelease`
vs
`BaseStrategyOptions.skipGitHubRelease`
rest is 1:1.

```ts
    const strategyOptions: BaseStrategyOptions = {
    ...options,
    skipGitHubRelease: options.skipGithubRelease,
    targetBranch,
    versioningStrategy,
    changelogNotes,
  };
```

It is not needed, just an idea.

chingor13: that sounds find to me - can you add a comment to the `skipGitHubRelease: options.skitGithubRelease` so that we can rename/clean it up in the next major version.

chingor13 (approved):
This looks great!

The only weirdness I see is that the tests for registering new types leaves the entries registered (because we're basically using global scope), but I don't see a good way around that right now.

Up to you if you want to do the field expansion change in this PR or a follow-up cleanup.

mdvorak:
> The only weirdness I see is that the tests for registering new types leaves the entries registered (because we're basically using global scope), but I don't see a good way around that right now.

You are right about that, this might break some tests if they would be executed in a different order (especially `cli.ts` which have lists of supported types in a snapshot).
I've added `unregister*` functions, which are only used in tests. I know, it is an anti-pattern, but tests depending on execution order are a problem just waiting to happen. I'd call it lesser evil 👿

mdvorak:
Ready :)

chingor13 (approved):

Issue #1331, opened 2022-03-09
Keep non-conforming commits as undefined type


mdvorak:
Discussion for #1326 

Conventional commits are great, but people make mistakes. Commits without conventional format got thrown away completely. This patch allows them to be displayed in Release Notes, and passed to a custom Release Notes generator as well. That way, information from those commits won't be lost during release.

Use-cases

- Enable undefined type and simply append non-conforming commits to Release Notes
- Support "dirty" repo not following conventional commits spec completely (this can happen easily when for example forking repo and updating from upstream)
- Custom Release Notes generator, which might need these commits to work with. This can be accomplished also by passing raw commit log to ChangelogNotes, which might be better for this use-case.

Since it adds a new (by default) hidden type, change should not be breaking.

mdvorak:
So what should be next steps? I'm willing to rework the PR as needed.

georgeguimaraes:
that's a great feature. I too feel like people do mistakes and their commits/PRs should not be ignored. 

should we consider undefined commits as releasable? if not, I guess we only need to update both `changelog-notes` to include undefined commits, right?

PR #1343, opened 2022-03-21
refactor: pass raw commit log to ChangelogNotes.buildNotes

mdvorak:
Related to #1331 and #1341, this improves extendibility of the tool. 

Sometimes changelog generator might use raw commits as a source (as actually does `GitHubChangelogNotes` in a way), or combination of both.


chingor13 (approved):
Thanks, this seems like a reasonable refactor.

PR #1345, opened 2022-03-22
test: add failing test for supporting the snooze feature

### LINKED ISSUE #1339: Snooze release PR
Release-Please currently tries to always keep a release PR open if there are changes to be released. Sometimes, you may not want to release a PR yet.

This feature request is to have the ability to snooze a release PR (close it), but only reopen it if there are new changes.

A use case, you merge a docs bugfix that could be released, but you don't want to release it yet. You would close it and "snooze" it until the next bugfix lands and you can choose to release it or snooze it again.

chingor13:
Fixes #1339

### DIFF
--- a/test/manifest.ts
+++ b/test/manifest.ts
@@ -48,11 +48,12 @@ const fixturesPath = './test/fixtures';
 
 function mockPullRequests(
   github: GitHub,
-  pullRequests: PullRequest[],
-  mergedPullRequests: PullRequest[] = []
+  openPullRequests: PullRequest[],
+  mergedPullRequests: PullRequest[] = [],
+  closedPullRequests: PullRequest[] = []
 ): sinon.SinonStub {
   async function* fakeGenerator() {
-    for (const pullRequest of pullRequests) {
+    for (const pullRequest of openPullRequests) {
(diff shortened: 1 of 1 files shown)


src/manifest.ts
```diff
@@ -713,24 +714,50 @@ export class Manifest {
     }
     logger.info(`found ${openPullRequests.length} open release pull requests.`);
 
+    const snoozedPullRequests: PullRequest[] = [];
```
bcoe, line 717: I think it would be worthwhile to encapsulate this logic in a helper.

src/manifest.ts
```diff
@@ -749,6 +776,25 @@ export class Manifest {
         }
       );
       return updatedPullRequest;
+    } else if (snoozed) {
+      // If unchanged, no need to push updates
+      if (snoozed.body === pullRequest.body.toString()) {
+        logger.info(
```
bcoe, line 782: I also think it would be worth pulling this logic into a helper, as the enclosing function is getting a bit long.

bcoe (approved):
Left a small nit about pulling snooze logic into helper.

PR #1346, opened 2022-03-22
feat: add dotnet-yoshi strategy

chingor13:
feat: allow configuring tag without "v"


chingor13:
> I'm not a huge fan of adding complexity just to support dropping a `v` prefix, is this feature a must?

This is to support google-cloud-dotnet's tags

bcoe (approved):

PR #1365, opened 2022-04-07
feat: provide option for overriding GitHub host

### LINKED ISSUE #1255: Provide option for overriding GitHub host
If someone is using GitHub enterprise, then the links should be relative to whatever domain they are using.

goatwu1993:

Fixes #1255 🦕


chingor13 (changes requested):
Thanks for submitting this.

Can you add a test for this option? The args parser is parsing the option, but we do not appear to be passing it down to the arguments for `buildNotes()`.

Issue #1366, opened 2022-04-10
Add Support for Files Manifest Package Paths


TigerC10:
Thanks for stopping by to let us know something could be better!

 **Is your feature request related to a problem? Please describe.**
Right now, the package paths for manifest releaser is required to be a directory.  However, I've got several "single file" components that live in a single directory.  It's frustrating that this is referred to as a _path_ and not a _directory_, took a lot of reading of the source code to understand this limitation.

 **Describe the solution you'd like**
I would like the package paths in the manifest releaser to support individual files.

 **Describe alternatives you've considered**
The only alternative is to split the single file components up into separate directories, which seems a little heavy handed.  A directory with 1 file in it is overkill.

 **Additional context**
N/A

chingor13:
Can you describe this use case a bit more?

In general, each strategy instance is responsible for updating certain files within a directory and knowing what the current and next version of that component should be. When a single file component is updated, which file(s) do you expect to be updated with versions (as well as the CHANGELOG.md)?

TigerC10:
> In general, each strategy instance is responsible for updating certain files within a directory and knowing what the current and next version of that component should be. When a single file component is updated, which file(s) do you expect to be updated with versions (as well as the CHANGELOG.md)?

The Manifest strategy is a little different compared to other strategies, in that it "wraps around" other strategies.  For my use case, I'm leveraging the simple strategy for all of my single-file components (so, manifest maintains version and changelog outputs to the same directory as the single file components).  The code in #1367 enables the simple strategy to be informed of changes to my single file component, and simple strategy behaves exactly as I expect it to.

The use case is that I have a monorepo style structure (perfect for manifest strategy), but while the bulk of the project is a complex project with a subdirectory that contains many files to be updated... There is a smaller `lib` directory that holds 12 different single file components which need to be versioned and published to a public asset directory (think javascript file uploaded to S3 storage bucket).  I would really rather not create a _folder_ with a single file in it for each of these 12 single file components.

Another use case is documentation.  Say I've got a set of OpenAPI specification documents in a single repo.  It's overkill to put them each in a folder dedicated just for a single file, but I need to version the OpenAPI spec and publish it separately.  If not OpenAPI specs, what about a set of governing policy documents written in Markdown?  These would likely be grouped in folders for their applicable department/locality/<insert applicable target here> rather than creating a folder for each policy document.

Anyway, I'm just saying: it'd be a lot better to allow for component paths to point to a specific file rather than force it to be a directory.

[...]

PR #1367, opened 2022-04-11
feat: Adding support for manifest strategy paths to be files

TigerC10:


TigerC10:

Fixes #1366 🦕

ScottGuymer:
I have found a similar use case where I would like to version and release single files.

In this case it is for GitHub Reusable Workflows. I have a number of shared workflows that I keep in a repo but GitHub mandate that they must all live in the `.github/workflows` folder and cant be nested or moved from there.

ScottGuymer:
@chingor13 is this something that could move forward?

ScottGuymer:
@yoshi-nodejs can this be reviwed?

npalm:
@yoshi-nodejs would be great when the module can support file-based releases.

TigerC10:
As my fork was fairly out of date, I've updated my fork to be in sync with the latest upstream.  No conflicts during update sync.

src/manifest.ts
```diff
@@ -1208,7 +1211,8 @@ export class Manifest {
         const strategy = await buildStrategy({
           ...config,
           github: this.github,
-          path,
+          path:
+            existsSync(path) && lstatSync(path).isFile() ? dirname(path) : path,
```
chingor13, line 1215: There is no cloned git repo so these `fs` commands are all going to return `false` all the time.

chingor13 (changes requested):
This change of behavior will need tests to make sure it works (it currently will not work because the `fs` functions won't work on GitHub repo paths when there is no cloned version on the filesystem).

PR #1412, opened 2022-05-04
feat: allow configuring separate-pull-requests per component

### LINKED ISSUE #1411: One PR keeps getting overwritten by two conflicting configurations
#### Steps to reproduce

  1. I have two release PRs configured at the base of the repository. One is a manifest release for many projects I oversee and the other is a non-manifest. The two I don't think can be combined, I could be wrong, because of I need the component to not do anything for the root release but for the manifest release it is used for proper tagging. [config 1](https://github.com/googleapis/google-cloud-go/blob/main/release-please-config-yoshi-submodules.json) [config 2](https://github.com/googleapis/google-cloud-go/blob/main/release-please-config.json)
  2. Because the action for both is configured at the root of the repository both releases fight over the same PR. The last one to run wins. [action config](https://github.com/googleapis/google-cloud-go/blob/main/.github/release-please.yml) I can't really show an example because you need to watch a PR overtime and the changes are blasted away due to the force push nature of release-please.

#### How this could be fixed

1. Have an option to specify a branch prefix for a release PR. That way I could configure these two to be separate PRs
2. In a manifest release support a special component name that will not be included as a part of tagging

chingor13:
Allows configuring `separate-pull-requests` for individual components. They will be excluded from any merged pull request.

Fixes #1411

### DIFF
--- a/src/plugins/merge.ts
+++ b/src/plugins/merge.ts
@@ -55,11 +55,23 @@ export class Merge extends ManifestPlugin {
     }
     logger.info(`Merging ${candidates.length} pull requests`);
 
+    const [inScopeCandidates, outOfScopeCandidates] = candidates.reduce(
+      (collection, candidate) => {
+        if (candidate.config.separatePullRequests) {
+          collection[1].push(candidate);
+        } else {
+          collection[0].push(candidate);
+        }
+        return collection;
+      },
+      [[], []] as CandidateReleasePullRequest[][]
+    );
+
     const releaseData: ReleaseData[] = [];
     const labels = new Set<string>();
     let rawUpdates: Update[] = [];
     let rootRelease: CandidateReleasePullRequest | null = null;
-    for (const candidate of candidates) {
+    for (const candidate of inScopeCandidates) {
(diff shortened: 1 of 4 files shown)


chingor13:
This is not currently working in combination with `includeComponentInTag: false`

src/plugins/merge.ts
```diff
@@ -55,11 +55,23 @@ export class Merge extends ManifestPlugin {
     }
     logger.info(`Merging ${candidates.length} pull requests`);
 
+    const [inScopeCandidates, outOfScopeCandidates] = candidates.reduce(
```
bcoe, line 58: rather than `[[], []] as CandidateReleasePullRequest`, I think you could:

```
candidtes.reduce<Array<Array<CandidateReleasePullRequest>>
```

Or something like that, and get type safety.

bcoe (changes requested):
Nit about using generics.

src/plugins/merge.ts
chingor13: Fixed, that works

bcoe (approved):

PR #1413, opened 2022-05-04
fix: node-workspace should not bump versions for peer dependencies

### LINKED ISSUE #1403: Node peer dependencies cause spurious version bumps
#### Environment details

  - OS: macOS 12.2.1
  - Node.js version: 16.14.0
  - npm version: 8.5.4
  - `release-please` version: 13.13.0

#### Steps to reproduce

1. Create a node workspace with two components, one which is in the `peerDependencies` field of the other (typically you'd use a version range like `1.x`, though this bug happens regardless.) Let's say that `bar` has a peer dependency on `foo`.
2. Make a `feat` commit for package `foo`, causing release-please to create a release PR that bumps the minor version of `foo`.
3. Observe that `bar` will also have a patch version bump in the release PR, but with no change to the version of `foo` in `peerDependencies` field, and a changelog consisting of a 'Dependencies' header and an empty body.

An example of this behaviour can be seen in our repository here: https://github.com/Financial-Times/dotcom-tool-kit/pull/209. Take `@dotcom-tool-kit/babel` as an example in this case: it's only been bumped because its peer dependency, `dotcom-tool-kit`, has been bumped, but no code was changed (excluding `@dotcom-tool-kit/babel`'s version number in its `package.json`.)

### Why this is happening

This is happening because peer dependencies are [included](https://github.com/googleapis/release-please/blob/main/src/plugins/node-workspace.ts#L314) in the dependency graph created in the node workspace plugin, but not included in the [`localDependencies` field from `@lerna/package-graph`](https://github.com/lerna/lerna/blob/main/core/package-graph/index.js#L50) which is [used](https://github.com/googleapis/release-please/blob/main/src/plugins/node-workspace.ts#L235) as the map of all dependencies that are then updated. So peer dependencies are considered when determining what packages need to have their dependencies updated within a workspace, but are not considered when actually doing the update.

### Proposed solution

I'd say peer dependencies should not be considered when determining which packages need dependency bumps – as stated earlier, peer dependencies very rarely use caret version ranges, instead opting to [give ranges as wide as possible](https://nodejs.org/en/blog/npm/peer-dependencies/#using-peer-dependencies) to be as compatible as possible. This is something that is more appropriate to manage manually, rather than being automated by a tool like release-please.

chingor13:
Fixes #1403

### DIFF
--- a/__snapshots__/node-workspace.js
+++ b/__snapshots__/node-workspace.js
@@ -85,6 +85,20 @@ Release notes for path: node1, releaseType: node
 This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).
 `
 
+exports['NodeWorkspace plugin run should ignore peer dependencies 1'] = `
+:robot: I have created a release *beep* *boop*
+---
+
+
+<details><summary>@here/pkgA: 3.3.4</summary>
+
+Release notes for path: node1, releaseType: node
+</details>
+
+---
+This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).
+`
+
 exports['NodeWorkspace plugin run walks dependency tree and updates previously untouched packages 1'] = `
 :robot: I have created a release *beep* *boop*
 ---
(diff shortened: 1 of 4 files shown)


christophehurpeau:
node-workspace should bump versions for peer dependencies but only when it doesn't match version ? i'ts probably something that should be configured

PR #1414, opened 2022-05-04
fix(cargo-workspace): stop defaulting to updating all components

### LINKED ISSUE #1407: Spurious version bumps for workspace dependencies in cargo workspace
Similar to https://github.com/googleapis/release-please/issues/1403, but opening another issue because I'm not sure the root cause is identical.

- `release-please` version: 13.14.0
- Example: https://github.com/CBenoit/my-gh-actions-playground/pull/2
- [Link to repository after merging](https://github.com/CBenoit/my-gh-actions-playground/tree/b-v0.3.0)

#### Steps to reproduce

  1. Start with an initial monorepo containing a cargo workspace of two crates A and B.
  2. Make B depends on A by updating `B/Cargo.toml`.
  3. Push a `feat` commit for crate B only.
  4. Observe the issue.
    a. **A also have a patch version bump in the release PR, but with no change**, and a changelog consisting of a 'Dependencies' header and an empty body.
    b. **`.release-please-manifest.json` is NOT updated with the minor bump of B, resulting in a mismatch between the manifest and the reality.**
    c. The PR body is referring to A using the default package name instead of the one specified in `release-please-config.json` (if any)

chingor13:
It's unclear why we always bump cargo versions for untouched components.

Fixes #1407

### DIFF
--- a/src/plugins/cargo-workspace.ts
+++ b/src/plugins/cargo-workspace.ts
@@ -12,18 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {
-  CandidateReleasePullRequest,
-  RepositoryConfig,
-  ROOT_PROJECT_PATH,
-} from '../manifest';
+import {CandidateReleasePullRequest, ROOT_PROJECT_PATH} from '../manifest';
 import {logger} from '../util/logger';
-import {
-  WorkspacePlugin,
-  DependencyGraph,
-  DependencyNode,
-  WorkspacePluginOptions,
-} from './workspace';
+import {WorkspacePlugin, DependencyGraph, DependencyNode} from './workspace';
 import {
   CargoManifest,
   parseCargoManifest,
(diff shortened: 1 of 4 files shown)


src/plugins/cargo-workspace.ts
```diff
@@ -82,17 +72,6 @@ interface CrateInfo {
  * into a single rust package.
  */
 export class CargoWorkspace extends WorkspacePlugin<CrateInfo> {
-  constructor(
-    github: GitHub,
-    targetBranch: string,
-    repositoryConfig: RepositoryConfig,
-    options: WorkspacePluginOptions = {}
-  ) {
-    super(github, targetBranch, repositoryConfig, {
```
bcoe, line 91: @fasterthanlime, I'm curious what your motivation was for this default behavior?

@chingor13 question on my mind is whether we should still make this configurable.

bcoe (approved):
I'm curious what the motivation for this default behavior was, makes me wonder if it should be configurable.

src/plugins/cargo-workspace.ts
chingor13: Yes, we can expose the option in the manifest, but will do it separately.

PR #1418, opened 2022-05-06
feat: implement python-poetry packages

### LINKED ISSUE #1229: python release type not able to find correct python packages when using poetry

goatwu1993:

Implement [packages](https://github.com/ChenJinGit/poetry/blob/50360f2dc96881398b428a08bd9288c4cc447d4e/docs/docs/pyproject.md#packages) of poetry pyproject.toml

Fixes #1229 🦕

### DIFF
--- a/src/strategies/python.ts
+++ b/src/strategies/python.ts
@@ -100,22 +100,38 @@ export class Python extends BaseStrategy {
       );
     }
 
-    if (!projectName) {
+    const packagePaths = new Set<string>();
+    if (
+      parsedPyProject?.tool?.poetry &&
+      parsedPyProject?.tool?.poetry.packages
+    ) {
+      parsedPyProject?.tool?.poetry.packages.forEach(pkg => {
+        // Use 'from' if it appears, else use 'include'.
+        // Skip glob 'extra/**/*.py by split
+        packagePaths.add((pkg.from || pkg.include).split('/')[0]);
+      });
+    } else if (projectName) {
+      packagePaths.add(projectName);
+    }
+
+    if (packagePaths.size === 0) {
       logger.warn('No project/component found.');
     } else {
-      updates.push({
-        path: this.addPath(`${projectName}/__init__.py`),
-        createIfMissing: false,
-        updater: new PythonFileWithVersion({
-          version,
-        }),
-      });
-      updates.push({
-        path: this.addPath(`src/${projectName}/__init__.py`),
-        createIfMissing: false,
-        updater: new PythonFileWithVersion({
-          version,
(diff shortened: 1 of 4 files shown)


src/strategies/python.ts
```diff
@@ -100,22 +100,38 @@ export class Python extends BaseStrategy {
...
-    if (!projectName) {
+    const packagePaths = new Set<string>();
+    if (
+      parsedPyProject?.tool?.poetry &&
+      parsedPyProject?.tool?.poetry.packages
+    ) {
+      parsedPyProject?.tool?.poetry.packages.forEach(pkg => {
+        // Use 'from' if it appears, else use 'include'.
+        // Skip glob 'extra/**/*.py by split
+        packagePaths.add((pkg.from || pkg.include).split('/')[0]);
```
chingor13, line 111: Could the glob be something like `extra/path/**/*.py`? For example, perhaps there are 2 packages nested under `extra`.

src/strategies/python.ts
```diff
@@ -100,22 +100,38 @@ export class Python extends BaseStrategy {
       );
     }
 
-    if (!projectName) {
+    const packagePaths = new Set<string>();
+    if (
+      parsedPyProject?.tool?.poetry &&
+      parsedPyProject?.tool?.poetry.packages
+    ) {
+      parsedPyProject?.tool?.poetry.packages.forEach(pkg => {
```
chingor13, line 108: nit: the `?.` let's us chain these together. Then in the block, typescript knows that the fields exist so we can drop the `?.`.

```suggestion
      parsedPyProject?.tool?.poetry?.packages
    ) {
      parsedPyProject.tool.poetry.packages.forEach(pkg => {
```

chingor13 (changes requested):
This is looking pretty good.

Question: these are just extra modules (that have an `__init__.py` file) that are all expected to be at the same version together? Would this prevent someone from using a manifest config that has multiple components at different versions?

goatwu1993:
Hmm. I took a look and seems there is more to consider.

I will make some refactoring, maybe split some poetry toml logic and create a new PR

PR #1429, opened 2022-05-13
fix: workspace plugins should update manifest versions

### LINKED ISSUE #1427: Manifest versions are not updated when a package version is bumped due to a dependency (node-workspace)
#### Problem Description

I am working in a monorepo that has dependencies between different packages. We have a package `modal` and a dependant package `typed-variable-modal`. If I create a breaking change `modal` the dependency in `typed-variable-modal` is updated, and a patch update is applied.

However, the version of `typed-variable-modal` in the manifest is not updated.

See [this release-please pull request](https://github.com/BeksOmega/blockly-samples/pull/19) for an example.

#### Environment details

This is triggered by the [release-please bot](https://github.com/googleapis/repo-automation-bots), for a `node` ReleaseType monorepo.

#### Steps to reproduce

  1. Create two packages in a monorepo, where one depends on the other.
  2. Bootstrap the monorepo with the release-please bot (see [example PR](https://github.com/BeksOmega/blockly-samples/pull/15)).
  3. Do a [major version bump](https://github.com/BeksOmega/blockly-samples/pull/18) (I haven't tested other kinds of version bumps) in the non-dependant package.
  4. Create a new release PR through the release-please CI.
  5. Observe how the [manifest file is not properly updated](https://github.com/BeksOmega/blockly-samples/pull/19/files#diff-e38bb01e41ed5baa3c9e8111b86009ca62d4498d2aefd14d367cd6300dd8b126).

chingor13:
Fixes #1427

### DIFF
--- a/src/plugins/workspace.ts
+++ b/src/plugins/workspace.ts
@@ -13,11 +13,16 @@
 // limitations under the License.
 
 import {ManifestPlugin} from '../plugin';
-import {CandidateReleasePullRequest, RepositoryConfig} from '../manifest';
+import {
+  CandidateReleasePullRequest,
+  RepositoryConfig,
+  DEFAULT_RELEASE_PLEASE_MANIFEST,
+} from '../manifest';
 import {logger} from '../util/logger';
 import {VersionsMap, Version} from '../version';
 import {Merge} from './merge';
 import {GitHub} from '../github';
+import {ReleasePleaseManifest} from '../updaters/release-please-manifest';
 
 export type DependencyGraph<T> = Map<string, DependencyNode<T>>;
 export interface DependencyNode<T> {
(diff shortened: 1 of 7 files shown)


src/plugins/workspace.ts
```diff
@@ -148,6 +158,20 @@ export abstract class WorkspacePlugin<T> extends ManifestPlugin {
     );
     newCandidates = await mergePlugin.run(newCandidates);
 
+    if (newCandidates.length === 1) {
```
bcoe, line 161: I'm not quite following the issue that this protects against?

chingor13: A plugin's interface is that it post processes an array of candidate pull requests and returns an array of candidate pull requests. It's possible that the plugin returns more than one, but in this case with the `Merge` plugin, we expect a single one.

bcoe: Non blocking, I think if you added a few targeted comments that described the algorithm, it would help me reason my way through the need for this check.

bcoe (approved):

PR #1432, opened 2022-05-16
fix: CommitSplit with includeEmpty should add commit to all packages

### LINKED ISSUE #1360: Empty commits dependent on what packages have more recently been changed
#### Environment details

  - OS: macOS 12.2.1
  - Node.js version: 16.14.0
  - npm version: 8.5.4
  - `release-please` version: 13.8.0

#### Steps to reproduce

  1. Create a monorepo with multiple packages managed by release-please.
  2. Based on the functionality added in #752, create an empty commit with a `release-as` footer specifying a new version, e.g., `7.7.7`.
  3. Generate a release PR for the empty commit. Instead of creating a release candidate for `7.7.7` for every package in the monorepo, observe that no PR is created.

#### Why this is happening

It appears that this is happening because `commit-split.ts` is only adding the empty commit to packages that have had more recent commits which bump the package's version. Any package that hasn't been modified by subsequent commits won't be included in the empty commit's release. If the empty commit is the latest commit then no packages will be released.

Perhaps we could pre-populate the `splitCommits` object with the path filters set in `this.packagePaths` if that is set?

chingor13:
When using `CommitSplit` with `includeEmpty` and `packagePaths`, include empty commits in all `packagePaths`, not just the ones that already have commits to them.

Fixes #1360

### DIFF
--- a/src/util/commit-split.ts
+++ b/src/util/commit-split.ts
@@ -114,8 +114,15 @@ export class CommitSplit {
         splitCommits[pkgName].push(commit);
       }
       if (commit.files.length === 0 && this.includeEmpty) {
-        for (const pkgName in splitCommits) {
-          splitCommits[pkgName].push(commit);
+        if (this.packagePaths) {
+          for (const pkgName of this.packagePaths) {
+            splitCommits[pkgName] = splitCommits[pkgName] || [];
+            splitCommits[pkgName].push(commit);
+          }
+        } else {
+          for (const pkgName in splitCommits) {
+            splitCommits[pkgName].push(commit);
+          }
         }
       }
     });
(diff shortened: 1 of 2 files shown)


src/util/commit-split.ts
```diff
@@ -114,8 +114,15 @@ export class CommitSplit {
         splitCommits[pkgName].push(commit);
       }
       if (commit.files.length === 0 && this.includeEmpty) {
-        for (const pkgName in splitCommits) {
-          splitCommits[pkgName].push(commit);
+        if (this.packagePaths) {
```
bcoe, line 117: My understanding is that `this.packagePaths` will be populated with all the package names, when you're using a manifest?

It might be worth adding a comment that mentions what this defaults to, as I had to read through the codebase to figure this out.

bcoe (approved):

src/util/commit-split.ts
chingor13: Done, added comments on the options.

PR #1446, opened 2022-05-23
feat: add maven-workspace plugin

chingor13:
This plugin analyzes a Maven monorepo's releases and version bumps any interdependencies. It behaves similarly to the `node-workspace` and `cargo-workspace` plugins.

Example PR: https://github.com/googleapis/java-cloud-bom/pull/4335

feat: PomXml updater can update dependency versions


chingor13:
cc @suztomo

src/plugins/maven-workspace.ts
```diff
@@ -0,0 +1,345 @@
...
+      logger.warn(`Missing project.version in ${path}`);
+      return;
+    }
+    const dependencies: Gav[] = [];
+    const testDependencies: Gav[] = [];
+    for (const dependencyNode of xpath.select(
+      XPATH_PROJECT_DEPENDENCIES,
+      document
+    ) as Node[]) {
+      const parsedNode = parseDependencyNode(dependencyNode);
```
bcoe, line 96: Is the assumption that this list of dependencies represents local dependencies within a monorepo?

bcoe (approved):
I understanding the algorithm correctly that it:

1. loads a list of the packages handled in the monorepo.
1. searches for all POM files in the repo.
1. for each POM file, looks at the `dependencies` stanza, if this artifact name lines up with libraries found in `1.`, updates the dependency?
3.

src/plugins/maven-workspace.ts
chingor13: This is all dependencies, but is only relevant when building the local dependency graph.

suztomo:
Thank you!

PR #1461, opened 2022-06-01
docs: Correct broken link

TimothyJones:
I noticed one of the links in the documentation was broken, so I fixed it:


chingor13 (approved):
Thanks!

PR #1467, opened 2022-06-08
fix: factory errors for unknown types throw ConfigurationError instead of Error

chingor13:
If a repository has configured an invalid type, we now throw a ConfigurationError instead of a generic Error to allow integrations to handle this condition more easily.

Towards https://github.com/googleapis/repo-automation-bots/issues/3870


chingor13:
> is our plan to capture this error and leave a comment on the upstream repo?

Yes, we already catch `ConfigurationError` and open issues.

PR #1488, opened 2022-06-25
docs: add missing option --include-v-in-tags to cli docs

trallnag:
Some time ago the option `--include-v-in-tags` was added in <https://github.com/googleapis/release-please/pull/1346>. As of today this has not been documented in the document for the CLI yet. This PR adds the two missing lines.

Note that there are probably other places where documentation of this new option is still missing.


chingor13 (approved):
Thank you!

The test failure looks like a regression in the upstream `@octokit/types` library.

PR #1053, opened 2021-09-14
feat: add dart updater and releaser

HelloCore:
Fixes #909

PR #1524, opened 2022-07-12
fix(dart): Add multiline flag to replace regex

### LINKED ISSUE #1523: Dart strategy does not update pubspec.yaml in practice
#### Environment details

  - OS: macOS 12.4
  - Node.js version: 18.2.0
  - npm version: 8.9.0
  - `release-please` version: 13.19.3

#### Steps to reproduce

  1. Create a dart repository
  2. Make a change
  3. Run release-please with dart strategy

*Expected*: pubspec.yaml and changelog.md to be updated
*Actual*: Only changelog.md updated

I've identified the issue at play, which is a missing `/m` flag on the version update regex. Pull request to follow.

radicand:
The dart pubspec.yaml is updated here with regex, and the file contains newlines. To properly match the replacement, the `/m` flag is needed for multiline match and replace. Prior to this fix, pubspec.yaml was never updated with the dart strategy.

Fixes #1523 🦕

### DIFF
--- a/src/updaters/dart/pubspec-yaml.ts
+++ b/src/updaters/dart/pubspec-yaml.ts
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {logger} from '../../util/logger';
-import {DefaultUpdater} from '../default';
+import { logger } from '../../util/logger';
+import { DefaultUpdater } from '../default';
 
 /**
  * Updates a Dart pubspec.yaml file.
--- a/src/updaters/dart/pubspec-yaml.ts
+++ b/src/updaters/dart/pubspec-yaml.ts
@@ -40,7 +40,7 @@ export class PubspecYaml extends DefaultUpdater {
       }
     }
     return content.replace(
-      /version: ([0-9.]+)\+?([0-9]*$)/,
+      /version: ([0-9.]+)\+?([0-9]*$)/m,
       `version: ${this.version}${buildNumber}`
     );
   }
--- a/src/updaters/dart/pubspec-yaml.ts
+++ b/src/updaters/dart/pubspec-yaml.ts
@@ -24,23 +24,30 @@ export class PubspecYaml extends DefaultUpdater {
    * @param {string} content The initial content
    * @returns {string} The updated content
    */
+
   updateContent(content: string): string {
-    const oldVersion = content.match(/version: ([0-9.]+)\+?([0-9]*$)/);
+    const oldVersion = content.match(/^version: ([0-9.]+)\+?(.*$)/m);
     let buildNumber = '';
 
     if (oldVersion) {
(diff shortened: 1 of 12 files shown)


chingor13:
Thanks for submitting this, however, the current tests are failing. Also, we'd like to have a failing test case for what this is fixing and to prevent future regressions.

radicand:
Sure thing, I will work on that and update the pull request

radicand:
@chingor13 - I've pushed the requested changes - the failing test case was not possible because the issue at play was the snapshots were bad to begin with. It should be evident in the diff what's changed here and that it now works. Please feel free to follow up with any additional questions.

radicand:
@chingor13 - I think I've resolved all the open issues at this point - please let me know if anything else is needed to merge this in.

src/updaters/dart/pubspec-yaml.ts
```diff
@@ -24,23 +24,30 @@ export class PubspecYaml extends DefaultUpdater {
...
+    const oldVersion = content.match(/^version: ([0-9.]+)\+?(.*$)/m);
     let buildNumber = '';
 
     if (oldVersion) {
-      buildNumber = `${oldVersion[2]}`;
-      if (buildNumber.length > 0) {
+      buildNumber = oldVersion[2];
+      const parsedBuild = parseInt(buildNumber);
+      if (!isNaN(parsedBuild)) {
+        buildNumber = `+${parsedBuild + 1}`;
```
chingor13, line 36: I think this is a behavior change. I think it left the build number intentionally, but I don't know enough about the dart ecosystem to know if this is actually desired or not (the original author [intended it so](https://github.com/googleapis/release-please/pull/1053#discussion_r709567486)). cc @HelloCore

chingor13 (changes requested):
Sorry for the delay in a review.

src/updaters/dart/pubspec-yaml.ts
radicand: I'd welcome feedback from @HelloCore here on intent, however, build number is not forced on the user if it does not exist (preserving existing behavior) - it only increments it if it does exist and is a number. When uploading apps to a store (eg., Google Play), build number can not be duplicated (each upload must have a different one) - there is a very good case for incrementing the build number as a result. It would introduce manual headache to *not* have build number updated.

HelloCore: Actually, I'm okay with either keeping the old build number or increasing it because I mostly override the build number by using the CI build number (`flutter build ipa --build-number=1234`). 
I intended to keep the old build number here just because I don't want to make unnecessary changes on `pubspec.yaml`.

chingor13: Great, thanks for the input. We'll go ahead with this incremennt

chingor13 (approved):

PR #1535, opened 2022-07-26
fix: convert ReleaserConfig JSON keys when bootstrapping

### LINKED ISSUE #1522: bootstrap command provides incorrect config
The bootstrapper is generating the config JSON with camelized keys - not the dasherized keys that are actually used.

chingor13:
We were directly writing the ReleaserConfig to the `release-please-config.json`, but the keys are actually dasherized.

Fixes #1522

### DIFF
--- a/__snapshots__/bootstrapper.js
+++ b/__snapshots__/bootstrapper.js
@@ -1,0 +1,9 @@
+exports['Bootstrapper should open a PR 1'] = `
+{
+  "packages": {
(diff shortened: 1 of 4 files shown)


bcoe (approved):

bcoe:
It would be preferable to make `releaserConfigToJsonConfig` a bit more ⭐ magic ⭐ , but not a blocker for me.

PR #1553, opened 2022-08-05
fix: php root-composer-update-packages version replace

### LINKED ISSUE #1552: Release type PHP Cannot read properties of undefined (reading 'version')
#### Environment details

  - OS: Mac OS 12.5
  - Node.js version: 14.16.0, 16.16.0
  - npm version: 8.15.0
  - `release-please` version: 13.19.8

#### Steps to reproduce

1. Execute `release-please release-pr --repo-url=<repo url> --token=<token> --release-type=php`
2. TypeError: Cannot read properties of undefined (reading 'version')
    at RootComposerUpdatePackages.updateContent (/usr/local/lib/node_modules/release-please/build/src/updaters/php/root-composer-update-packages.js:39:76)

> The version is set in the root of `composer.json`. https://getcomposer.org/doc/04-schema.md#version

anderson-comvex:
The actual `parsed` object from the JSON string is an object and assumes the `version` key is required from the `composer.json`, so we can simply replace the value for the given key.

Since `parsed` is an object, by doing `parsed.replace` assumes you are trying to access the "property" replace from the `parsed`, which `replace` is undefined.

You can see and example in the [codesandbox](https://codesandbox.io/s/release-please-1553-89fvfe?file=/src/index.ts)

Fixes #1552


src/updaters/php/root-composer-update-packages.ts
```diff
@@ -36,9 +36,9 @@ export class RootComposerUpdatePackages extends DefaultUpdater {
       for (let [key, version] of this.versionsMap.entries()) {
         version = version || '1.0.0';
         logger.info(
-          `updating ${key} from ${parsed.replace[key]} to ${version}`
+          `updating ${key} from ${parsed[key]} to ${version}`
```
chingor13, line 39: We are actually intentionally trying to update the `replace` field ([example](https://github.com/googleapis/google-cloud-php/blob/85dc8e73fadd89578dd48fde50dfa8d6cf2cc89c/composer.json#L74)), but we should be resilient to the case when `replace` does not exist.

I suggest we keep this `replace` logic (with a guard only if it's set) and add an additional update of the root `version` value (also only if it's already set -- [the field is optional](https://getcomposer.org/doc/04-schema.md#version))

chingor13 (changes requested):
Thanks for submitting this!

src/updaters/php/root-composer-update-packages.ts
anderson-comvex: @chingor13 

I updated the replace logic and also added tests for the cases you mentioned. So we could increase the test coverage in the given scenarios.

chingor13 (approved):
Thanks!

PR #1566, opened 2022-08-10
feat: allow nesting multiple commits that are parsed independently

chingor13:
Some automations make compound pull requests that need to contain complex nested commits.

In your commit's body, you can include separate commit messages in between `BEGIN_NESTED_COMMIT` and `END_NESTED_COMMIT`. These messages will be extracted and parsed as a separate commit (with the same commit SHA) for the purpose of generating release notes.

Example:
```
fix: some fix

Some rationale for the fix

BEGIN_NESTED_COMMIT
build: drop node12 support

BREAKING CHANGE: We are dropping node 12 support
END_NESTED_COMMIT
BEGIN_NESTED_COMMIT
fix: another fix

Extended commit message for another fix
END_NESTED_COMMIT
```

Towards https://github.com/googleapis/repo-automation-bots/issues/3342


SurferJeffAtGoogle:
With these new keywords, the PR description becomes more machine readable but less human readable.  What other options have you considered?

chingor13:
> What other options have you considered?

I also considered just a bunch of dashes which could be turned into a horizontal rule in markdown. It's possible that this causes unintended behavior if someone is using markdown in their pull request bodies (like renovate PRs perhaps)

chingor13:
Another option is to not build this at all and have owlbot build a `BEGIN_COMMIT_OVERRIDE`/`END_COMMIT_OVERRIDE` that contains the aggregated release notes. This would put the onus on owlbot to aggregate the commit messages for the release notes.

suztomo:
Does this work with "BEGIN_COMMIT_OVERRIDE"?

chingor13:
> Does this work with "BEGIN_COMMIT_OVERRIDE"?

Yes, `BEGIN_COMMIT_OVERRIDE` would work override the commit message, which would then be split via the `BEGIN_NESTED_COMMIT`. Although, I'm not sure why you would override the message to be a complex, nested commit (versus just using several conventional commit messages).

bcoe:
> With these new keywords, the PR description becomes more machine readable but less human readable. What other options have you considered?

What if we provided a mechanism for people to easily preview what their merged commit message will look like, perhaps a form they could test with prior to clicking merge?

Once you've figured out the format of commit messages, through trial and error, it's pretty easily to cleanup the message when merging -- having a way to verify correctness while learning might help people.

SurferJeffAtGoogle:
@chingor13 Is this new syntax enabled by a flag, that defaults to false?  Does it only appear for PRs that have multiple commits?

chingor13:
> Is this new syntax enabled by a flag, that defaults to false?

This is not implemented as a flag - the format is unlikely to be erroneously used.

> Does it only appear for PRs that have multiple commits?

It is parsed for any commit, but likely will only be utilized by owlbot's multi-commit feature.

chingor13:
Note the #1598 is an alternate implementation using markdown `<hr/>` tags (`---`) as separators. That implementation is more likely to possibly have unintended commit splitting because `---` is more likely to be used.

SurferJeffAtGoogle (approved):

PR #1571, opened 2022-08-11
fix: throw underlying API error when the manifest fetch commits when determining the latest released version

### LINKED ISSUE #1567: Version occasionally reverting to initial versions
It appears that when we are using `Manifest.fromConfig()`, our search for the latest release version can return invalid results.

1. We grab the latest X commits on the target branch
2. We look for a merged release pull request
3. We fallback to latest GitHub releases that have a SHA matching a recent commit
4. We fallback to latest GitHub tags that have a SHA matching a recent commit

If step 1 fails (GraphQL), then we have an empty commit set, so step 2, 3, and 4 all fail. Thus release-please assumes we've never created releases before and propose a new release with the initial default version.

Example: https://github.com/googleapis/java-storage/pull/1564 (look at the initial proposed PR).

chingor13:
Occasionally, GitHub's GraphQL endpoint throws 502s when it takes too long to execute a query. We usually silently retry these requests (with exponential backoff). Previously, if too many attempts are made, we return an empty response. We introduce a new `requireResponse` option which will tell our GraphQL executor to throw the underlying API error if a response is required.

Fixes #1567

### DIFF
--- a/src/github.ts
+++ b/src/github.ts
@@ -143,6 +143,8 @@ interface ReleaseHistory {
 interface CommitIteratorOptions {
   maxResults?: number;
   backfillFiles?: boolean;
+  // if set, throw an error if we require a response
+  requireResponse?: boolean;
 }
 
 interface ReleaseIteratorOptions {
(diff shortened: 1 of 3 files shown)


src/github.ts
```diff
@@ -143,6 +143,8 @@ interface ReleaseHistory {
 interface CommitIteratorOptions {
   maxResults?: number;
   backfillFiles?: boolean;
+  // if set, throw an error if we require a response
+  requireResponse?: boolean;
```
bcoe, line 147: Should we just always be throwing after too many failures? Wondering if we need to make this configurable.

chingor13: Ok, configured it to retry 5 times by default with exponential backoff (max of 20 seconds).

bcoe (approved):
LGTM.

PR #1572, opened 2022-08-11
feat: parse versioning type from the manifest config

### LINKED ISSUE #1569: Versioning strategy is ignored when used with manifest releaser
I'm using release-please with the github bot configuration, because of that i'm not sure if this issue belongs here or in the bot repo.

#### Environment details

  - `release-please` version: v13.20.0

#### Config files

**.github/release-please.yml**
```yaml
handleGHRelease: true
manifest: true
versioning: always-bump-patch
```

**release-please-config.json**
```json
{
  "packages": {
    ".": {
      "release-type": "maven",
      "skip-snapshot": true
    }
  }
}
```

#### Issue description

When enabling manifest, other options configured in release-please.yml seem to be ignored. When I disable the manifest option and set releaseType to simple, version strategy is correctly set and only increments minor version.

This seems to be intended, as when using manifest configuration, configs should be in `release-please-config.json`.
However, the `versioning` configuration isn't exposed in the manifest config.

#### Suggested solution

Expose the `versioning-strategies` configuration in `release-please-config.json` so that it can be used with manifest configs.

For example:

```json
{
  "packages": {
    ".": {
      "release-type": "maven",
      "skip-snapshot": true,
      "versioning": "always-bump-patch"
    }
  }
}
```

chingor13:
Fixes #1569

### DIFF
--- a/docs/manifest-releaser.md
+++ b/docs/manifest-releaser.md
@@ -114,9 +114,10 @@ minimal content required defines at least one package:
   }
 }
 ```
-The following example shows all the possibilities. Note: for illustration
-purposes the top level values set here are **NOT** the defaults (those are
-documented in comments)
+The following example shows most of the available options. See the documented
+[JSON schema](/schemas/config.json) for the complete configuration definition.
+Note: for illustration purposes the top level values set here are **NOT** the
+defaults (those are documented in comments)
 ```js
 {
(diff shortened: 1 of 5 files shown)


bcoe (approved):
LGTM.

PR #1580, opened 2022-08-16
feat(node-workspace): Allow to specify version prefix in config

danez:
This allows a `versionPrefix` config option that can be set to a string (like `~`, `^`, ` `, `>=`) that will be used when updating linked local dependencies in a package.

### DIFF
--- a/docs/manifest-releaser.md
+++ b/docs/manifest-releaser.md
@@ -172,6 +172,10 @@ defaults (those are documented in comments)
   // absence defaults to false
   "bump-patch-for-minor-pre-major": true,
 
+  // prefix linked package version updates with this string
+  // absence defaults to `^`
+  "version-prefix": true,
+
   // set default conventional commit => changelog sections mapping/appearance.
   // absence defaults to https://git.io/JqCZL
   "changelog-sections": [...],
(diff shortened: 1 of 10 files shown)


docs/manifest-releaser.md
```diff
@@ -172,6 +172,10 @@ defaults (those are documented in comments)
   // absence defaults to false
   "bump-patch-for-minor-pre-major": true,
 
+  // prefix linked package version updates with this string
+  // absence defaults to `^`
+  "version-prefix": true,
```
chingor13, line 177:
```suggestion
  "version-prefix": "^",
```

chingor13 (changes requested):
Thanks for this pull request and sorry for the delay in the review.

I'm torn on whether this option should be global (like it's implemented here) or should be localized to the `node-workspace` plugin (and potentially other workspace plugins). The reason being that these version prefixes only really apply to dependency declarations (which are handled by workspace plugins).

If localized to the `node-workspace` plugin, we would add a config like:

```json
{
  "plugins": [
    {
      "type": "node-workspace",
      "versionPrefix": "~"
    }
  ]
}
```

and then have the version prefix as a constructor argument to the workspace plugins.

What do you think?

chingor13:
Another option too, is to have the workspace plugins be smarter when updating the dependencies and preserve whatever prefix is currently being used.

danez:
> Another option too, is to have the workspace plugins be smarter when updating the dependencies and preserve whatever prefix is currently being used.

I actually like this the best, will try to find some time to look into this.

PR #1585, opened 2022-08-18
feat: use file-cache from git-file-utils

chingor13:
We extracted the file cache classes into its own utility library. This PR removes the class from this library and uses the version from that utility library.


luxaritas:
Huh looks like it doesn't like the uri format - that what the meta-schema uses. I guess it needs to be added from ajv-formats?

tmatsuo (approved):

PR #1584, opened 2022-08-18
feat: Allow $schema key in manifest config schema

luxaritas:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/release-please/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Fixes #1582 🦕

PR #1589, opened 2022-08-19
fix: Allow $schema key in manifest config schema

luxaritas:
Did some additional testing on #1584 and noticed a couple issues:
1) `$schema` needed to be added to the properties array for it to be picked up
2) If you want to reference the schema via a relative (local) path, you need to use uri-reference instead of uri


chingor13 (approved):

chingor13:
Thanks!

PR #1591, opened 2022-08-19
fix: add REST API call to fetch pull requests without files

chingor13:
The GraphQL API returns 502 for some repositories when iterating over pull requests with files.

This PR adds a new option to the `pullRequestIterator` allowing you to skip fetching files. If you do not need the list of files, we can use the REST API which is much more responsive.

### DIFF
--- a/src/github.ts
+++ b/src/github.ts
@@ -546,16 +546,42 @@ export class GitHub {
   /**
    * Iterate through merged pull requests with a max number of results scanned.
    *
-   * @param {number} maxResults maxResults - Limit the number of results searched.
-   *   Defaults to unlimited.
+   * @param {string} targetBranch The base branch of the pull request
+   * @param {string} status The status of the pull request
+   * @param {number} maxResults Limit the number of results searched. Defaults to
+   *   unlimited.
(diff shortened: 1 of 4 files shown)


src/github.ts
```diff
@@ -546,16 +546,42 @@ export class GitHub {
...
+   * @param {boolean} includeFiles Whether to fetch the list of files included in
+   *   the pull request. Defaults to `true`.
    * @yields {PullRequest}
    * @throws {GitHubAPIError} on an API error
    */
   async *pullRequestIterator(
+    targetBranch: string,
+    status: 'OPEN' | 'CLOSED' | 'MERGED' = 'MERGED',
+    maxResults: number = Number.MAX_SAFE_INTEGER,
+    includeFiles = true
```
bcoe, line 562: nit: I would be tempted to pull this into an interface, so that it's more clear what each field means when invoking:

```
pullRequestIterator(opts: PullRequestOpts = {targetBranch: 'main', status: 'MERGED', maxResults: Number.MAX_SAFE_INTEGER, includeFiles: true});
```

bcoe (approved):
Left a nit, that the interface is a bit opaque in isolation, and that accepting an options bag might be nice.

PR #1598, opened 2022-08-24
feat: allow separator to split multiple commit messages

chingor13:
Alternate to #1566 that uses a markdown horizontal rule as the separator (3 or more dashes on an empty line).

https://github.com/googleapis/repo-automation-bots/issues/3342


chingor13:
Closing in favor of #1566

PR #1607, opened 2022-08-26
feat(plugins): introduce processCommits hook for plugins

bcoe:
Introduces processCommits hook for WorkspacePlugins, and implements the
SentenceCase plugin, which can be used to uppercase the first word
in a commit messages' description.

CC: @martinvarelaj

### DIFF
--- a/docs/manifest-releaser.md
+++ b/docs/manifest-releaser.md
@@ -520,3 +520,7 @@ Example:
   ]
 }
 ```
+
+### sentence-case
+
+Capitalize the leading word in a commit message, taking into account common exceptions, e.g., gRPC.
--- a/docs/manifest-releaser.md
+++ b/docs/manifest-releaser.md
@@ -524,3 +524,15 @@ Example:
 ### sentence-case
 
 Capitalize the leading word in a commit message, taking into account common exceptions, e.g., gRPC.
+
+As an example:
+
+`fix: patch issues in OpenSSL`
+
+Will be output in the CHANGELOG thusly:
+
+```
+Bug Fixes:
+
+* Patch issues in OpenSSL`
+```
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -47,6 +47,8 @@ export abstract class ManifestPlugin {
    * @param {Commit[]} commits The set of commits that will feed into release pull request.
    * @returns {Commit[]} The modified commit objects.
    */
+  // TODO: for next major version, let's run the default conventional commit parser earlier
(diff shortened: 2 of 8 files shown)


docs/manifest-releaser.md
```diff
@@ -520,3 +520,7 @@ Example:
   ]
 }
 ```
+
+### sentence-case
+
+Capitalize the leading word in a commit message, taking into account common exceptions, e.g., gRPC.
```
chingor13, line 526: nit: can you add an example

src/plugins/sentence-case.ts
```diff
@@ -0,0 +1,71 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {ManifestPlugin} from '../plugin';
+import {Commit} from '../commit';
+
+// A list of words that should not be converted to uppercase:
+const SpecialWords: Set<string> = new Set(['gRPC', 'npm']);
```
chingor13, line 19: nit: let's make this constant `SPECIAL_WORDS`
```suggestion
const SPECIAL_WORDS: Set<string> = new Set(['gRPC', 'npm']);
```

src/plugin.ts
```diff
@@ -42,6 +42,15 @@ export abstract class ManifestPlugin {
     this.logger = logger;
   }
 
+  /**
+   * Perform post-processing on commits, e.g, sentence casing them.
+   * @param {Commit[]} commits The set of commits that will feed into release pull request.
+   * @returns {Commit[]} The modified commit objects.
+   */
+  processCommits(commits: Commit[]): Commit[] {
```
chingor13, line 50: TODO for next major version, let's run the default conventional commit parser earlier (outside of the strategy classes) and pass in the `ConventionalCommit[]` objects in. This will make it cleaner for plugin commit processors and the commit post-processing in the strategies.

chingor13 (commented):
Similar to https://github.com/googleapis/repo-automation-bots/pull/4125#pullrequestreview-1067349100, can we make the list of special words configurable?

docs/manifest-releaser.md
bcoe: Done.

src/plugins/sentence-case.ts
bcoe: done.

src/plugin.ts
bcoe: Added TODO.

PR #1621, opened 2022-09-01
feat: allow specifying glob option for extra-files

### LINKED ISSUE #1619: Support `extra-files` globs
Sometimes we want to attempt to bulk update N extra files. For example, in https://github.com/chingor13/functions-framework-dotnet/pull/7, we want to update a bunch of `.csproj` files but do not necessarily want to keep updating the config every time we add a new `.csproj` file.

Investigate iterating over files with a file glob:

```json
{
  "extra-files": [
    {
      "type": "xml",
      "glob": "src/**/*.xml",
      "xpath": "//xpath/to/version",
    }
  ]
}
```

chingor13:
When specifying `extra-files` in a manifest, you can add a `glob: true` entry and the `path` specified will be treated as a file glob.

Fixes #1619

### DIFF
--- a/src/strategies/base.ts
+++ b/src/strategies/base.ts
@@ -352,6 +352,12 @@ export abstract class BaseStrategy implements Strategy {
           this.targetBranch
         )
       ).map(file => `/${file}`);
+    } else if (this.path === ROOT_PROJECT_PATH) {
+      // root component, ignore path prefix
+      return this.github.findFilesByGlobAndRef(
+        extraFile.path,
+        this.targetBranch
+      );
     } else {
       // glob is relative to current path
       return this.github.findFilesByGlobAndRef(
--- a/src/strategies/base.ts
+++ b/src/strategies/base.ts
@@ -311,7 +311,7 @@ export abstract class BaseStrategy implements Strategy {
       latestVersion: latestRelease?.tag.version,
     });
     const updatesWithExtras = mergeUpdates(
-      updates.concat(...this.extraFileUpdates(newVersion, versionsMap))
+      updates.concat(...(await this.extraFileUpdates(newVersion, versionsMap)))
     );
     const pullRequestBody = await this.buildPullRequestBody(
       component,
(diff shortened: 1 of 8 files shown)


src/strategies/base.ts
```diff
@@ -333,51 +333,96 @@ export abstract class BaseStrategy implements Strategy {
     };
   }
 
-  protected extraFileUpdates(
+  // Helper to convert extra files with globs to the file paths to add
+  private async extraFilePaths(extraFile: ExtraFile): Promise<string[]> {
+    if (typeof extraFile !== 'object') {
```
tmatsuo, line 338: Why this happens? Does something pass string to this function?

src/strategies/base.ts
```diff
@@ -333,51 +333,96 @@ export abstract class BaseStrategy implements Strategy {
...
+      return [extraFile];
+    }
+
+    if (!extraFile.glob) {
+      return [extraFile.path];
+    }
+
+    if (extraFile.path.startsWith('/')) {
+      // glob is relative to root, strip the leading `/` for glob matching
```
tmatsuo, line 347: I don't understand why the glob doesn't work with the leading slash ...

test/fixtures/manifest/config/extra-files-glob.json
```diff
@@ -0,0 +1,30 @@
+{
```
tmatsuo, line 1: Is this file used?

src/strategies/base.ts
chingor13: Yes, this config is coming from a config file which is an array of a set of different types -- one of which is a string.

https://github.com/googleapis/release-please/blob/main/schemas/config.json#L114-L178

test/fixtures/manifest/config/extra-files-glob.json
chingor13: All config files in that directory are turned into tests that verify that they adhere to the JSONSchema

src/strategies/base.ts
chingor13: We could perhaps fix this in the upstream library, but the filenames coming from GitHub do not have a leading `/` (so the leading slash on the glob does not match)

tmatsuo: I see!

test/fixtures/manifest/config/extra-files-glob.json
tmatsuo: thx

tmatsuo (approved):

PR #1644, opened 2022-09-19
fix: detect uppercase readme.md files for TF

### LINKED ISSUE #1643: TF Readme updater not detecting README.md
I believe this is due to some underlying API change where `findFilesByFilenameAndRef` became case sensitive. Previously `README.md` used to be updated https://github.com/terraform-google-modules/terraform-google-cloud-storage/pull/77/files

bharathkkb:

Fixes #1643 🦕

### DIFF
--- a/src/strategies/terraform-module.ts
+++ b/src/strategies/terraform-module.ts
@@ -39,11 +39,20 @@ export class TerraformModule extends BaseStrategy {
 
     // Update version in README to current candidate version.
     // A module may have submodules, so find all submodules.
-    const readmeFiles = await this.github.findFilesByFilenameAndRef(
-      'readme.md',
-      this.targetBranch,
-      this.path
-    );
+    const readmeFiles = await Promise.all([
+      this.github.findFilesByFilenameAndRef(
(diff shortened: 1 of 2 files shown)


bharathkkb:
Tested via CLI in https://github.com/bharathkkb-org/terraform-google-cloud-storage/pull/30/files

chingor13 (approved):
An alternative is to either make the `findFilesByFilenameAndRef` helper do a case insensitive search.

chingor13:
> An alternative is to either make the `findFilesByFilenameAndRef` helper do a case insensitive search.

We can do this in a followup. This implementation should not actually add any extra API calls as we're using a file cache.

bharathkkb:
> An alternative is to either make the findFilesByFilenameAndRef helper do a case insensitive search.

Yeah I was debating but did not want to break any other updater if this was expected. FWIW, initially I think findFilesByFilenameAndRef was case insensitive before it stopped working for us (https://github.com/terraform-google-modules/terraform-google-cloud-storage/pull/77/files)

PR #1646, opened 2022-09-19
feat: added expo strategy and updater

### LINKED ISSUE #1568: React Native Expo support (or custom Strategy/Parser path configuration)
I'd like to use `release-please` to release React Native Expo projects. These projects will require a custom Strategy and Parser because they store version information in a file called `app.json`. The version numbers in `app.json` include a non-semver versioning pattern for the Android version number but semver for the iOS and web versions in the same file. Reading through the docs and the code, it doesn't seem possible to provide a path to my own custom Strategy and Parser files.  

Either of:
- My preferred solution would be to allow configuring `release-please` with the path to my own custom Strategy/Parser files. 
- I'm also happy to submit a PR with a React Native Expo Strategy if you're open to that.

The current `release-please` API will require me to update the `app.json` version numbers in a separate CI process after the release PR is merged. This means that there will always be a second commit right after the release commit and the releases don't accurately describe the state of the release. I could instead amend the PR commit with the updated `app.json` but I'm hesitant to amend git history in an automated process.

I'm open to any other options for achieving this with the existing API.

soulfresh:
Added `strategies/expo` and `updaters/expo/app-json` for versioning of Expo based React Native projects. This updater performs the same functions as the `node` updater but also updates the `app.json` file found in Expo projects.

Fixes #1568  🦕


src/updaters/expo/app-json.ts
```diff
@@ -0,0 +1,91 @@
...
+    // @see https://medium.com/@maxirosson/versioning-android-apps-d6ec171cfd82
+    const versionCode =
+      expoMajorVersion * 10000000 +
+      this.version.major * 10000 +
+      this.version.minor * 100 +
+      this.version.patch;
+    logger.info(
+      `updating Android version from ${parsed.expo.android.versionCode} to ${versionCode}`
+    );
+    parsed.expo.android.versionCode = versionCode.toString();
```
soulfresh, line 87: @chingor13 Here's the Android `versionCode` update strategy for Expo based projects that we discussed a couple weeks ago. As you can see, the `versionCode` is based on semver but also includes the Expo SDK version number. This is the implementation used on most Expo projects.

src/strategies/expo.ts
```diff
@@ -0,0 +1,55 @@
...
+      updater: new AppJson({version, expoSDKVersion}),
+    });
+
+    return updates;
+  }
+
+  /**
+   * Determine the Expo SDK version by parsing the package.json dependencies.
+   */
+  async getExpoSDKVersion(): Promise<string> {
```
chingor13, line 44: We prefer to pass around `Version` instances rather than version strings.
```suggestion
  async getExpoSDKVersion(): Promise<Version> {
```

You can return `Version.parse(versionString)`.

src/updaters/expo/app-json.ts
```diff
@@ -0,0 +1,91 @@
...
+      buildNumber: string;
+    };
+    android: {
+      versionCode: string;
+    };
+  };
+}
+
+export interface AppJsonOptions extends UpdateOptions {
+  expoSDKVersion: string;
```
chingor13, line 33: same for here: prefer to use `Version` type.

chingor13 (changes requested):
Thanks for this! It's almost there.

You'll also need to add to `src/factory.ts` so that `release-please` can load the `expo` strategy.

src/strategies/expo.ts
soulfresh: This should have been obvious to me looking at the code 🤦

soulfresh:
@chingor13 I made the updates you suggested. I obviously haven't tested this on a repo yet or I would have caught the fact that the new expo type wasn't available on the factory. I got pulled into something else but I will test my fork against our project here soon. 

Additionally, there are 4 tests that don't pass for me on a fresh checkout of the repo. I've setup the project as described in `CONTRIBUTING.md` but I'm seeing test errors in the `CLI` test like the following:
```
 1) CLI
       --help
         release-pr:
     Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/marc/Development/apartment-snapshot/release-please/build/test/cli.js)
      at listOnTimeout (node:internal/timers:557:17)
      at processTimers (node:internal/timers:500:7)
```

All other tests/checks pass.

soulfresh:
@chingor13 I just figured out what those test timeouts were. I updated the cli fixtures with the updated help text and now the tests should be passing.

src/updaters/expo/app-json.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    logger.info(
+      `updating Expo version from ${parsed.expo.version} to ${this.version}`
+    );
+    parsed.expo.version = this.version.toString();
+
+    logger.info(
+      `updating iOS version from ${parsed.expo.ios.buildNumber} to ${this.version}`
+    );
+    parsed.expo.ios.buildNumber = this.version.toString();
```
marijnvdwerf, line 62: Question: does this also work if the app.json doesn't have this data (`ios.buildnumber`)? Expo doesn't require you to have it

soulfresh:
Sorry for the slow response. I was out of town and then caught a cold. I just pushed an update with your suggestions. Good catch. 

However, I noticed that Expo has changed their documentation around versioning since the last time I researched it. They previously recommended using the `standard-version-expo` npm package for versioning `app.json` which formed the basis for the versioning strategy I used in this PR. However, they've recently updated their documentation around versioning to suggest multiple different ways that Expo applications can use/set the iOS `buildNumber` and Android `versionCode`. This appears to be related to the new cloud build service (EAS) they've introduced. Here's the updated documentation I'm referring to: https://docs.expo.dev/build-reference/app-versions/

If the strategy for versioning `buildNumber` and `versionCode` can no longer be expect to be consistent between Expo projects, I'm no longer sure that this feature should be added to `release-please`. It seems like Expo projects should just use the `node` project type and find some other way to increment the `buildNumber` and `versionCode`.

Then again, merging this PR still provides some value for users who don't want to have to manage `buildNumber` and `versionCode` separately from the semver version. 🤷‍♂️

chingor13 (approved):
Thanks!

PR #1651, opened 2022-09-22
fix: handle pull request files returned as null

chingor13:
Towards https://github.com/googleapis/repo-automation-bots/issues/4477


bcoe (approved):
Do you have any idea why this list is returned as null, are there actually no files associated?

PR #1660, opened 2022-09-27
feat: add `group-priority` plugin

### LINKED ISSUE #1657: Java monorepo release PRs can contain mix of snapshot and non-snapshot PRs
Ideally, we would like to separate the snapshot PRs from the non-snapshots

chingor13:
This plugin allows you to configure pull request by priority. If enabled and if a pull request of a prioritized group is found, release-please will limit the proposed release pull requests to the prioritized group only.

For example, java snapshot PRs are now marked as part of the `snapshot` group. If you configure the `group-priority` plugin with the group set to `['snapshot']`, then release-please will only open pull requests for `snapshot` pull requests if there are any.

Fixes #1657

### DIFF
--- a/src/factories/plugin-factory.ts
+++ b/src/factories/plugin-factory.ts
@@ -17,6 +17,7 @@ import {
   PluginType,
   RepositoryConfig,
   SentenceCasePluginConfig,
+  GroupPriorityPluginConfig,
 } from '../manifest';
 import {GitHub} from '../github';
 import {ManifestPlugin} from '../plugin';
@@ -27,6 +28,7 @@ import {VersioningStrategyType} from './versioning-strategy-factory';
 import {MavenWorkspace} from '../plugins/maven-workspace';
 import {ConfigurationError} from '../errors';
 import {SentenceCase} from '../plugins/sentence-case';
+import {GroupPriority} from '../plugins/group-priority';
 
 export interface PluginFactoryOptions {
   type: PluginType;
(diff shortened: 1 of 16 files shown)


src/factories/plugin-factory.ts
```diff
@@ -81,6 +83,13 @@ const pluginFactories: Record<string, PluginBuilder> = {
       options.repositoryConfig,
       (options.type as SentenceCasePluginConfig).specialWords
     ),
+  'group-priority': options =>
```
bcoe, line 86: Eventual cleanup, it would be nice if we could figure out how t instantiate plugins without needing to enumerate them all in this lookup. Perhaps we could introduce a generated `index.ts` to the plugins folder, which would walk the plugins folder during a build step.

src/manifest.ts
```diff
@@ -198,9 +198,13 @@ export interface SentenceCasePluginConfig extends ConfigurablePluginType {
 export interface WorkspacePluginConfig extends ConfigurablePluginType {
   merge?: boolean;
 }
+export interface GroupPriorityPluginConfig extends ConfigurablePluginType {
+  groups: string[];
+}
 export type PluginType =
   | DirectPluginType
   | ConfigurablePluginType
+  | GroupPriorityPluginConfig
```
bcoe, line 207: In general, I think it would be worth thinking about an eventual refactor that allows us to add additional plugins, options, etc., updating less lists of types, factories, etc.

On the positive, I like that the manifest + plugin based approach requires less updates to add new features, than adding new command line flags.

src/plugins/group-priority.ts
```diff
@@ -0,0 +1,93 @@
...
+export class GroupPriority extends ManifestPlugin {
+  readonly groups: string[];
+  constructor(
+    github: GitHub,
+    targetBranch: string,
+    repositoryConfig: RepositoryConfig,
+    groups: string[]
+  ) {
+    super(github, targetBranch, repositoryConfig);
+    this.groups = groups;
```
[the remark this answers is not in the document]
chingor13: Added a jsdoc block to the constructor

chingor13:
> I think we should add some documentation that describes the plugin.

Added a description/example to the markdown docs with the existing plugin documentation.
Fixed the description in the config schema.
Added jsdoc to the plugin constructor.

bcoe (approved):

PR #1665, opened 2022-09-29
feat: add initial version to base strategy

### LINKED ISSUE #1666: Support having an alternate initial version
Node often needs to release libraries less than 1.0.0. We can do so by adding a commit override message, but it gets tiring since it needs to be done everytime. It would be better if we could set this in the config options for release-please so we don't need to add the override each time.

sofisl:
Fixes #1666

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -41,6 +41,7 @@ import {
   FileNotFoundError,
   ConfigurationError,
 } from './errors';
+import path from 'path';
 
 type ExtraJsonFile = {
   type: 'json';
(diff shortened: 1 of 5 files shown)


src/manifest.ts
```diff
@@ -98,6 +98,7 @@ export interface ReleaserConfig {
   separatePullRequests?: boolean;
   labels?: string[];
   releaseLabels?: string[];
+  alternateInitialVersion?: string;
```
chingor13, line 101: Let's name this `initialVersion` here and elsewhere.

chingor13 (changes requested):
You'll also need to populate the field from config. The 2 entrypoints to the code are `Manifest.fromConfig()` (configuration options provided explicitly) and `Manifest.fromManifest()` (configuration loaded from `release-please-config.json` in the repository).

You can see an example of parsing the configuration option in: #1555

chingor13 (approved):

PR #1667, opened 2022-09-29
feat(maven-workspace): update all discovered pom.xml within the components

### LINKED ISSUE #1664: Release-please to have the maven-workspace consider all pom.xml artifacts for cross project version bumping
Currently, annotation based version updates do not happen cross project/module. 
For https://github.com/googleapis/google-cloud-java/tree/main/google-cloud-gapic-bom, the pom.xml needs to be updated with versions mentioned outside of its directory.
As discussed, this will require changes to maven-workspace plugin configuration.

chingor13:
Maven projects can have multiple artifacts within each component. When using the `maven-workspace` plugin, we would like to be able to version bump any of the artifacts in a released component.

Consider a repository with 3 components:
1. `bom`
2. `lib1`
3. `lib2`

The `bom` project is a bill of materials project that references the versions from libraries contained in `lib1` and `lib2`.

| pom file | artifact name | depends on |
| --- | --- | --- |
| bom/pom.xml | com.google.example:my-bom | com.google.example:lib1-bom, com.google.example:lib2-bom |
| lib1/pom.xml | com.google.example:lib1-parent | |
| lib1/bom/pom.xml | com.google.example:lib1-bom | |
| lib1/artifact1/pom.xml | com.google.example:lib1-artifact1 | |
| lib2/pom.xml | com.google.example:lib2-parent | |
| lib2/bom/pom.xml | com.google.example:lib2-bom | |
| lib2/artifact2/pom.xml | com.google.example:lib2-artifact2 | |

When `lib1` component is updated, we want to update the `bom` component. However, `bom` references `lib1-bom` and `lib1-artifact1` is the actual primary artifact name for that component.

With this PR, you can configure the `considerAllArtifacts` option for the `maven-workspace` plugin. If set, we actually look at every maven artifact declared in the repository and associate those artifacts with their component. It will update another component if any of those artifact versions has affected the other components.

Example config: https://github.com/chingor13/release-please-playground/blob/fd84c4a1c671812b9dbef736a00c083dc602b751/release-please-config.json#L1-L20
Example PR: https://github.com/chingor13/release-please-playground/pull/39

Fixes #1664

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "./build/src/index.js",
   "bin": "./build/src/bin/release-please.js",
   "scripts": {
-    "test": "cross-env ENVIRONMENT=test LC_ALL=en c8 mocha --recursive --timeout=5000 build/test",
+    "test": "cross-env ENVIRONMENT=test LC_ALL=en mocha --recursive --timeout=5000 build/test",
     "docs": "echo add docs tests",
     "test:snap": "SNAPSHOT_UPDATE=1 LC_ALL=en npm test",
     "clean": "gts clean",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "./build/src/index.js",
   "bin": "./build/src/bin/release-please.js",
   "scripts": {
-    "test": "cross-env ENVIRONMENT=test LC_ALL=en mocha --recursive --timeout=5000 build/test",
+    "test": "cross-env ENVIRONMENT=test LC_ALL=en c8 mocha --recursive --timeout=5000 build/test",
     "docs": "echo add docs tests",
     "test:snap": "SNAPSHOT_UPDATE=1 LC_ALL=en npm test",
     "clean": "gts clean",
--- a/src/plugins/maven-workspace.ts
+++ b/src/plugins/maven-workspace.ts
@@ -310,7 +310,7 @@ export class MavenWorkspace extends WorkspacePlugin<MavenArtifact> {
 
   protected bumpVersion(artifact: MavenArtifact): Version {
     const strategy = new JavaSnapshot(new AlwaysBumpPatch());
-    return strategy.bump(Version.parse(artifact.version), [fakeCommit]);
+    return strategy.bump(Version.parse(artifact.version), [FAKE_COMMIT]);
   }
 
   protected updateCandidate(
@@ -561,7 +561,11 @@ function parseMavenArtifact(
     pomContent,
   };
 }
-const fakeCommit: ConventionalCommit = {
+
+// We use a fake commit to leverage the Java versioning strategy
+// (it should be a patch version bump and potentially remove the
+// -SNAPSHOT portion of the version)
+const FAKE_COMMIT: ConventionalCommit = {
   message: 'fix: fake fix',
   type: 'fix',
   scope: null,
--- a/src/plugins/maven-workspace.ts
+++ b/src/plugins/maven-workspace.ts
@@ -79,7 +79,7 @@ export class MavenWorkspace extends WorkspacePlugin<MavenArtifact> {
     options: MavenWorkspacePluginOptions = {}
   ) {
     super(github, targetBranch, repositoryConfig, options);
-    this.considerAllArtifacts = options.considerAllArtifacts ?? false;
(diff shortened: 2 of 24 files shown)


src/plugins/maven-workspace.ts
```diff
@@ -57,72 +66,26 @@ const XPATH_PROJECT_VERSION =
...
-      name: `${groupId}:${artifactId}`,
-      version: versionNodes[0].firstChild!.textContent!,
-      dependencies,
-      testDependencies,
-      pomContent: content.parsedContent,
-    };
+    return parseMavenArtifact(content.parsedContent, path, this.logger);
   }
 
   protected async buildAllPackages(
```
bcoe, line 91: I don't find that the method name `buildAllPackages` describes well what's happening, which if I understand correctly is we're essentially building a graph of dependencies that need to be updated.

It would be helpful for me if the graph building logic was pulled into a helper, called something like `buildUpdateGraph`, and if this method had a detailed comment that explained the behavior. For future generations working in the codebase, with some of these more complex algorithms, I think it would even be useful to provide an ascii graph (with something like https://asciiflow.com) which shows how the dependency graph is built.

bcoe (commented):
Left some initial suggestions, but haven't read too deeply yet.

I think it would be good to get a stakeholder on the Java team to review too, perhaps @suztomo?

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "./build/src/index.js",
   "bin": "./build/src/bin/release-please.js",
   "scripts": {
-    "test": "cross-env ENVIRONMENT=test LC_ALL=en c8 mocha --recursive --timeout=5000 build/test",
+    "test": "cross-env ENVIRONMENT=test LC_ALL=en mocha --recursive --timeout=5000 build/test",
```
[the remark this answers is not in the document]
chingor13: This one is unintentional. I was running the tests dozens of times and the coverage report was intrusive to my local testing.

src/plugins/maven-workspace.ts
chingor13: This is an abstract method declared by the base `workspace` plugin. The idea here is to scan the entire repository for packages that could interact with each other. In this case, the template type `T` is a `MavenArtifact`. The method is implemented by the per-language workspace plugins and is documented [here](https://github.com/googleapis/release-please/blob/main/src/plugins/workspace.ts#L224-L229).

This method does not build a graph - shared code in the base `workspace` plugin is building the graph of all the artifacts detected from this method.

suztomo: Adding a line 

```
"considerAllArtifacts": true
```

is fine to me.

suztomo: (Not a big deal) Are you sure it's good idea to have "considerAllArtifacts" and "type" at the same level? I felt natural to have "configuration" field to hold an object like below:

```
  "plugins": [
    {
      "type": "maven-workspace",
      "configuration" : {
        "considerAllArtifacts": true
      }
    }
  ]
```

chingor13: > However, I also feel that I don't think of a situation where I want to set the parameter to `false`

After further consideration, I could see us defaulting this parameter to true. It could cause expected pom.xml files to be touched if they exist within a directory that is managed by release-please, but in those cases, it could be configured to only consider the root pom.xml for each configured component.

> I felt natural to have "configuration" field to hold an object like below

This is probably a better design, but all the other configurable plugins have their options at the same level as `type`. For consistency, we should probably keep them at the same level.

suztomo: Thank you for explanation.

chingor13: @suztomo Switched this option to default to `true`. You can now opt-out of this behavior by setting the `considerAllArtifacts` to `false`.

suztomo:
(This question I should have asked earlier)

I remember you introduced the maven-workspace for Java monorepo (and java-cloud-bom repo) to express cross-component dependencies. A BOM component uses the versions of different components. For me, it's Maven's pom files. But how about others? For others, it could be other text-based files that have the Release Please annotations for the versions of different components.

Is there anything special about Maven for "maven-workspace" plugin?

suztomo (approved):

suztomo:
That's nice design. Thank you for explanation.

PR #1669, opened 2022-09-30
feat: add ability to set extra labels on components

### LINKED ISSUE #1624: `label` and `release-label` for each packages are ignored
I configured `label` and `release-label` for each packages, but it is not reflected in to release PR.

I think it is a bug because [the JSON Schema of the config](https://github.com/googleapis/release-please/blob/main/schemas/config.json) allows `label` or `release-label` inside packages.
However, I'm sorry if it is correct behavior.

An example of my `release-please-config.json` follows:
```json
{
  "separate-pull-requests": true,
  "packages": {
    "packages/a": {
      "label": "autorelease: pending,Package-A",
      "release-label": "autorelease: tagged,Package-A",
    },
    "packages/b": {
      "label": "autorelease: pending,Package-B",
      "release-label": "autorelease: tagged,Package-B",
    }
  }
}
```

#### Environment details

  - OS: macOS
  - Node.js version: v16.11.0
  - npm version: 8.0.0
  - `release-please` version: 13.21.0

#### Steps to reproduce

  1. configure release-please-config.json with `label` or `release-label` properties in `packages`
  2. run `release-please release-pr` to create a release PR

=> The created release PR does not have the labels specified in the config.

#### Root cause

The ReleaserConfig has `labels` and `releaseLabels` properties on the type definition,

https://github.com/googleapis/release-please/blob/3169b4331ddad030a528cfc4003b5fef04ca8588/src/manifest.ts#L95

but `mergeReleaserConfig()` doesn't return then.

https://github.com/googleapis/release-please/blob/3169b4331ddad030a528cfc4003b5fef04ca8588/src/manifest.ts#L1508

chingor13:
We are adding a new `extra-label` option which is non-load bearing (we use the existing `label` and `release-label` to indentify which PRs are release-please's and what state they are in).
This new option allows maintainers to organize their components (e.g. labeling with and API component label or perhaps a language)

Fixes #1624

BEGIN_NESTED_COMMIT
fix: fix schema definition for label and release-label

These fields are not actually configurable per-component.
END_NESTED_COMMIT

### DIFF
--- a/schemas/config.json
+++ b/schemas/config.json
@@ -66,12 +66,8 @@
           "description": "Open the release pull request in draft mode. Defaults to `false`.",
           "type": "boolean"
         },
-        "label": {
-          "description": "Comma-separated list of labels to add to newly opened pull request",
-          "type": "string"
-        },
-        "release-label": {
-          "description": "Comma-separated list of labels to add to a pull request that has been released/tagged",
+        "extra-label": {
+          "description": "Comma-separated list of labels to add to a newly opened pull request",
           "type": "string"
         },
         "include-component-in-tag": {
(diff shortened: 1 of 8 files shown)


sofisl (approved):
NB, but it would be helpful to understand a better distinction between labels and extra labels (especially since they are both arrays, so it's not clear where I'd put one over another)

PR #1691, opened 2022-10-11
fix: add initial-version to allowed properties in manifest schema

sofisl:


schemas/config.json
```diff
@@ -393,6 +393,7 @@
     "tag-separator": true,
     "extra-files": true,
     "version-file": true,
-    "snapshot-label": true
+    "snapshot-label": true,
+    "initial-version": true,
```
chingor13, line 397:
```suggestion
    "initial-version": true
```

chingor13 (changes requested):
Trailing comma is invalid in JSON

chingor13 (approved):

Issue #1706, opened 2022-10-18
FR(python): add support for updating *version.py


parthea:
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

 **Is your feature request related to a problem? Please describe.**
Yes, we are working on migrating the version out of `setup.py` to another file. Some python client libraries already have `version.py` (for example, [this file](https://github.com/googleapis/python-dialogflow-cx/blob/main/google/cloud/dialogflowcx_v3/types/version.py)), so we would like to store the python version in `gapic_version.py` instead.

 **Describe the solution you'd like**
Expand support for `release-please` to update `*version.py` instead of `version.py`

 **Describe alternatives you've considered**
I considered using `version.py` which is already supported by release please however this would conflict with another version file (for example, [this file](https://github.com/googleapis/python-dialogflow-cx/blob/main/google/cloud/dialogflowcx_v3/types/version.py)) which already exists in the python client library.

 **Additional context**

parthea:
After offline discussion with @chingor13 , I'm going to explore using the `extra-files` option

PR #1707, opened 2022-10-18
feat: expand support for python versions to include path *version.py

parthea:
Fixes https://github.com/googleapis/release-please/issues/1706 🦕

### DIFF
--- a/src/strategies/python.ts
+++ b/src/strategies/python.ts
@@ -116,9 +116,9 @@ export class Python extends BaseStrategy {
         );
     }
 
-    // There should be only one version.py, but foreach in case that is incorrect
-    const versionPyFilesSearch = this.github.findFilesByFilenameAndRef(
-      'version.py',
+    // There should be only one *version.py, but foreach in case that is incorrect
+    const versionPyFilesSearch = this.github.findFilesByGlobAndRef(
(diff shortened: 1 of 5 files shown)


src/strategies/python.ts
```diff
@@ -116,9 +116,9 @@ export class Python extends BaseStrategy {
         );
     }
 
-    // There should be only one version.py, but foreach in case that is incorrect
-    const versionPyFilesSearch = this.github.findFilesByFilenameAndRef(
-      'version.py',
+    // There should be only one *version.py, but foreach in case that is incorrect
+    const versionPyFilesSearch = this.github.findFilesByGlobAndRef(
+      '**/*version.py',
```
chingor13, line 121: I feel like this would work for your use case, but could also find unexpected other files to replace. Could there possibly be another submodule that has a version.py file that we wouldn't want to touch?

chingor13 (commented):
An alternate implementation would be to look at the `extra-files` option and apply the python specific updater to those files. This would mean extra configuration for the repository, but it feels like an exceptional case.

parthea:
After offline discussion with @chingor13 , I'm going to explore using the `extra-files` option

PR #1719, opened 2022-10-24
feat: extract and return the created release id

### LINKED ISSUE #1645: Expose the release's unique id in `CreatedRelease`

Setting the existing `draft` flag to `true` will create a "draft" Github release. Presumably, the user wishes to eventually turn that release into a non-draft release. To do so, a call to [`update a release`](https://docs.github.com/en/rest/releases/releases#update-a-release) will be necessary. Such a call requires the release's unique ID as a parameter which is not made available by `release-please`.

Specifically, this prevents `release-please-action` from having a `release_id` output parameter (for example) which could then be used in future steps to "publish" the draft release.

Expose the release's unique id as a public member of `CreatedRelease`.

An workaround is to lookup the [release by tag name](https://docs.github.com/en/rest/releases/releases#get-a-release-by-tag-name). Unfortunately, from empirical evidence, `draft` releases return a `404` response for calls to this REST endpoint. A workaround to that is to use the GraphQL interface which does not "hide" `draft` releases when querying by tag, i.e.:

```graphql
gh api graphql -f query='query {
  repository(name: "some-repo", owner: "some-owner") {
    release(tagName: "v0.0.1") {
      databaseId
    }
  }
}'
{
  "data": {
    "repository": {
      "release": {
        "databaseId": 1234567
      }
    }
  }
}
```

Using the [list releases](https://docs.github.com/en/rest/releases/releases#list-releases) REST endpoint would also work, but it becomes tedious (and error prone) to filter out by tag name.

 **Additional context**

The use case is as follows:

* use `release-please` to create a `draft` release (through `release-please-action`)
* actually publish the repository's artifacts (i.e.: docker or otherwise)
* when successful, publish the release (convert it to non-draft)

plaflamme:
- [x] ~Appropriate docs were updated (if necessary)~ N/A

Fixes #1645 🦕

NOTE: the code coverage seems to decrease, but I'm not familiar enough with the testing infrastructure to know how to address that. Any guidance would be appreciated.

### DIFF
--- a/src/github.ts
+++ b/src/github.ts
@@ -174,6 +174,7 @@ export interface ReleaseOptions {
 }
 
 export interface GitHubRelease {
+  id: number;
   name?: string;
   tagName: string;
   sha: string;
@@ -1362,6 +1363,7 @@ export class GitHub {
         target_commitish: release.sha,
       });
       return {
+        id: resp.data.id,
         name: resp.data.name || undefined,
         tagName: resp.data.tag_name,
         sha: resp.data.target_commitish,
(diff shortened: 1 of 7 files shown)


chingor13 (approved):
Thanks, LGTM

PR #1723, opened 2022-10-25
feat(node-workspace): maintain current version range prefix

OrKoN:

### DIFF
--- a/src/plugins/node-workspace.ts
+++ b/src/plugins/node-workspace.ts
@@ -346,8 +346,8 @@ export class NodeWorkspace extends WorkspacePlugin<Package> {
 
   private detectRangePrefix(version: string): string {
     return (
-      ['^', '~'].find(
-        supportedRangePrefix => version[0] === supportedRangePrefix
+      ['^', '~', '>', '<', '<=', '>='].find(supportedRangePrefix =>
+        version.startsWith(supportedRangePrefix)
       ) || ''
     );
   }
--- a/src/plugins/node-workspace.ts
+++ b/src/plugins/node-workspace.ts
@@ -167,13 +167,17 @@ export class NodeWorkspace extends WorkspacePlugin<Package> {
     for (const [depName, resolved] of graphPackage.localDependencies) {
       const depVersion = updatedVersions.get(depName);
       if (depVersion && resolved.type !== 'directory') {
+        const currentVersion = this.combineDeps(pkg)?.[depName];
+        const prefix = currentVersion?.length
+          ? this.detectRangePrefix(currentVersion)
+          : '';
         updatedPackage.updateLocalDependency(
           resolved,
           depVersion.toString(),
-          '^'
+          prefix
         );
         this.logger.info(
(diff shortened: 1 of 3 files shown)


OrKoN:
Hey, I have seen https://github.com/googleapis/release-please/pull/1580 and saw a suggestion to maintain the current range prefix when doing a release. That would work for us in the Puppeteer project (currently, we rely on a custom script to pin dependencies that does not update the changelog). Let me know if you think this approach would make sense and I will look into improving test coverage (hints are welcome).

src/plugins/node-workspace.ts
```diff
@@ -343,6 +343,22 @@ export class NodeWorkspace extends WorkspacePlugin<Package> {
   protected pathFromPackage(pkg: Package): string {
     return pkg.location;
   }
+
+  private detectRangePrefix(version: string): string {
+    return (
+      ['^', '~'].find(
```
danez, line 349: Should we add `>`, `>=`, `<`, and `<=` too?

danez (changes requested):
I was just about to start looking into #1580 again and noticed this PR 👍

You can copy the[ tests from](https://github.com/googleapis/release-please/pull/1580/files#diff-2d241afd82e6bbd092d7b04dceadd3cf072f7731340b0a64988ee7c004b26745) #1580 and modify them slightly.
The current tests should fail anyway and also need probably fixing.

src/plugins/node-workspace.ts
```diff
@@ -343,6 +343,22 @@ export class NodeWorkspace extends WorkspacePlugin<Package> {
   protected pathFromPackage(pkg: Package): string {
     return pkg.location;
   }
+
+  private detectRangePrefix(version: string): string {
```
chingor13, line 347: non-blocking: we could consider defining an enum of possible prefix types. This could also be done in a cleanup at a later time as these are private interfaces.

chingor13 (changes requested):

test/plugins/node-workspace.ts
```diff
@@ -143,6 +143,42 @@ describe('NodeWorkspace plugin', () => {
...
+        candidate => candidate.config.releaseType === 'node'
+      );
+      expect(nodeCandidate).to.not.be.undefined;
+      const updates = nodeCandidate!.pullRequest.updates;
+      assertHasUpdate(updates, 'plugin1/package.json');
+      assertHasUpdate(updates, 'node1/package.json');
+
+      expect(updates[1].updater).instanceof(RawContent);
+      const updater = updates[1].updater as RawContent;
+      snapshot(updater.rawContent);
```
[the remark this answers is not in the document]
OrKoN: Updated!

src/plugins/node-workspace.ts
OrKoN: Added an enum!

chingor13 (approved):

Daniel950502 (commented):
#1749

PR #1730, opened 2022-10-29
fix(cargo-workspace): Update target dependencies in Cargo workspace packages

### LINKED ISSUE #1729: Cargo workspace plug-in does not update target-specific workspace dependencies

This is a client library issue (assuming this project is the client library).

No, I don't think so (I wrote a failing test case for the project).

No

#### Environment details

  - OS: Linux
  - Node.js version: v18.9.0
  - npm version: v8.9.1
  - `release-please` version: v14.15.1

#### Steps to reproduce

  1. Set up release-please on a Rust Cargo workspace project
  2. Express a target-specific dependency relationship across two packages in the workspace
  3. Push a change that would cause the depended-on package to rev its version
  4. Review the release PR; the dependent package won't be updated with a new version, and its dependency entry of the dependend-on package won't have its version bumped to the latest version

Thanks!

cdata:
Hello, I wrote a fix while tracking down #1729 

I'm happy to discuss an alternative approach, but since I already wrote this fix I figured I would put it up for consideration.

TL;DR of the issue is that a Cargo manifest can have compilation target-scoped dependencies, and these must be taken into consideration in addition to the top-level set of dependencies.

Fixes #1729 :crab:

### DIFF
--- a/__snapshots__/cargo-workspace.js
+++ b/__snapshots__/cargo-workspace.js
@@ -27,6 +27,15 @@ Release notes for path: packages/rustA, releaseType: rust
     * pkgB bumped from 2.2.2 to 2.2.3
 </details>
 
+<details><summary>pkgE: 3.3.4</summary>
+
+### Dependencies
+
+* The following workspace dependencies were updated
+  * dependencies
+    * pkgA bumped from 1.1.1 to 1.1.2
+</details>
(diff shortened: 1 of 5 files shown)


chingor13 (approved):
Thanks!

PR #1756, opened 2022-11-21
fix(dart): throw MissingRequiredFileError if pubspec.yaml is missing

chingor13:
Towards https://github.com/googleapis/repo-automation-bots/issues/4664

Throwing this error will allow the downstream automation to handle this FileNotFoiundError better (indicating that it's an repository configuration error).

fix(helm): throw MissingRequiredFileError if Chart.yaml is missing


bcoe (approved):
I would update the commit message to mention the helm files as well.

PR #1762, opened 2022-11-26
fix: handle incrementing ruby patch versions larger than 1 digit

### LINKED ISSUE #1761: Support longer patch numbers in ruby version.rb

When creating a release for a ruby project, it can't handle incrementing patches higher than 9.

I want to be able to increment patches higher than 9. For example: a `version.rb` file with `1.2.10` should increment to `1.2.11`. It looks like the regex doesn't currently handle this.

**Additional context**

This is basically piggy-backing off of https://github.com/googleapis/release-please/issues/1018 to allow for higher patch versions. I also took a look at the python version updater and it seems to already handle this scenario.

invalidusrname:
This fixes the linked issue by applying a small fix to updaters/version-rb.ts.

Fixes #1761 🦕

### DIFF
--- a/__snapshots__/version-rb.js
+++ b/__snapshots__/version-rb.js
@@ -24,6 +24,31 @@ end
 
 `
 
+exports['version.rb updateContent updates long patch versions in version.rb 1'] = `
+# Copyright 2019 Google LLC
+#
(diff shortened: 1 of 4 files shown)


chingor13 (approved):
Thanks!

PR #1775, opened 2022-12-12
feat!: strategies can parse multiple releases from single release PR

chingor13:
This will give strategies an option to create multiple tags from a single PR. For example, `java-yoshi` for a monorepo could configure itself to be a single component (create a single release PR), but then create several releases/tags.


bcoe (changes requested):
I think a test would help me better understand how this is used.

What would configuration look like in a repo that takes advantage of this functionality (perhaps an update to reference docs would help too).

bcoe (approved):

PR #1783, opened 2022-12-15
test: add failing test for valid cargo circular dependency in workspace

### LINKED ISSUE #1662: `cargo-workspace` plugin cannot handle dependency cycles in dev-dependencies

~**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.~

~This is the client library for . We will only be able to assist with issues that pertain to the behaviors of this library. If the issue you're experiencing is due to the behavior of the product itself, please visit the [ Support page]() to reach the most relevant engineers.~

It is a library issue.

  - Search the issues already opened: https://github.com/googleapis/release-please/issues: **Couldn't find anything.**
  - Search the issues on our "catch-all" repository: https://github.com/googleapis/google-cloud-node: **Couldn't find anything.**
  - Search or ask on StackOverflow (engineers monitor these tags): http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js: **I don't think this applies here.**

~Please create an issue in the [support console](https://cloud.google.com/support/) to ensure a timely response.~

If the support paths suggested above still do not result in a resolution, please provide the following details.

#### Environment details

  - OS: Manjaro Linux
  - Node.js version: 18.8.0
  - npm version: 8.19.1
  - `release-please` version: 14.7.2

#### Steps to reproduce

  1. Have a cargo workspace that has dependency cycles with _dev_ dependencies.
  2. Try to use the cargo-workspace plugin of release-please.

Cargo supports dependency cycles in the form of A -dep-> B -dev-dep-> A. In other words, I can have a crate A that depends on B but B can depend on A again in its `dev-dependencies`. `release-please` doesn't seem to allow this which I think is an unnecessary restriction. Version bumping of `dev-dependencies` should not impact the version of a package so perhaps `dev-dependencies` should just be ignored here: https://github.com/googleapis/release-please/blob/main/src/updaters/rust/common.ts#L71

Thanks!

chingor13:
Fixes #1662


Mansion-dotcom (commented):
Approved

PR #1793, opened 2023-01-04
feat: add changelog.json updater

bcoe:
feat: introduce java-yoshi-mono-repo release strategy

Introduces JavaYoshiMonoRepos strategy, which includes a machine readable changelog.json.


src/strategies/base.ts
```diff
@@ -46,6 +46,7 @@ const DEFAULT_CHANGELOG_PATH = 'CHANGELOG.md';
 
 export interface BuildUpdatesOptions {
   changelogEntry: string;
+  commits: ConventionalCommit[];
```
chingor13, line 49: This may be a breaking interface change - strategies are a public class that can be extended. Make it optional?

src/updaters/changelog-json.ts
```diff
@@ -0,0 +1,103 @@
...
+      // Including chores in changelog.json would cause the file to grow
+      // too quickly, as it would include the chore of creating the
+      // previous release itself (so all paths would always be updated).
+      if (commit.type === 'chore' && !change.breakingChangeNote) continue;
+      changes.push(change);
+    }
+    // If all commits were ignored, simply return the original changelog.json.
+    if (changes.length === 0) {
+      return JSON.stringify(parsed, null, 2);
+    }
```
chingor13, line 89: Can probably just return `content` unless you are wanting to force the formatting on the original file.

src/updaters/changelog-json.ts
```diff
@@ -0,0 +1,103 @@
...
+interface Change {
+  type: string;
+  scope?: string;
+  sha: string;
+  message: string;
+  breakingChangeNote?: string;
+}
+
+/**
+ *
```
chingor13, line 38: nit: fill this out

test/helpers.ts
```diff
@@ -271,6 +271,13 @@ export function assertHasUpdate(
   return found!;
 }
 
+export function getUpdate(updates: Update[], path: string): Update {
```
chingor13, line 274: `assertHasUpdate` returns the found update as well. All the places you are using `getUpdate`, you are using `assertHasUpdate` immediately before.

chingor13 (changes requested):
For the most part, this LGTM.

To make the review diff easier, can you split into 2 PRs (they are very loosely coupled)?

src/updaters/changelog-json.ts
bcoe: done.

PR #1808, opened 2023-01-18
feat: introduce ChangelogJson updater

bcoe:
Introduces changelog.json updater.


src/updaters/changelog-json.ts
```diff
@@ -0,0 +1,104 @@
...
+      for (const note of commit.notes) {
+        if (note.title === BREAKING_CHANGE_TITLE) {
+          change.breakingChangeNote = note.text;
+        }
+      }
+
+      // Including chores in changelog.json would cause the file to grow
+      // too quickly, as it would include the chore of creating the
+      // previous release itself (so all paths would always be updated).
+      if (commit.type === 'chore' && !change.breakingChangeNote) continue;
```
chingor13, line 84: I don't think this is the updater's job to determine whether to include these or not -- this feels like something the strategy should be deciding/customizing.

test/fixtures/strategies/java-yoshi/artifact-map.json
```diff
@@ -0,0 +1,3 @@
+{
```
chingor13, line 1: This file belongs in the follow-up PR that adds the Java implementation

test/fixtures/strategies/java-yoshi/changelog.json
```diff
@@ -0,0 +1,3 @@
+{
```
chingor13, line 1: This file belongs in the follow-up PR that adds the Java implementation

__snapshots__/java-yoshi-mono-repo.js
```diff
@@ -0,0 +1,26 @@
+exports['JavaYoshi buildUpdates does not update changelog.json if no artifacts matched in artifact-map.json 1'] = `
```
chingor13, line 1: This file belongs in the follow-up PR that adds the Java implementation

chingor13 (changes requested):

test/updaters/changelog-json.ts
```diff
@@ -0,0 +1,116 @@
...
+import {parseConventionalCommits} from '../../src/commit';
+import {buildMockCommit} from '../helpers';
+
+const UUID_REGEX =
+  /[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}/g;
+const ISO_DATE_REGEX =
+  /[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\.[0-9]+Z/g; // 2023-01-05T16:42:33.446Z
+
+describe('changelog.json', () => {
+  it('appends new release to empty changelog', async () => {
```
chingor13, line 28:
```suggestion
  it('prepends new release to empty changelog', async () => {
```

test/updaters/changelog-json.ts
```diff
@@ -0,0 +1,116 @@
...
+    });
+    const newContent = changelogJson.updateContent(oldContent);
+    snapshot(
+      newContent
+        .replace(/\r\n/g, '\n') // make newline consistent regardless of OS.
+        .replace(UUID_REGEX, 'abc-123-efd-qwerty')
+        .replace(ISO_DATE_REGEX, '2023-01-05T16:42:33.446Z')
+    );
+  });
+  it('adds latest release to front of list', async () => {
```
chingor13, line 50:
```suggestion
  it('prepends latest release to existing changelog', async () => {
```

chingor13 (changes requested):

src/updaters/changelog-json.ts
bcoe: It feels like significantly less toil to do it here, since we'd instead have to do it in every single strategy that has changelog.json support.

chingor13: I could see having a shared function that could filter commits by type - something like:

```ts
function filterCommits(commits: ConventionalCommit[], allowedTypes: string[], includeAnyBreakingChange: boolean = true): ConventionalCommit[] {}
```

Another option could be to include the filter options in the constructor for this updater. But having a hard-coded business decision in this class doesn't feel like the right place for that configuration.

bcoe: sounds reasonable, I'll just start with the logic in the Java Mono Repo strategy for now.

chingor13 (approved):

PR #1810, opened 2023-01-18
feat: introduce JavaYoshiMonoRepo strategy

bcoe:
This splits out the MonoRepo strategy, used to create a single release artifact for the googleapis/google-cloud-java repo, from the JavaYoshi strategy used for the split repo case.

This allows logic to be introduced that remaps the combined SDK release back to individual release entries for each artifact, which matches the information architecture on cloud.google.com (_which has release notes split up by product, see: [example](https://cloud.google.com/bigquery/docs/release-notes)_).


src/strategies/java-yoshi-mono-repo.ts
```diff
@@ -0,0 +1,305 @@
...
+          version,
+          changelogEntry: options.changelogEntry,
+        }),
+      });
+
+      // The artifact map maps from directory paths in repo to artifact names on
+      // Maven, e.g, java-secretmanager to com.google.cloud/google-cloud-secretmanager.
+      const artifactMap = await this.getArtifactMap('artifact-map.json');
+      if (artifactMap) {
+        const changelogUpdates: Array<Updater> = [];
```
bcoe, line 192: TODO: before merge, filter out `chore:` commit types.

bcoe:
CC: @suztomo, perhaps this explains the design better. @chingor13 requested that I break out the separate `JavaYoshiMonoRepo` class, so that we can support the alternate logic required by the googleapis/google-cloud-java mono-repo.

`JavaYoshi` would then implement a simpler implementation, which does not require splitting commits, adding support for a `changelog.json` to split repos.

suztomo:
So JavaYoshiMonoRepo is just special and other strategies do not need to be renamed to accomodate changelog.json, right? That's good, then.

test/strategies/java-yoshi-mono-repo.ts
```diff
@@ -373,6 +373,66 @@ describe('JavaYoshiMonoRepo', () => {
       );
     });
 
+    it('ommits non-breaking chores from changelog.json', async () => {
```
chingor13, line 376: nit: ommits -> omits

src/strategies/java-yoshi-mono-repo.ts
```diff
@@ -201,7 +203,15 @@ export class JavaYoshiMonoRepo extends Java {
...
-                commits: splitCommits[path],
+                // We filter out "chore:" commits, to reduce noise in the upstream
+                // release notes. We will only show a product release note entry
+                // if there has been a substantial change, such as a fix or feature.
+                commits: splitCommits[path].filter(commit => {
+                  const isBreaking = commit.notes.find(note => {
+                    return note.title === BREAKING_CHANGE_NOTE;
+                  });
+                  return commit.type !== 'chore' || isBreaking;
+                }),
```
chingor13, line 214: nit: consider filtering prior to passing to commit split to avoid filtering the same commit multiple times if it touches a bunch of paths.

chingor13 (approved):

test/strategies/java-yoshi-mono-repo.ts
bcoe: done.

src/strategies/java-yoshi-mono-repo.ts
bcoe: done.

PR #1811, opened 2023-01-18
feat(cli): add ability to dynamically load release-please plugin

chingor13:
feat: export Strategy interface so external packages can implement

This should allow you to publish private configs in your own npm package and load via the CLI (you can already load configs programatically via `registerReleaseType` etc).

To make an external package dynamically loadable via the release-please CLI, export an `init()` method from your main file that registers your release-please extensions.

### DIFF
--- a/src/bin/release-please.ts
+++ b/src/bin/release-please.ts
@@ -833,7 +833,9 @@ export const parser = yargs
         if (plugin?.init) {
           console.log(`loading plugin: ${pluginName}`);
         } else {
-          console.warn(`plugin: ${pluginName} did not have an init() function.`);
+          console.warn(
+            `plugin: ${pluginName} did not have an init() function.`
+          );
         }
       } catch (e) {
         console.warn(`failed to require plugin: ${pluginName}:`, e);
--- a/src/bin/release-please.ts
+++ b/src/bin/release-please.ts
@@ -828,11 +828,15 @@ export const parser = yargs
   .middleware(argv => {
     for (const pluginName of argv.plugin) {
       console.log(`requiring plugin: ${pluginName}`);
-      const plugin = require(pluginName.toString());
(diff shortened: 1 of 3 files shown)


src/bin/release-please.ts
```diff
@@ -820,6 +820,22 @@ export const parser = yargs
...
   })
+  .option('plugin', {
+    describe: 'load plugin named release-please-<plugin-name>',
+    type: 'array',
+    default: [],
+  })
+  .middleware(argv => {
+    for (const pluginName of argv.plugin) {
+      console.log(`requiring plugin: ${pluginName}`);
+      const plugin = require(pluginName.toString());
```
bcoe, line 831: You might want to do this in a `try`/`catch` and provide a nice error message.

bcoe (commented):
This approach seems reasonable. Keep in mind however, it won't work with the action, which I think some Googlers (the SQL connector folks?) are using.

bcoe:
> This approach seems reasonable. Keep in mind however, it won't work with the action, which I think some Googlers (the SQL connector folks?) are using.

A reasonable fix for this would be getting more people on the app if necessary, i.e., if they're using yoshi-specific strategies.

chingor13:
> > This approach seems reasonable. Keep in mind however, it won't work with the action, which I think some Googlers (the SQL connector folks?) are using.
> 
> A reasonable fix for this would be getting more people on the app if necessary, i.e., if they're using yoshi-specific strategies.

We can programmatically load the yoshi extensions in the action if we want. We won't be able to load arbitrary external plugins but that's probably ok.

src/bin/release-please.ts
chingor13: Done.

bcoe (approved):

PR #1829, opened 2023-01-26
refactor: rename Salesforce to sfdx

### LINKED ISSUE #1814: Implement Salesforce `sfdx-project.json` version updater
As a Salesforce developer building 2GP package
I want release please to be able to update the version in my [`sfdx-project.json` file](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_ws_config.htm)
So I can benefit from release please setting my package version

scolladon:

Fixes #1814 🦕

### DIFF
--- a/__snapshots__/cli.js
+++ b/__snapshots__/cli.js
@@ -42,8 +42,8 @@ Options:
    [choices: "dart", "dotnet-yoshi", "elixir", "expo", "go", "go-yoshi", "helm",
                   "java", "java-backport", "java-bom", "java-lts", "java-yoshi",
        "java-yoshi-mono-repo", "krm-blueprint", "maven", "node", "ocaml", "php",
-    "php-yoshi", "python", "ruby", "ruby-yoshi", "rust", "sfdx", "simple",
-                                                             "terraform-module"]
+      "php-yoshi", "python", "ruby", "ruby-yoshi", "rust", "salesforce", "sfdx", 
+                                                   "simple", "terraform-module"]
   --config-file                 where can the config file be found in the
                                 project? [default: "release-please-config.json"]
   --manifest-file               where can the manifest file be found in the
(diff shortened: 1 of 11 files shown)


scolladon:
Hi @chingor13 !

I don't know why the tests are failing.
It does not seems to be related to my PR.

Should I do something on my end ?

chingor13 (changes requested):
We can't really take this now, because it'd be a breaking change for users. It'd be possible to rename the class while keeping an alias for users' configuration that could specify `salesforce` or `sfdx` as the `release-type`.

scolladon:
> We can't really take this now, because it'd be a breaking change for users. It'd be possible to rename the class while keeping an alias for users' configuration that could specify `salesforce` or `sfdx` as the `release-type`.

I have added back the `salesforce` release-type mapper
But I have not added it back in the documentation
This way, previous user of the `salesforce` release type will be able to keep using it without breaking changes
And new comers will be able to use `sfdx` instead

__snapshots__/cli.js
```diff
@@ -42,7 +42,7 @@ Options:
    [choices: "dart", "dotnet-yoshi", "elixir", "expo", "go", "go-yoshi", "helm",
                   "java", "java-backport", "java-bom", "java-lts", "java-yoshi",
        "java-yoshi-mono-repo", "krm-blueprint", "maven", "node", "ocaml", "php",
-    "php-yoshi", "python", "ruby", "ruby-yoshi", "rust", "salesforce", "simple",
+    "php-yoshi", "python", "ruby", "ruby-yoshi", "rust", "sfdx", "simple",
```
chingor13, line 45: Please leave both `salesforce` and `sfdx` as options for the CLI in case those are being used

__snapshots__/cli.js
```diff
@@ -241,7 +241,7 @@ Options:
...
```
chingor13, line 244: Please leave both `salesforce` and `sfdx` as options for the CLI in case those are being used

chingor13 (changes requested):
Thanks, almost there.

Also, is there an issue driving the need for this rename?

scolladon:
> Thanks, almost there.
> 
> Also, is there an issue driving the need for this rename?

I have not created a new issue, I reused the first one

chingor13 (approved):

dschach:
This is a good PR. Thank you, @scolladon

PR #1839, opened 2023-02-01
feat(changelog.json): add pr suffix to issues array

bcoe:
In discussion with downstream implementers, we decideed that it would make it easier to customize the CHANGELOG generated if we pre-parsed the PR # suffix that GitHub adds to squashed commits.

### DIFF
--- a/src/updaters/changelog-json.ts
+++ b/src/updaters/changelog-json.ts
@@ -67,9 +67,6 @@ export class ChangelogJson extends DefaultUpdater {
     const changes = [];
     for (const commit of this.commits) {
       const issues = new Set<string>();
-      for (const ref of commit.references) {
-        issues.add(ref.issue);
-      }
       // The commit.message field contains the type/scope prefix.
       let message = commit.message.replace(COMMIT_PREFIX, '');
       // When squashing commits, GitHub adds a suffix refrencing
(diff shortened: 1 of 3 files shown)


bcoe:
@chingor13 in discussion with @martinvarelaj, he was hesitant to want to perform any processing on the commit message downstream and preferred if we placed all the referenced PRs and Issues in the issues array.

src/updaters/changelog-json.ts
```diff
@@ -65,13 +66,23 @@ export class ChangelogJson extends DefaultUpdater {
...
+      let message = commit.message.replace(COMMIT_PREFIX, '');
+      // When squashing commits, GitHub adds a suffix refrencing
+      // the # of the PR, e.g., chore(main): release 15.5.1 (#1838)
+      // this logic removes this suffix and prepends it to the
+      // issues array.
+      const match = message.match(PR_SUFFIX_REGEX);
+      if (match && match.groups?.pr) {
+        message = message.replace(match[0], '');
+        issues.unshift(match.groups.pr);
+      }
```
chingor13, line 80: Is it possible that we might need to de-dup the issue refs?

bcoe: Good question, probably worth doing so just in case?

bcoe: I've switched to using a `Set`, to eliminate duplicates.

chingor13 (approved):

PR #1841, opened 2023-02-03
feat(changelog.json): implement changelog.json for python

bcoe:
Add support for generating `changelog.json` to python strategy.

Can be seen in action [here](https://github.com/bcoe/google-cloud-python/pull/4).

### DIFF
--- a/src/strategies/python.ts
+++ b/src/strategies/python.ts
@@ -137,7 +137,7 @@ export class Python extends BaseStrategy {
     });
 
     // If a machine readable changelog.json exists update it:
-    const artifactName = projectName ?? await this.getArtifactName();
+    const artifactName = projectName ?? (await this.getNameFromSetupPy());
     if (options.commits && artifactName) {
       const commits = filterCommits(options.commits, this.changelogSections);
       updates.push({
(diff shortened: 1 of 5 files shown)


src/strategies/python.ts
```diff
@@ -148,6 +167,35 @@ export class Python extends BaseStrategy {
...
+    const ARTIFACT_NAME_REGEX = /name *= *['"](?<name>.*)['"](\r|\n|$)/;
+    const setupPyContents = await this.getSetupPyContents();
+    if (setupPyContents) {
+      const match = setupPyContents.match(ARTIFACT_NAME_REGEX);
+      if (match && match?.groups?.name) {
+        return match.groups.name;
+      }
+    }
+    return null;
+  }
```
chingor13, line 180: This strategy actually supports several types of python projects. There are other files besides setup.py that can specify the package name (see the buildUpdates() for the others)

chingor13 (changes requested):

src/strategies/python.ts
bcoe: Switched logic to 

`const artifactName = projectName ?? (await this.getNameFromSetupPy());`

If an artifact name is provided to the constructor, or found in a `pyproject.toml`, it will be used.

chingor13 (approved):

Issue #1842, opened 2023-02-06
Update versions of linked packages in package-lock.json to match package.json


padaszewski:
Related to https://github.com/googleapis/release-please/issues/464

 **Is your feature request related to a problem? Please describe.**
Let's consider following scenario:
![modules](https://user-images.githubusercontent.com/30108423/216920771-447d0ec2-769b-463b-9f48-ea196eea5882.png)
If there is an update for one of the modules A,B,C then the corresponding `package-lock.json` will be updated. However, if I update the CoreModule, then the other modules will also be updated (`node-modules` plugin), but the version of the CoreModule will not be updated in the `package-lock.json` of the other modules. After releasing it is necessary to run `npm install` (which causes a change in the module dir and may cause unwanted version updates in the future). 
 **Describe the solution you'd like**
The version of a submodule is updated in the `package-lock.json` file of the module that depends on it (which also gets a patch update because of the `node-modules` plugin).
 **Describe alternatives you've considered**
No alternative considered.

PR #1844, opened 2023-02-06
feat: Update linked packages in package-lock.json

padaszewski:
Hi, I'm making a start for #1842 as this is very important for me.
I'm not fully into the code and this is just the starting point in my opinion.
Would love if someone could assist on this one!

Addresses #1842 🦕

### DIFF
--- a/src/plugins/node-workspace.ts
+++ b/src/plugins/node-workspace.ts
@@ -339,16 +339,18 @@ export class NodeWorkspace extends WorkspacePlugin<Package> {
     _updatedVersions: VersionsMap
   ): CandidateReleasePullRequest[] {
     // NOP for node workspaces
-    candidates.forEach(c => {
-      c.pullRequest.updates.push({
-        path: 'package-lock.json',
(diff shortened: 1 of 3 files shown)


padaszewski:
Updated CLA

padaszewski:
This should be the right direction, but I'm still struggling with the test. Any help would be greatly appreciated!

chingor13 (changes requested):
Please add tests to ensure we don't regress.

> I'm still struggling with the test

To test, you can run:

* `npm i`
* `npm t`

If you don't have node available, you can run this in a stock GitHub Codespace (IIRC, node 19 is installed which should suffice).

As for the contents of the test, you can probably:
1. augment an existing node-workspace test to ensure that it has an update for the package-lock.json file
2. add a test to the package-lock-json updater to ensure it updates multiple versions

padaszewski:
Sorry, I have currently no time to get back to this, could anyone continue on that one?

PR #1852, opened 2023-02-13
fix(cargo-workspaces): expand globs in crate paths

### LINKED ISSUE #1851: Glob paths for Cargo workspaces are unsupported
Many projects that use Cargo workspaces have a top-level Cargo.toml and several crates in a subdirectory. A common pattern for this case is to use globs in the workspace definiton:

```toml
[workspace]
members = ["crates/*"]
```

Glob paths are currently unsupported by release-please. release-please just [iterates over each path](https://github.com/googleapis/release-please/blob/main/src/plugins/cargo-workspace.ts#L105) in `workspace.members`.

This seems like it would be a simple fix, since the [glob](https://www.npmjs.com/package/glob) NPM package supports a superset of the [glob grammar](https://docs.rs/glob/latest/glob/struct.Pattern.html) used by Cargo.

Stonks3141:

Fixes #1851 🦕


chingor13 (changes requested):
I don't think this is going to work like you think. In release-please, there is no cloned repo so a file glob won't actually look at files on disk. Instead, you will probably want to use our implementation of findFilesByName on the GitHub client instance

Stonks3141:
That should be fixed now, the github client has a `findFilesByGlob` method.

chingor13 (changes requested):
Thanks for this, changes look ok, but the tests are failing because it's making a new API call that isn't mocked. We use sinon to mock out methods. Example:

```ts
sandbox.stub(github, 'findFilesByGlob').resolves(['path/to/file1.txt', 'path/to/file2.txt']);
```

Stonks3141:
Thanks for the review! I couldn’t figure out how to run the tests locally or on my fork, so I just copied a nearby test and hoped for the best! I should be able to work on it tomorrow.

Stonks3141:
This should hopefully be fixed, I applied your suggestion.

chingor13 (approved):

PR #1865, opened 2023-02-27
fix(sequential-calls): Use push instead of concat when returning releases

### LINKED ISSUE #1864: When `sequential-calls` is enabled `createReleases` always returns an empty array
#### Environment details

We're using the GitHub action https://github.com/google-github-actions/release-please-action, and have a script that reads the output, see:
https://github.com/cloudquery/cloudquery/blob/0fc7caf321d0ce9513449826158d131d43cac464/.github/workflows/release_pr.yml#L24

#### Steps to reproduce

  1. Enable `sequential-calls` in the manifest releaser configuration
  2. See that the action output is always empty (the cause is release-please code, but the issue is visible in the action)

erezrokah:

Fixes #1864 🦕
Fixes https://github.com/google-github-actions/release-please-action/issues/546
Also fixes https://github.com/cloudquery/cloudquery/issues/7787

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -1116,7 +1116,7 @@ export class Manifest {
           releasesByPullRequest[pullNumber],
           pullRequestsByNumber[pullNumber]
         );
-        resultReleases.concat(releases);
+        resultReleases.push(...releases);
       }
       return resultReleases;
     } else {


src/manifest.ts
```diff
@@ -1116,7 +1116,7 @@ export class Manifest {
           releasesByPullRequest[pullNumber],
           pullRequestsByNumber[pullNumber]
         );
-        resultReleases.concat(releases);
+        resultReleases.push(...releases);
```
erezrokah, line 1119: `concat` does not modify the array, it always returns a new one

erezrokah:
Could also fix https://github.com/google-github-actions/release-please-action/issues/546

erezrokah:
Hi lovely release please team 👋 Is there anything missing on my end I should add to this PR?

chingor13 (approved):

erezrokah:
Thanks for the review, merge and quick release 🚀

PR #1875, opened 2023-03-13
feat: allow excluding file paths to filter out commits

### LINKED ISSUE #908: feature(monorepos): allow excluding file paths when create release PRs

More likely to happen in monorepos. We recently added a feature to a package in a monorepo that also updated tests snapshots in another package https://github.com/netlify/build/pull/2785

That resulted in creating a release for both packages, where the desired result is to create a release only for the first one.

Ability to exclude specific paths from the release diffing logic, for example `exclude: ["packages/build/tests"]`

We could separate https://github.com/netlify/build/pull/2785 into two PRs, one for each package but that means tests for both PRs will fail until we merge both.

 **Additional context**
We might be able to use GitHub commit diff API to get a list of changed paths, and return an empty commit array if all changed paths are excluded.
I think the place to implement it will be around https://github.com/googleapis/release-please/blob/d1e5714dd6f5082da1e5eb59662cd15846f40dc7/src/release-pr.ts#L330

alex-enchi:
Idea is to exclude commits if all files in those commits are from specific paths.

In my case I have multiple packages in `packages` and `.` umbrella package. Some of packages are different and handled by different workflow and I need to exclude those from umbrella.

Tested only with `config-file`

Note: I think it is possible to break it with `"package/foo":{"exclude-paths":["package/foo"]}` but I did not test it yet.

Fixes #908🦕

### DIFF
--- a/src/util/commit-exclude.ts
+++ b/src/util/commit-exclude.ts
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import {Commit} from '../commit';
-import {ReleaserConfig} from '../manifest';
+import {ReleaserConfig, ROOT_PROJECT_PATH} from '../manifest';
 import {normalizePaths} from './commit-utils';
 
 export type CommitExcludeConfig = Pick<ReleaserConfig, 'excludePaths'>;
@@ -36,7 +36,7 @@ export class CommitExclude {
     Object.entries(commitsPerPath).forEach(([path, commits]) => {
       if (this.excludePaths[path]) {
         commits = commits.filter(commit =>
-          this.shouldInclude(commit, this.excludePaths[path])
+          this.shouldInclude(commit, this.excludePaths[path], path)
         );
       }
       filteredCommitsPerPath[path] = commits;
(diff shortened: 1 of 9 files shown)


src/util/commit-exclude.ts
```diff
@@ -0,0 +1,55 @@
...
+    });
+    return filteredCommitsPerPath;
+  }
+
+  private shouldInclude(commit: Commit, excludePaths: string[]): boolean {
+    return (
+      !commit.files ||
+      !commit.files.every(file =>
+        excludePaths.some(path => file.indexOf(`${path}/`) === 0)
+      )
```
chingor13, line 52: I think there's an edge case here that will be difficult to account for if you split the exclude from the include logic. The original `CommitSplit` class does not strip files out that don't match the path -- it's just used to determine if that commit touches that directory (it can touch multiple components).

This looks like it's checking if there's any file that does not match the exclude path pattern. Consider a commit that touches: 
* `a/b/c`
* `d/e/f`
* `d/e/g`

We have 2 components with paths `a` and `d` and `d` is configured to exclude `d/e/*`. This commit has a file (`a/b/c`) that does not match the exclude pattern, and will not be excluded. If I understand the feature correctly, this commit should be excluded (there's no commit under path `d` that does not include `d/e/*`).

chingor13 (changes requested):
Thanks for submitting this!

I think you'll actually want to combine the exclude logic into the `CommitSplit` class (see the inline comment).

src/util/commit-exclude.ts
alex-enchi: hm, true there is such case. 
Main reason for splitting this into two (aside of not adding additional logic tree to commit split) was the need to handle exclusion on `.` root scope
Thanks for catching up that usecase! Will update soon with changes

alex-enchi: I spent some time on adding the logic into commit split, but in the end decided to handle exclusion logic separately.
I have updated it to handle exclusion only based on relevant to project files.

Main reasoning of keeping it as "split" then "filter" is logic is a bit cleaner that way
1. no special case to iterate over `.`
2. commit split is not bloated with additional cases to handle and does only 1 thing, it splits commits per package

What do you think?
All my attempts to include that logic in commit-split itself did not look pretty, unfortunately.

alex-enchi: @chingor13 i think now it covers all needed usecases. Could you take a look when you have time?

chingor13 (approved):

Issue #1440, opened 2022-05-18
Report rate-limiting errors clearly


Rauno56:
**Is your feature request related to a problem? Please describe.**

I suspect our releases [here](https://github.com/open-telemetry/opentelemetry-js-contrib/runs/6486601021?check_suite_focus=true), are erroring because of GH rate limiting, but we have no way to certainly tell - I suspect rate-limiting errors are just reported as Server Errors.

 **Describe the solution you'd like**

A perfect solution would include
1. warning about the rate-limit being close(GH has headers in the response to check for that),
2. error messages to indicate whether it was because of rate-limiting or not and
3. a way to avoid the rate-limiting by slowing the requests down or breaking the release up into parts.

 **Additional context**

* Link to our failing CI run: https://github.com/open-telemetry/opentelemetry-js-contrib/runs/6486601021?check_suite_focus=true
* Docs about the headers: https://docs.github.com/en/rest/overview/resources-in-the-rest-api#checking-your-rate-limit-status
* General docs about rate limits: https://docs.github.com/en/developers/apps/building-github-apps/rate-limits-for-github-apps

dermidgen:
Bump; this would really be nice. Large repos with many open PRs hit rate limits pretty quickly.

PR #1885, opened 2023-03-19
feat: add rate limiting Octokit plugin

erezrokah:

Fixes #https://github.com/googleapis/release-please/issues/1503 🦕
Also related to https://github.com/googleapis/release-please/issues/1440.
Could replace https://github.com/googleapis/release-please/pull/1504

This PR adds the recommended throttling plugin, see https://octokit.github.io/rest.js/v19#throttling.

This probably needs more testing and consideration before merging as it can have a big impact. Opening for feedback.
I can follow up on the failing tests if maintainers think this is a good approach

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "@lerna/package-graph": "^6.4.1",
     "@lerna/run-topologically": "^6.4.1",
     "@octokit/graphql": "^5.0.0",
+    "@octokit/plugin-throttling": "^5.0.1",
     "@octokit/request": "^6.0.0",
     "@octokit/request-error": "^3.0.0",
     "@octokit/rest": "^19.0.0",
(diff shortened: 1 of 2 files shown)


erezrokah:
Maybe we can put this behind an option so the default behavior is the same

chingor13 (changes requested):
I would prefer if we dependency inject the configured octokit APIs rather than trying to configure it here. This would let the caller use any octokit plugins without our `GitHub` class needing to understand every configuration option.

For example, in the GitHub app we run, we provide an authenticated Octokit instance created by probot. This would mean the instantiator of the `GitHub` wrapper instance would need to configure the throttle plugin. Within this repo, it would mean configuring it only in the CLI.

erezrokah:
Going to close as stale and I'll re-open once I have time to address the comments

PR #1908, opened 2023-04-10
fix(changelog-type): Fix github changelog-type to make it work

OleksandrKrupko:
Fixes https://github.com/google-github-actions/release-please-action/issues/744 🦕

### DIFF
--- a/src/github.ts
+++ b/src/github.ts
@@ -1477,7 +1477,13 @@ export class GitHub {
       previous_tag_name: previousTag,
       target_commitish: targetCommitish,
     });
-    return resp.data.body;
+
+    const versionWithoutV = resp.data.name.replace(/^v/, '');
+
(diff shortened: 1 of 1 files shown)


src/github.ts
```diff
@@ -1477,7 +1477,13 @@ export class GitHub {
       previous_tag_name: previousTag,
       target_commitish: targetCommitish,
     });
-    return resp.data.body;
+
+    const versionWithoutV = resp.data.name.replace(/^v/, '');
+
+    return resp.data.body.replace(
```
OleksandrKrupko, line 1483: I'm open to any kind of combination of version header & body, The one with just a version like in the default strategy and a list of changes looks like the most concise to me, though I didn't make the version header a link, because GitHub already adds the compare changes link to the bottom

PR #1909, opened 2023-04-10
build: update to latest cloud-rad doc generator

### LINKED ISSUE #1897: Kokoro fails: ERROR: The "export * as ___" syntax is not supported yet; as a workaround, use "import * as ___" with a separate "export { ___ }" declaration
```
ERROR: The "export * as ___" syntax is not supported yet; as a workaround, use "import * as ___" with a separate "export { ___ }" declaration
/[workspace/build/src/index.d.ts:1](https://cs.corp.google.com/piper///depot/google3/workspace/build/src/index.d.ts?l=1):8
```

chingor13:
This library is node >= 14. The 0.2.5 version of the doc generator does not appear to support node 14 yet.

Fixes #1897

### DIFF
--- a/.kokoro/release/docs-devsite.sh
+++ b/.kokoro/release/docs-devsite.sh
@@ -25,5 +25,5 @@ if [[ -z "$CREDENTIALS" ]]; then
 fi
 
 npm install
-npm install --no-save @google-cloud/cloud-rad@^0.2.5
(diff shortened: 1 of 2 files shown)


chingor13:
We're going to stop running the docs and docs-devsite jobs instead. This is a dev-tool and doesn't need reference docs published.

PR #1717, opened 2022-10-21
feat(api): add commit fields to access authors and PR number

minoism:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/release-please/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Fixes #1716

PR #1938, opened 2023-04-30
feat(api): Add the ability to include commit authors and PR references to default-changelog-notes

### LINKED ISSUE #1716: Add commit fields to access authors and PR number

Once these fields are added, we can write the authors of each commit in the changelog, just like the Github default release note.

I want to customize the changelog, but I can't access the authors and the number of PR associated with each commit.

 **Additional context**
Just want this:
> fix: Commit message by @author in #1

jamszh:

### Context:

Fixes #1716  🦕

This PR is a follow-up to #1717
We can probably close it as my PR is based off it.

### Changes:

- Utilise the `commitPartial` attribute from `BaseStrategyOptions`
- Add `commitPartialPath` to the manifest config and friends
- Commit authors and PR reference links can be added to the commit entries for a release by providing a custom `.release-please-commit-partial.hbs` file.
  - You can find the template [file here](https://github.com/conventional-changelog/conventional-changelog/blob/83773c5e1a8874ea9809ca9db1dff60b7df1daae/packages/conventional-changelog-conventionalcommits/templates/commit.hbs) (we default to this template when no `commitPartialPath` is provided)

### TESTING

I have done testing on this dummy repo.
https://github.com/jamszh/test-release-please

You can provide a path to a custom `commit-partial.hbs` file like [this one](https://github.com/jamszh/test-release-please/blob/main/.release-please-commit-partial.hbs). 

```bash
./build/src/bin/release-please.js release-pr --token $GITHUB_TOKEN --repo-url jamszh/test-release-please --target-branch main --commit-partial-path .release-please-commit-partial.hbs 
```

Here is an example [auto-release PR](https://github.com/jamszh/test-release-please/pull/2) generated by the command above.

For testing, the $GITHUB_TOKEN is a fine grain token with access to the test-release-please repo. I used these permissions.
image


docs/manifest-releaser.md
```diff
@@ -180,6 +180,11 @@ defaults (those are documented in comments)
   // absence defaults to https://github.com
   "changelog-host": "https://example.com",
 
+  // set custom formatted commit entry
+  // refer to the default conventional-changelog-writer:
+  // https://github.com/conventional-changelog/conventional-changelog/blob/11195f2822b389f6b5dc7216c4cf079e0b2883e7/packages/conventional-changelog-writer/index.js#L40
+  "commit-partial": "see above",
```
jamszh, line 186: Not sure if there is any alternative to dumping the contents of `commit.hbs` into this.

test/changelog-notes/default-changelog-notes.ts
```diff
@@ -20,12 +20,16 @@ import {
   safeSnapshot,
 } from '../helpers';
 import {DefaultChangelogNotes} from '../../src/changelog-notes/default';
-import {parseConventionalCommits} from '../../src/commit';
+import {ConventionalCommit, parseConventionalCommits} from '../../src/commit';
 import {PullRequestBody} from '../../src/util/pull-request-body';
 import {Version} from '../../src/version';
+const presetFactory = require('conventional-changelog-conventionalcommits');
+
+const COMMIT_WITH_AUTHOR_FORMAT = '{{#if authors}} by {{authors}}\n {{~/if}}\n';
```
jamszh, line 28: How to add PR author to a commit entry for the generated release notes.

I wonder if this feature be default or opt-in.

docs/manifest-releaser.md
jamszh: Oh I just realised I can probably just provide a file path. That is much better.

JohnCampionJr:
This would be very nice to have!

chingor13 (changes requested):
Sorry for the late review, but this is looking great!

One question -- is this opt-in? IMO, all the new features should not change existing behavior.

src/factory.ts
```diff
@@ -120,10 +142,12 @@ export async function buildStrategy(
     bumpMinorPreMajor: options.bumpMinorPreMajor,
     bumpPatchForMinorPreMajor: options.bumpPatchForMinorPreMajor,
   });
+  const commitPartial = await getCommitPartialContent(options);
   const changelogNotes = buildChangelogNotes({
     type: options.changelogType || 'default',
     github: options.github,
     changelogSections: options.changelogSections,
+    commitPartial,
```
jamszh, line 150: This entire feature is entirely opt-in.

`commitPartial` already existed as as optional argument prior to this PR.
It was just never used.

jamszh:
> Sorry for the late review, but this is looking great!
> 
> One question -- is this opt-in? IMO, all the new features should not change existing behavior.

Yep.
See: https://github.com/googleapis/release-please/pull/1938/files#r1322981745

jamszh:
> This would be very nice to have! Another related thought, maybe a list of excluded authors? For example, I wouldn't want myself listed by every single commit I do.

Yeah I can see that being annoying.
I guess we could also remove the `@` from the author's user ID when building the notes.

jxn-30:
What is the current state of this PR?
Would love to see this, especially in combination with the beforementioned exclude-list!

Thanks for your work!

jamszh:
> What is the current state of this PR? Would love to see this, especially in combination with the beforementioned exclude-list!
> 
> Thanks for your work!

I have updated the PR again. It should be ready for another review from @chingor13.
I'm thinking instead of an exclusion list, it might more sensible to add further capabilities to this `.hbs` file.
Perhaps there is a way to list the authors at the PR level, rather than the commit level.

chingor13 (changes requested):
This looks so close. I think we will want to pass the template string in the config, rather than making an extra GitHub API request to fetch the template from a file on the repository. The template contents should be small enough (generally a one-liner)

jamszh:
> This looks so close. I think we will want to pass the template string in the config, rather than making an extra GitHub API request to fetch the template from a file on the repository. The template contents should be small enough (generally a one-liner)

The example template I used is quite big.
https://github.com/jamszh/test-release-please/blob/main/.release-please-commit-partial.hbs

Though i'm not too sure how others will use it

Are you saying we should have the contents of the 'commit-partial.hbs' file as part of the config?
It doesn't feel like a one-liner to me, especially if users want something more customized (hence more complicated).

andrew-field:
Would love to see this merged. Thanks for your work!

Oudwins:
This would be amazing. Anything I can do to help push it over the edge?

flying-sheep:
@chingor13 what’s holding this up?

semaj-jing:
I've resolved the conflicts. 

This PR has some overlap with #2628. Both introduce a way of attaching an author to a release note, however the solution provided by this PR is slightly more customisable (requires providing a `commit-partial.hbs`) and allows for structuring the entire note as opposed to simply appending a single author.

Regardless, I've ensured that the functionality introduced in #2628 is preserved. Both are opt-in.


I've re-tested - everything seems to be good.


Here is the PR with generated release notes with the update changes in this branch.

https://github.com/jamszh/test-release-please/pull/5

PR #1968, opened 2023-05-31
fix(helm): maintain any existing comments in Chart.yaml

### LINKED ISSUE #1944: Preserve comments in Helm Chart.yaml when updating version
We currently use comments in our Chart.yaml so that Renovate can automatically update the `appVersion` field when new Docker images are available. With the current Helm updater, all comments are stripped out of the Chart.yaml, which rather breaks that workflow.

## Cause
This is caused by the use of [js-yaml](https://github.com/nodeca/js-yaml) which does not support preserving comments. [A similar issue](https://github.com/googleapis/release-please/issues/704#issuecomment-760345979) was seen in the Rust Cargo.toml updater, and a fix for this was implemented in #705. This was also clearly of concern when the KRM updater was developed - [this updater uses Regex](https://github.com/googleapis/release-please/blob/2b56b2213c10c6a0d3265746b10b19fbfce47292/src/updaters/krm/krm-blueprint-version.ts#L28) rather than using the js-yaml library.

## Alternatives 
[eemeli/yaml](https://github.com/eemeli/yaml) seems to be a reliable library which supports maintaining comments when parsing/dumping Yaml files. I've prototyped switching to this library [in my fork](https://github.com/googleapis/release-please/commit/7dedf0af81a746b1b1061f3ee7aec7a9162fe062), specifically for the Helm updater at this point. I'm happy to adapt this to a PR if the change is acceptable.

## Reproduction
This can be reproduced easily with the following Chart.yaml:
```
apiVersion: v2
name: hello
description: A Helm chart for Kubernetes
type: application
version: 0.5.0
# renovate: image=nginx
appVersion: 1.24.0
```
release-please will update this to:
```
apiVersion: v2
name: hello
description: A Helm chart for Kubernetes
type: application
version: 1.0.0
appVersion: 1.24.0
```

ben-foxmoore:
As raised in #1944, the current approach for Helm charts (using `js-yaml` library) strips all comments from the file. This PR switches to using the popular `yaml` library ([eemeli/yaml](https://github.com/eemeli/yaml)), and then patching the version value directly. This helps maintain any comments which have been added to the Chart.yaml - Renovate comments are the driving cause in our case.

Fixes #1944 🦕

### DIFF
--- a/__snapshots__/chart-yaml.js
+++ b/__snapshots__/chart-yaml.js
@@ -1,12 +1,27 @@
 exports['ChartYaml updateContent updates version in Chart.yaml 1'] = `
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 name: helm-test-repo
 version: 1.1.0
 apiVersion: v2
+# renovate: image=imageName
 appVersion: 2.0.0
 dependencies:
   - name: another-repo
     version: 0.15.3
-    repository: linkToHelmChartRepo
+    repository: "linkToHelmChartRepo"
 maintainers:
   - Abhinav Khanna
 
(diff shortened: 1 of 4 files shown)


ben-foxmoore:
CLA has just been signed.

chingor13 (approved):
Thanks! Looks great!

raja-anbazhagan:
@chingor13 @ben-foxmoore.  This issue relapsed again on latest versions. 

https://github.com/googleapis/release-please/blob/422df5a6a9d44a52c6c682563d1795b8fd246215/package.json#L89C1-L90C1


I have Chart.yaml with # x-release-please-start-version # x-release-please-end comments. But they are absolutely being scrapped.

raja-anbazhagan:
> This issue relapsed again on latest versions.

nevermind. I just had to use `generic`  type  to override file extension based updates.

image

PR #1982, opened 2023-06-13
docs: troubleshooting with "autorelease: pending" label

suztomo:
@chingor13 @diegomarquezp @rajatbhatta, Shall we add the explanation of "autorelease: pending" label explanation for Release Please troubleshooting?


suztomo:
I’m afraid that the root cause is not clear when the label is left with a release pull request. This documentation update does not address that.

chingor13 (approved):

GregKruz (approved):

PR #1997, opened 2023-06-28
feat: allow custom notes in body text

### LINKED ISSUE #1274: Additional custom description for Releases

Before I release a project, I like to include some written text with the release in order to better describe especially major changes. For that, I edit the release-please's comment on the PR which works nicely as release-please then use this text to create the actual GitHub's Release notes. There are two caveats though:

 1. I have to manually sync these changes with a commit to `CHANGELOG`
 2. If I don't do this as the last step before the release and some new commit lands into it, release-please override these changes in this comment and I have to fish in the edit history for it.

I would like to add a section into the release-please PR comment marked with comments that would be persisted across the PR's comment updates. For example:
```
<!-- RELEASE_PLEASE_EXTRA_DESCRIPTION -->
This block will be persisted across PR's comment update.
<!-- /RELEASE_PLEASE_EXTRA_DESCRIPTION -->
```

Also, GitHub supports [`issue_comment`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment) that allows tracking of updated comments. So the user of for example the Release Please Github Actions could trigger release-please also on updated comment. So the Release Please could have another "category of action" for this type of event to update `CHANGELOG` with updated comment.

There are manual workaround but they are more and more annoying to me.

btkostner:

Fixes #1274

This allows setting up custom notes in release changelogs as described in #1274. It does so by parsing magic strings in the commit message and setting the parsed text as a note. Because we are pigging back off notes, support for rendering this information is already done. Let me know how you feel about this approach and I can update as needed.


btkostner:
I just realized that the template will not render these notes like I want. I’ll update and add more tests in that area.

chingor13:
> I’ll update and add more tests in that area.

If you ever get back to this, feel free to re-open. The annotations used might warrant more discussion in the feature request issue though.

PR #2031, opened 2023-07-30
feat: allow specifying components by `Component` commit footer

### LINKED ISSUE #1905: Support specifying a component in "Relase-As:", empty commits monorepos
It's convenient to be able to use an empty commit to specify a release version as described [here](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md). In the [OpenFeature](https://github.com/open-feature) project, we have a number of monorepos that use release-please, and sometimes we want to create a release for a _specific_ component in a monorepo without changes (releasing a 1.0 is a good example).

There doesn't seem to be a way to do this with an empty commit. Currently doing the above results in releasing all components at the specified version. It would be great if you could do something like add `Component: my-library` in the commit body to tell the manifest release which component you mean.

Apologies if this is already possible... I couldn't find doc about it.

axieum:

Closes #1905 🦕


rozsazoltan:
@axieum This modification will be completed? I find it a very useful idea.

axieum:
> @axieum This modification will be completed? I find it a very useful idea.

Please see the comment https://github.com/googleapis/release-please/issues/1905#issuecomment-1657165294 on the original issue. We need to find the best way to approach parsing the commit footers before writing the implementation.

But alas, the failing tests are here and ready for whenever it wants to be picked up again.

PR #2036, opened 2023-08-05
feat: add support for crystal lang projects

LucianBuzzo:
This change adds support for projects written in [Crystal](https://crystal-lang.org/). In a Crystal project, version information is stored in the `version` field of a file called `shard.yml` in the root of your project.
Functionally, this new crystal strategy is the same as the pre-existing Helm strategy, so I've used that as the basis for my implementation.

### DIFF
--- a/src/strategies/crystal.ts
+++ b/src/strategies/crystal.ts
@@ -1,0 +1,81 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


LucianBuzzo:
I've now signed the CLA, so the check should pass if re-run.

LucianBuzzo:
@chingor13 What do you think of this?

LucianBuzzo:
@bcoe I think you're a regular maintainer here, could you take a look at this PR when you have a moment? Thanks! 🙏

chingor13 (changes requested):
Apologies I was OOO for several weeks.

Thank you for this and it's looking like a good start. We'll want some unit tests to make sure we don't have regressions (Perhaps look at the Helm strategy and chart.yaml updater tests).

Additionally, in order to use the strategies, it must be registered in the factory.ts so it can be loaded at configuration time.

chingor13:
If you have time to add tests, the please reopen this PR.

Issue #2037, opened 2023-08-08
Add default reviewers for release PRs


marcel-stainless:
When a release PR is created, it needs to be approved by a certain set of approvers. Would be nice if there could be a default set of reviewers for each project that release-please would then add

chingor13:
As mentioned in #2038, reviewers feels out of scope of this automation. There is often more logic desired for assigning or requesting reviews.

This most straightforward way on GitHub is to add a CODEOWNERS file/rule for the files that need updating.

PR #2038, opened 2023-08-08
RFC: Add reviewers as an option

marcel-stainless:
RFC for #[2037>](https://github.com/googleapis/release-please/issues/2037) 🦕

We want to be able to add default reviewers to release PRs. The attached code works for the library case, but before I add more support for CLI and config file, tests and docs, I wanted to know if this is an acceptable extension of what release-please should do (or if there is a better way to do this through plugins?). Also, where would be the best location to add this config. release-please-manifest.conf?


chingor13 (changes requested):
Thanks for submitting this. 

Adding reviewers feels out of scope of this automation. There is often more logic desired for assigning or requesting reviews.

This most straightforward way on GitHub is to add a CODEOWNERS file/rule for the files that need updating.

chingor13:
Feel free to add more discussion on #2037

PR #2041, opened 2023-08-09
feat: adds metadata versioning for terraform modules

### LINKED ISSUE #2040: Include metadata.yaml files for terraform modules version bumps
We currently use release-please for terraform modules. Recently we added a metadata file to these modules which specifies the version no. of the module. When release-please makes a release it updates the README.md and versions.tf but misses the metadata.yaml file leaving it stale. This feature extends our existing update strategy for terraform modules to include metadata.yaml files as well.

g-awmalik:

Fixes #2040

### DIFF
--- a/src/strategies/terraform-module.ts
+++ b/src/strategies/terraform-module.ts
@@ -91,16 +91,11 @@ export class TerraformModule extends BaseStrategy {
     });
 
     // Update metadata.yaml to current candidate version.
-    let metadataFiles: string[];
-    try {
-      metadataFiles = await this.github.findFilesByFilenameAndRef(
-        'metadata.yaml',
-        this.targetBranch,
-        this.path
-      );
-    } catch (error) {
-      metadataFiles = [];
-    }
+    const metadataFiles = await this.github.findFilesByFilenameAndRef(
+      'metadata.yaml',
+      this.targetBranch,
+      this.path
+    );
 
     metadataFiles.forEach(path => {
       updates.push({


g-awmalik:
@bcoe - please review when you get a chance.

src/strategies/terraform-module.ts
```diff
@@ -88,6 +89,26 @@ export class TerraformModule extends BaseStrategy {
         }),
       });
     });
+
+    // Update metadata.yaml and metadata.display.yaml to current candidate version.
+    const metadataFiles = await Promise.all([
```
bcoe, line 94: I think this could be simplified to:

```js
const metadataFiles = await this.githubFilesByFilename(...)
```

You shouldn't need the `Promise.all` unless you are chaining together multiple async requests.

src/strategies/terraform-module.ts
```diff
@@ -88,6 +89,26 @@ export class TerraformModule extends BaseStrategy {
...
+      this.github.findFilesByFilenameAndRef(
```
bcoe, line 95: Does this method return an empty array if no files are found for the search?

test/updaters/metadata.ts
```diff
@@ -0,0 +1,38 @@
...
+import {MetadataVersion} from '../../src/updaters/terraform/metadata-version';
+import {Version} from '../../src/version';
+
+const fixturesPath = './test/updaters/fixtures';
+
+describe('metadata.yaml', () => {
+  describe('updateContent', () => {
+    it('updates version in metadata.yaml', async () => {
+      const oldContent = readFileSync(
+        resolve(fixturesPath, './metadata.yaml'),
```
bcoe, line 28: Thanks for the tests, looks good!

bcoe (changes requested):
One nit, I think the logic can be simplified slightly.

Also, I left a question about the `findFilesByFilenameAndRef` method, it would be good to confirm that it returns an empty array if no results are found.

src/strategies/terraform-module.ts
g-awmalik: Done.

src/strategies/terraform-module.ts
g-awmalik: Honestly, I can't tell looking at the source. I'm also a TS novice so I might be missing it completely. But I've wrapped the call in a try/catch so even if it errors out, we'll set `metadataFiles` to an empty array.

g-awmalik:
> One nit, I think the logic can be simplified slightly.
> 
> Also, I left a question about the `findFilesByFilenameAndRef` method, it would be good to confirm that it returns an empty array if no results are found.

PTAL.

bcoe (approved):

src/strategies/terraform-module.ts
```diff
@@ -88,6 +89,28 @@ export class TerraformModule extends BaseStrategy {
...
+    let metadataFiles: string[];
+    try {
+      metadataFiles = await this.github.findFilesByFilenameAndRef(
+        'metadata.yaml',
+        this.targetBranch,
+        this.path
+      );
+    } catch (error) {
+      metadataFiles = [];
+    }
```
bcoe, line 103:
```suggestion
    const metadataFiles = await this.github.findFilesByFilenameAndRef(
      'metadata.yaml',
      this.targetBranch,
      this.path
    );
```

Issue #510, opened 2020-07-31
Support for pre-release


jcornaz:
**Is your feature request related to a problem? Please describe.**

I would like to cut a [pre-release](https://semver.org/#spec-item-9)

 **Describe the solution you'd like**

Have a (documented) way to configure *release-please* so that it knows that I am publishing a pre-release.

It could be as simple as an argument of the CLI (example: `--prerelease=alpha`) and/or it could be via configuration file to map branch name with release type.

 **Describe alternatives you've considered**

Well, except using another tool than *release-please*, I don't see how to achieve that.

 **Additional context**

I'm not sure what to add here. The context is releasing a software early as a pre-release...

cobraz:
I'd love to help out with this!

parthea:
It appears that release please only supports the `alpha` suffix for pre-releases, although I haven't tried it. https://github.com/googleapis/release-please/blob/master/src/release-pr.ts#L308 . For python, I'd like to use these suffixes `devN`, `aN`, `bN`, and `rcN`. See 
https://packaging.python.org/guides/distributing-packages-using-setuptools/#pre-release-versioning

chingor13:
v13 now supports more robust version handling (prerelease and build numbering). If we build something for this, we need to design how we signal to release please that we need to bump to a prerelease candidate versus a stable release. The CLI could accomplish this via a CLI option, but the GitHub action and app would not be able decide whether or not to use that flag.

jcornaz:
> 

The github action is configurable: https://github.com/marketplace/actions/release-please-action#configuration

Example: 

```yaml
      - uses: GoogleCloudPlatform/release-please-action@v2
        with:
          pre-release: beta
```

For the gihub app I don't know. But many github apps are configurable via a configuration file in the `.github` directory.

jcornaz:
Hi, May I kindly ask where are we with this issue? May I help somehow? Are there still open questions that need to be solved first?

bcoe:
@jcornaz @cobraz @chingor13 my suggestion is to submit a `rfcs/pre-release.md` as a PR, which provides a proposed design for how we should handle pre-releases, we can move the discussion there.

Once we get a design in writing we like, then we can work on implementation. I'd love to start moving release-please towards this RFC approach, so that our design is intentional.

erezrokah:
Thanks @bcoe, I have a suggestion that could improve the current state a bit (happy to write an RFC for it if you think it makes sense).
[...]

PR #1981, opened 2023-06-12
feat: add PrereleaseVersioningStrategy

chingor13:
This adds a new versioning strategy that can handle prerelease versioning used by google-cloud-dotnet.

For a `fix` type of release, it will bump the prerelease number and preserve any padded `0`'s.
Example:
* `1.2.3-beta01` => `1.2.3-beta02`
* `1.2.3-beta09` => `1.2.3-beta10`.

PR #2051, opened 2023-08-24
doc: RFC for handling prerelease strings.  ref #510

broofa:
Per [this comment in #510](https://github.com/googleapis/release-please/issues/510#issuecomment-1020248692), creating an RFC doc for specifying how prerelease versions should be handled.

### DIFF
--- a/rfcs/pre-release.md
+++ b/rfcs/pre-release.md
@@ -35,6 +35,6 @@ The SemVer spec does not go into detail about the semantics of prerelease versio
 
 > Pre-release versions have a lower precedence than the associated normal version
 
-Other than being preceeded by a hyphen and consisting of "a series of dot separated identifiers", prerelease versions can be *anything*.  I.e. there's no semantic correlation to the major-minor-patch fields.  Nor is there any requirement for it contain a numeric component.
+Other than being preceeded by a hyphen and consisting of "a series of dot separated identifiers", prerelease versions can be *anything*.  There's no requirement there even be a numeric component to them, let alone any semantic relationship to the normal major-minor-patch fields.
 
-Thus, (for now) we should use the simplest-possible logic for implementing version bumps, as described above.
+Thus, applying the simple and (hopefully) intuitive right-most-number logic described above should will minimize the amount of confusion this causes.  If/when users demand more involved logic, new schemes can be added at a future date.


chingor13 (changes requested):
Thanks for submitting this. We are definitely open to adding additional versioning strategies.

Additional considerations:
* which commit types influence each version bump
* how (or if) the primary semver numbers will change
* how (or if) the release becomes non-prerelease

Additional note: we added a `prerelease` versioning strategy that follows our googleapis/google-cloud-dotnet libraries handle pre-release versioning in #1981. If that one does not suit your needs, it could be used as a blueprint for a similar new versioning strategy.

broofa:
@chingor13 Thanks for the feedback.  You ask good questions.  If you don't mind, I'd like to reply here and get your thoughts before updating the PR...

> Additional considerations:
> 
> * which commit types influence each version bump

This is a tricky question.

For starters, the [Conventional Commits spec](https://www.conventionalcommits.org/en/v1.0.0/#specification) doesn't address prerelease fields at all.  There's simply no mention of "prerelease" in the spec.  I suspect this is because the authors don't consider the prerelease field to be versionable in the way "primary fields" (major / minor / patch) are.

Yet [the SemVer spec](https://semver.org/) does devote a fair bit of verbiage to the prerelease field.  Notably, [it says](https://semver.org/#spec-item-9) _"Pre-release versions have a lower precedence than the associated primary version"_.  But, unlike the minor and patch fields, no guidance is given on how it should behave when higher-precedence fields are bumped.  E.g. The patch field "MUST" be set to zero if either the major or minor fields is incremented.  But it says nothing about the prerelease field.

Furthermore, in practice the prerelease string is most-often used to track "significant" branches of development off the main branch.  For example it's not uncommon for a project that has a 1.x release in production to push out a 2.x beta release with semvers like "2.0.0-beta-1".  The semantic being, "**this is a _candidate for the 2.0.0 release_**".

But what happens if someone adds a "BREAKING CHANGE" commit to that branch?  Does the version get bumped to "3.0.0-beta-1"?  That's what Conventional Commits would ordinarily mandate.  And if I'm reading the `prelease` versioning strategy PR you link to properly, that's what would happen there, right?

... but that doesn't really make sense.  In this [common] scenario, the breaking change is redundant with the fact this is a "2.x-beta" candidate.

In other words ...

* If a semver has a prerelease field, then I think the assumption needs to be that the primary fields reflect the _intended_ version, once the prerelease tag is removed.  As such, if a prerelease field is present I don't think the commit type should influence primary fields.

> * how (or if) the primary semver numbers will change

See above.  tl;dr: They shouldn't.

> * how (or if) the release becomes non-prerelease

See above.  I think the conclusion is that the prerelease field is removed, with no change to the primary semver numbers.

> Additional note: we added a `prerelease` versioning strategy that follows our googleapis/google-cloud-dotnet libraries handle pre-release versioning in #1981. If that one does not suit your needs, it could be used as a blueprint for a similar new versioning strategy.

Interesting.  I wasn't aware of this work.  My initial reaction is that this seems complicated and problematic.  Again, if a semver has a prerelease field, I think the semantic needs to be that the primary fields are the _intended_ version, upon actual release.  And, as such, they shouldn't change.

Thoughts?

broofa:
In thinking about this further I find myself a bit conflicted.  The semantic of prerereleases is pretty clear: They are candidates for the version mentioned in the primary fields.  E.g. "2.0.0-beta.1" is a _candidate_ for the 2.0.0 release.

As such, my argument for not touching the primary fields is sound... _but I'm realizing that's only if it's a major-release candidate_(!)

Things get tricky when we start talking about minor or patch release candidates.  For example, "2.3_beta.1" would mean be a candidate for the "2.3" release. (Implying that the current release is 2.2.x.   But what happens if there's a BREAKING CHANGE commit?(??)

(... which I suspect @chingor13 's point to begin with. 😄 )

Three options come to mind:
plays)
* Apply the appropriate change to the primary fields and reset the prerelease field.  E.g. the version might become "3.0.0_beta.0"

broofa:
So... I may need to backpeddle on all of this.  I was only considering major-prerelease versions.  I wasn't thinking about lessor versions like "2.3_beta.1".  😬

When you start thinking through the lessor-version scenarios, I think you kind of inevitably land on a `prerelease` strategy like the one you've implemented for `googleapis/google-cloud-dotnet`.

I need to think through this further but, alas, don't have time at the moment.  Before I go, though, I'll note that the `semver` module has a bunch of [test cases for bumping prereleases](https://github.com/npm/node-semver/blob/123e0b03287e1af295ef82d55f55c16805596f35/test/fixtures/increments.js#L39).    Do you happen to know if the `prerelease` strategy you've implemented aligns with those?

thebergamo:
If you don't mind I would like to add my two cents here, even thought I think it should be covered in the test cases aforementioned.

In general, coming from for example [semantic-release](https://github.com/semantic-release/semantic-release) module, I would expect that the pre-release would be just a case to be bumped based on the build itself (of course considering changes).

For instance, if in my first commit to introduce this new release would be simply some new feature, we should just bump the `minor` version, but if for some reason we have to introduce a `BREAKING CHANGE` it should simply bump `major` and reset the  pre-release version(as per proposed being .0 again).

If after first Breaking change other breaking changes or features get added, it shouldn't change the target version for example `1.0.0.rc-0++`(++ in here is just to refer as incrementing this version) same as other bumps if they happen, the most important is bumped and pre-release restart the increments only if version changes.

This is very useful feature for example in cases were we have this PR releases as things are there to be released in a trunk(for sprint wise for instance) but we would like still to be able to `pre-release` to some environment for some QA tests for instance and having the pre-release there would be nice for visibility of the versions being tested and linked to the PR as well.

PR #2063, opened 2023-09-07
test: also test on node 18

chingor13:
Use `--no-experimental-fetch` node option for node 18

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [14, 16]
+        node: [14, 16, 18]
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -32,7 +35,9 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v3
         with:
-          node-version: 14
+          node-version: 18
+      - name: Override NODE options for the most recent version
+        run: echo "NODE_OPTIONS=--no-experimental-fetch" >> $env:GITHUB_ENV
```
bcoe, line 40: This actually works in Node 18, I'm surprised.

bcoe (approved):

PR #2072, opened 2023-09-12
feat: Use default updaters based on file extension

### LINKED ISSUE #2067: Create default updaters for each language based on file extension

When the release is created, I'd like to update the version number in two JSON files (`src/manifest-chromium.json` and `src/manifest-firefox.json`). You need to add the following configuration:

```JSON
{
  "extra-files": [
    {
      "type": "json",
      "path": "src/manifest-chromium.json",
      "jsonpath": "$.version"
    },
    {
      "type": "json",
      "path": "src/manifest-firefox.json",
      "jsonpath": "$.version"
    }
  ]
}
```

I would just like to add only the files to the configuration and release-please find the right strategy to update the version.

```JSON
{
  "extra-files": ["src/manifest-chromium.json", "src/manifest-firefox.json"]
}

```

Create default updaters for each language based on file extension.

| Extension | Updater | Compatibility |
| --------- | ------- | ------------- |
| `.json` | [GenericJson](https://github.com/googleapis/release-please/blob/main/src/updaters/generic-json.ts) with jsonpath: `"$.version"` | [`package.json`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#version) (npm), [`manifest.json`](https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/manifest.json/version) (WebExtension), [`composer.json`](https://getcomposer.org/doc/04-schema.md#version) (Composer) |
| `.xml` | [GenericXml](https://github.com/googleapis/release-please/blob/main/src/updaters/generic-xml.ts) with xpath: `"/*/version"` | [`pom.xml`](https://maven.apache.org/ref/3.9.4/maven-model/maven.html#project) (Maven) |
| `.yaml`, `.yml` | [GenericYaml](https://github.com/googleapis/release-please/blob/main/src/updaters/generic-yaml.ts) with jsonpath: `"$.version"` | [`pubspec.yaml`](https://dart.dev/tools/pub/pubspec#version) (Dart), [`Chart.yaml`](https://helm.sh/docs/topics/charts/#the-chartyaml-file) (Helm) |
| `.toml` | [GenericToml](https://github.com/googleapis/release-please/blob/main/src/updaters/generic-toml.ts) with jsonpath: `"$.version"` | |

**Additional context**

Before, I used to use [standard-version](https://github.com/conventional-changelog/standard-version/blob/a316dd0/lib/updaters/types/json.js#L13), which automatically updated the JSON file.

Now, I'm using release-please-action, which doesn't allow you to specify the `type` and `jsonpath` of the `extra-files`. https://github.com/google-github-actions/release-please-action/issues/797

regseb:

Fixes #2067


src/strategies/base.ts
```diff
@@ -428,6 +428,30 @@ export abstract class BaseStrategy implements Strategy {
...
+          path: this.addPath(extraFile),
+          createIfMissing: false,
+          updater: new GenericToml('$.version', version),
+        });
+      } else if (extraFile.endsWith('.xml')) {
+        extraFileUpdates.push({
+          path: this.addPath(extraFile),
+          createIfMissing: false,
+          updater: new GenericXml('/*/version', version),
+        });
```
chingor13, line 454: This is potentially a breaking change for users. These file types support the `Generic` updater's annotations. We do have a `CompositeUpdater` which we should probably use both the `Generic` and `Generic<Format>` updaters. This allows substituting both the exact version field and any additional annotated values.

chingor13 (changes requested):
Thanks for this and it seems reasonable as long as we can avoid breaking existing users.

src/strategies/base.ts
```diff
@@ -428,6 +428,30 @@ export abstract class BaseStrategy implements Strategy {
...
+        extraFileUpdates.push({
+          path: this.addPath(extraFile),
+          createIfMissing: false,
+          updater: new GenericToml('$.version', version),
+        });
+      } else if (extraFile.endsWith('.xml')) {
+        extraFileUpdates.push({
+          path: this.addPath(extraFile),
+          createIfMissing: false,
+          updater: new GenericXml('/*/version', version),
```
chingor13, line 453: Additionally, we might want to have a confidence test to ensure that this xpath only updates the first-level `version` field.

i.e.

```xml
<some-root>
  <version>1.2.3</version><!-- this one should be updated -->
  <blah>
    <version>2.3.4</version><!-- this one should NOT be updated -->
  </blah>
</some-root>
```

chingor13: Or if it's basic knowledge of xpath, then add a comment describing what this is matching.

regseb: [`/*/version`](http://xpather.com/f7pP8ml2):

- `/` to start the path to the root of the XML document.
- `*` to pass the root element (without filtering on its name).
- `/` to select all directed children.
- `version` to select elements named `version`.

I added a comment.

chingor13 (approved):
Thanks!

PR #2083, opened 2023-09-19
feat: prerelease type

### LINKED ISSUE #2082: Prerelease type for new prerelease versions
The prerelease version updater is a great feature that allows to bump the prerelease type in a particular way. However, when the version doesn't have a prerelease part, it doesn't do anything. Therefore, it would be nice to have a way that allows to have bumps such as 
- `1.3.3` to `1.3.4-PTPT` fore patches
- `1.3.3` to `1.4.0-PTPT` for features
- and `1.3.3` to `2.0.0-PTPT` for breaking changes

where `PTPT` is a custom prerelease type.

kalosisz:

Fixes #2082 🦕 by introducing an argument to the `prerelease` versioning strategy.

It also deals with bumping versions with prerelease, but not having a number in the prerelease part in the following ways  
- `2.0.0-beta` will get bumped to `2.0.0-beta.1` if there are breaking changes or less
- `1.2.0-beta` will get bumped to `1.2.0-beta.1` if there are featurer or less
- `1.2.2-beta` will get bumped to `1.2.2-beta.1` if there are only patches.

The current behaviour of prelrease, ie. bumping `1.1.1-beta01` to `1.1.1-beta02` is preserved.

### DIFF
--- a/docs/manifest-releaser.md
+++ b/docs/manifest-releaser.md
@@ -172,6 +172,9 @@ defaults (those are documented in comments)
   // absence defaults to false
   "bump-patch-for-minor-pre-major": true,
 
+  // setting the type of prerelease in case of prerelease strategy
+  "prerelease-type": true,
+
   // set default conventional commit => changelog sections mapping/appearance.
   // absence defaults to https://git.io/JqCZL
   "changelog-sections": [...],
--- a/docs/manifest-releaser.md
+++ b/docs/manifest-releaser.md
@@ -173,7 +173,7 @@ defaults (those are documented in comments)
   "bump-patch-for-minor-pre-major": true,
 
   // setting the type of prerelease in case of prerelease strategy
-  "prerelease-type": true,
+  "prerelease-type": "beta",
 
   // set default conventional commit => changelog sections mapping/appearance.
   // absence defaults to https://git.io/JqCZL
--- a/schemas/config.json
+++ b/schemas/config.json
@@ -22,7 +22,7 @@
         },
         "prerelease-type": {
           "description": "Configuration option for the prerelese versioning strategy. If prerelease strategy used and type set, will set the prerelese part of the version to the provided value in case prerelease part is not present.",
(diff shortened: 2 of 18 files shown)


docs/manifest-releaser.md
```diff
@@ -172,6 +172,9 @@ defaults (those are documented in comments)
   // absence defaults to false
   "bump-patch-for-minor-pre-major": true,
 
+  // setting the type of prerelease in case of prerelease strategy
+  "prerelease-type": true,
```
chingor13, line 176: Should this be a string?

schemas/config.json
```diff
@@ -20,6 +20,10 @@
           "description": "Feature changes only bump semver patch if version < 1.0.0",
           "type": "boolean"
         },
+        "prerelease-type": {
+          "description": "Configuration option for the prerelese versioning strategy. If prerelease strategy used and type set, will set the prerelese part of the version to the provided value in case prerelease part is not present.",
+          "type": "boolean"
```
chingor13, line 25: Same here - should this be a string?

test/factory.ts
```diff
@@ -60,19 +61,24 @@ describe('factory', () => {
       expect(await strategy.getComponent()).not.ok;
       expect(strategy.changelogNotes).instanceof(DefaultChangelogNotes);
     });
-    it('should build a with configuration', async () => {
+    it('should build with prerelease type', async () => {
```
chingor13, line 64: nit: can we add this as a separate test and leave the existing one?

chingor13 (changes requested):
Thanks for this. A few minor comments/nits.

docs/manifest-releaser.md
kalosisz: done

kalosisz:
hey @chingor13, please check the changes :) thanks

chingor13 (approved):
Thanks!

PR #2097, opened 2023-10-15
docs(readme): add link to release-plz

MarcoIeni:
Hi 👋
Thanks for creating release-please. I liked the workflow of this tool so much that it inspired me to create [release-plz](https://release-plz.ieni.dev/).
Release-plz is similar to release-please, but it's meant to only work in Rust:
- it's integrated with [crates.io](https://crates.io/): it publishes packages there and it checks if a package is already published (it doesn't rely on git tags for this).
- it uses [cargo-semver-checks](https://github.com/obi1kenobi/cargo-semver-checks) to detect api breaking changes. If so, it bumps the major version.
- No configuration required: you can `git clone` a repository and run `release-plz update` on it and it will update Changelog and Cargo.toml.
- More about this [here](https://release-plz.ieni.dev/docs/why#differences-with-release-please).

I would like to add release-plz to the readme because Rust developers might be interested in this tool.
However, if you prefer not to link to external projects in your readme, I understand.

Again, thanks again for creating this tool. 🙏


chingor13:
Thanks for this, but I don't think we are ready to "recommend" products we are actively watching/integrating with. Super excited that this project motivated you!

Issue #2109, opened 2023-10-31
node-workspace calculates new candidates incorrectly, resulting in broken or failed PRs


TimothyJones:
During a manifest release, sometimes new candidate pull requests are made because their dependencies are updating. These are made by calling the workspace's `newCandidate` method.

However, the logic for a new candidate in node-workspace is missing some of the information required to correctly calculate what the PR should look like. For [example](
https://github.com/googleapis/release-please/blob/main/src/plugins/node-workspace.ts#L309), the branch name is calculated from the target branch:

```
      headRefName: BranchName.ofTargetBranch(this.targetBranch).toString(),
```

But, it should instead be calculated using the component, as it is in [the base strategy](https://github.com/googleapis/release-please/blob/main/src/strategies/base.ts#L295)

```
    const branchComponent = await this.getBranchComponent();
    const branchName = branchComponent
      ? BranchName.ofComponentTargetBranch(branchComponent, this.targetBranch)
      : BranchName.ofTargetBranch(this.targetBranch);
```

This is causing PRs to clobber each other, resulting in incorrect behaviour:

```
✔ Updating 4 packages
❯ package: @contract-case/case-boundary
❯ version: 0.9.0 from release-please
❯ package: @contract-case/case-connector
❯ version: 0.6.6 from release-please
❯ package: @contract-case/cli
❯ version: 0.1.7 forced bump
❯ package: @contract-case/contract-case-jest
❯ version: 0.11.1 forced bump
✔ Updating exising candidate pull request for @contract-case/case-boundary, path: packages/case-boundary
✔ Updating @contract-case/case-boundary to 0.9.0
✔ Updating exising candidate pull request for @contract-case/case-connector, path: packages/case-connector
✔ Updating @contract-case/case-connector to 0.6.6
✔ @contract-case/case-connector.@contract-case/case-boundary updated to 0.9.0
✔ appending dependency notes to changelog
✔ appending dependency notes to changelog
✔ Creating new candidate pull request for @contract-case/cli
✔ Updating @contract-case/cli to 0.1.7
✔ @contract-case/cli.@contract-case/case-boundary updated to 0.9.0
✔ Creating new candidate pull request for @contract-case/contract-case-jest
✔ Updating @contract-case/contract-case-jest to 0.11.1
.......
✔ Successfully updated reference release-please--branches--main--components--@contract-case/case-boundary to 5934f1a6e7167bcf4a1173054282104835e47e7d
[...]

Issue #1946, opened 2023-05-09
Release please no longer making release PRs (There are untagged, merged release PRs outstanding - aborting)


TimothyJones:
If the support paths suggested above still do not result in a resolution, please provide the following details.

#### Environment details

  - OS: google-github-actions/release-please-action@v3
  - Node.js version: google-github-actions/release-please-action@v3
  - npm version: google-github-actions/release-please-action@v3
  - `release-please` version: google-github-actions/release-please-action@v3

#### Information

Somehow my repository ( https://github.com/case-contract-testing/contract-case ) has got into a state where release PRs are no longer being created due to:

```
⚠ There are untagged, merged release PRs outstanding - aborting
```

I tried to fix this by manually tagging a past release PR (which worked for generating [this release PR](https://github.com/case-contract-testing/contract-case/pull/112)), but after merging it, no tags or releases were generated, and the repo was left in this state again. 

#### Steps to reproduce

  1. Unclear, I'm afraid

TimothyJones:
I think I misread the documentation of how to add projects - I was editing the .release-please-manifest.json for new projects. I'll close this for now, and will open another issue if I this happens again.

TimothyJones:
So, this is still happening. 

I had thought that it might be because I'd corrupted the tags somehow. I tried cloning the repository without any tags and doing a bootstrap, but it still won't make releases after the release PR is merged.

The logs from a merge of a PR that should have made a release can be seen [here](https://github.com/case-contract-testing/contract-case/actions/runs/5153850050/jobs/9281695916)

Any ideas?

TimothyJones:
In case you can't see the log run, I've put it in a gist [here](https://gist.github.com/TimothyJones/4796a1631db931ff63cccfdedcbc2330)

phucvinh57:
Same issue

phucvinh57:
@TimothyJones Github PRs created by `release-please-actions` are tagged with labels named like `autorelease:*`. I tried to remove these labels from all PRs (include closed PR), then re-run workflows and suceed !

![image](https://github.com/googleapis/release-please/assets/69946748/913238f9-180f-421a-888c-ea62ebceac38)

TimothyJones:
Thanks. That's how I manually fix this every time I merge a release pr:

* Remove that label from all PRs
* Tag the last merge commit with all the relevant versions

However, release-please still doesn't create any releases on the next merge of a release PR. For example:

[...]

Issue #2089, opened 2023-09-29
Changelog doesn't include version header for workspace dependency bumps


chrispcampbell:
#### Environment details

  - OS: Ubuntu 20.04 (GitHub Actions runner) 
  - `release-please` version: 15.13.0 (via `release-please-action` v3.7.12)

#### Related issues

Note: This issue was mentioned in a [later comment](https://github.com/googleapis/release-please/issues/1978#issuecomment-1625631187) in related issue #1978, but I decided to file a separate issue for this changelog update problem because it seems separate from the other issues described in #1978, and maybe the fix is simpler for this one if done in isolation.

#### Steps to reproduce

See an [example PR](https://github.com/climateinteractive/SDEverywhere/pull/369) created by release-please in our [SDEverywhere repo](https://github.com/climateinteractive/SDEverywhere).  This is a monorepo that uses the node-workspace plugin.

In this example, release-please created a PR that included 1 package that had a fix, and 2 packages that had a "workspace dependencies" version bump.  

The changelog update for the `compile` package entry is correct and contains the version header:
<img width="894" alt="image" src="https://github.com/googleapis/release-please/assets/438425/cdde5b4e-206e-4964-9b8a-805d2588114e">

However, the changelog updates for the other two packages are missing the version header:
<img width="894" alt="image" src="https://github.com/googleapis/release-please/assets/438425/1ab35516-2ef5-4390-bd0b-3dbf17b02a5e">

Every time this happens I have to go in and manually edit the `CHANGELOG.md` files, otherwise the changelog becomes very confusing.

The release note entries in the PR summary are also missing the headers, but this is less of a problem for me than the missing entries in the `CHANGELOG.md` files.  

In the following screenshot of the PR summary, note that the `compile` package entry is correct and contains the version header, but the `cli` and `create` entries are missing the version header.  Additionally, those show up with the full package name (e.g., `@sdeverywhere/create`) while the `compile` package does not have the org prefix.  (This is a minor detail and doesn't bother me as much, but still worth noting.)

<img width="683" alt="image" src="https://github.com/googleapis/release-please/assets/438425/db0d30c0-8fc4-4970-93f8-8d2f78d68b5d">

chrispcampbell:
This issue may have been resolved by PR #2112 (released in 16.6.0) but I won't be able to verify until there is a new release of release-please-action based on release-please >= 16.6.0.

[...]

PR #2112, opened 2023-11-03
fix: Use strategies for workspace plugins.

### LINKED ISSUE #1978: Dependency version bumps using the node-workspace plugin to not update version numbers for "extra-files"
#### Environment details

  - OS: Ubuntu 22.04.2
  - Node: 16
  - `release-please` version: 15.10.4

#### Steps to reproduce

1. Using a mono repo with multiple node packages, and with dependencies between those packages. Repo in question: https://github.com/launchdarkly/js-core
2. A package, that is dependent on another package, contains files which are specified in "extra-files" that require version updates. File: https://github.com/launchdarkly/js-core/blob/cdc0acc6cb73eeb98baf7335e950ebc00f4f841f/packages/sdk/vercel/src/createPlatformInfo.ts#L13, Configuration: https://github.com/launchdarkly/js-core/blob/cdc0acc6cb73eeb98baf7335e950ebc00f4f841f/release-please-config.json#L16
3. Update a package that another package depends on.

When the version is bumped by the node-workspaces plugin it seems to only be bumped in the package.json and not in any files specified in the "extra-files". Release PR that demonstrates this: https://github.com/launchdarkly/js-core/pull/147

kinyoklion:

Fixes #1978,  #2089 and maybe #2109 🦕

The main goal here is to generate a base pull request using the strategy defined for the release. Then add to that the updates that are needed specifically to the workspace. This will result in generating the correct changelog header as well as updating "extra-files".

I need to implement tests for this yet, but I wanted to get some opinions on the approach before piling too much code on. I've only made changes for the node workspace, but the concept could be extended to to the other workspace plugins.

Example pr created with these changes showing fixes for #1978 #2089:
https://github.com/kinyoklion/release-please-monorepo-tests/pull/10/files


src/plugins/node-workspace.ts
```diff
@@ -271,20 +278,31 @@ export class NodeWorkspace extends WorkspacePlugin<Package> {
     );
     const packageJson = updatedPackage.toJSON() as PackageJson;
     const version = Version.parse(packageJson.version);
+
+    const strategy = this.strategiesByPath[updatedPackage.location];
+    const latestRelease = this.releasesByPath[updatedPackage.location];
+
+    const basePullRequest = strategy
```
kinyoklion, line 285: Make a base pull request that contains the "extra-files" updaters as well as the changelog header.

src/plugins/node-workspace.ts
```diff
@@ -297,16 +315,17 @@ export class NodeWorkspace extends WorkspacePlugin<Package> {
           createIfMissing: false,
           updater: new Changelog({
             version,
-            changelogEntry: appendDependenciesSectionToChangelog(
-              '',
-              dependencyNotes,
-              this.logger
-            ),
+            versionHeaderRegex: `\n###? v?[${version
```
kinyoklion, line 318: The base PR will include the version. The default regex would insert changes above the last version. Which makes it backward.

I need to come back and cleanup the "append" updater if the PR moves forward.

src/strategies/base.ts
```diff
@@ -254,19 +254,19 @@ export abstract class BaseStrategy implements Strategy {
     commits: ConventionalCommit[],
     latestRelease?: Release,
     draft?: boolean,
-    labels: string[] = []
+    labels: string[] = [],
+    bumpOnlyOptions?: BumpReleaseOptions
```
kinyoklion, line 258: I needed some indicator to bypass all the early out conditions associated with no commits or changelog entries.

An alternate approach to this whole thing may be to generate all the PRs always, and then prune ones we don't want.

src/plugins/node-workspace.ts
```diff
@@ -297,16 +315,17 @@ export class NodeWorkspace extends WorkspacePlugin<Package> {
...
+              .replace('.', '.')}]`,
+            changelogEntry: dependencyNotes,
           }),
         },
       ],
       labels: [],
-      headRefName: BranchName.ofTargetBranch(this.targetBranch).toString(),
+      headRefName:
+        basePullRequest?.headRefName ??
+        BranchName.ofTargetBranch(this.targetBranch).toString(),
```
kinyoklion, line 328: For #2109

kinyoklion:
I am working on the CLA situation.

src/strategy.ts
```diff
@@ -47,7 +51,8 @@ export interface Strategy {
     commits: Commit[],
     latestRelease?: Release,
     draft?: boolean,
-    labels?: string[]
+    labels?: string[],
+    bumpOnlyOptions?: BumpReleaseOptions
```
chingor13, line 55: I'm a bit unclear on what these options are for. Perhaps a docstring is good enough to describe what it's use is for.

chingor13 (changes requested):
Thank you for this and apologies for the late review.

In general I think this is a fine approach, although I don't fully understand the edge cases you are trying to handle.  Test case would demonstrate the primary scenario it's fixing and prevent regressions in the future. Test cases would also demonstrate the edge cases you're trying to handle.

kinyoklion:
> Thank you for this and apologies for the late review.
> 
> In general I think this is a fine approach, although I don't fully understand the edge cases you are trying to handle. Test case would demonstrate the primary scenario it's fixing and prevent regressions in the future. Test cases would also demonstrate the edge cases you're trying to handle.

Thank you @chingor13,

Now that I know the general approach is fine I will add some test cases that make this more clear.

The situation is not really an edge case. If you have any workspace dependent packages, then it doesn't generate change logs correctly or update files with version numbers.

Imagine you have two packages. Package A and package B. Package B depends on package A.

You commit a change which affects files in package A: "feat: Add my feature."

The release PR that gets generated will have an incorrect changelog. It will just append a "dependencies" section to the top of the changelog.

Example:
image

The second major problem is that if package B had any `extra-files`, then the version numbers in those files would not be updated.

Both of these are because version bumps only from dependencies do not run any of the steps associated with the base strategy being used.

Thank you,
Ryan

src/strategy.ts
kinyoklion: Yes, I will add some documentation. 

I wasn't 100% sure on this, but basically these serve as an indicator that you are creating a release with a version number update, even if there were no associated conventional commits. 

So it is:
A: An indicator that we want to make a release even though there are no calculated changes.
B: The version number we want for that release.

__snapshots__/node-workspace.js
```diff
@@ -138,6 +138,84 @@ Release notes for path: node1, releaseType: node
...
+
+
+<details><summary>@here/pkgA: 3.3.4</summary>
+
+Release notes for path: node1, releaseType: node
+</details>
+
+<details><summary>@here/pkgB: 2.2.3</summary>
+
+## [2.2.3](https://github.com/googleapis/node-test-repo/compare/pkgB-v2.2.2...pkgB-v2.2.3) (1983-10-10)
```
kinyoklion, line 153: This line here represents one of the primary differences. Without this change, there would be no header line here. Both in the PR and in the Changelog.

__snapshots__/node-workspace.js
```diff
@@ -138,6 +138,84 @@ Release notes for path: node1, releaseType: node
...
+
+---
+This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).
+`
+
+exports['NodeWorkspace plugin run includes headers for packages with configured strategies 2'] = `
+other notes
+`
+
+exports['NodeWorkspace plugin run includes headers for packages with configured strategies 3'] = `
```
kinyoklion, line 195: This demonstrates the correct header line being added to the changelog.

__snapshots__/node-workspace.js
```diff
@@ -138,6 +138,84 @@ Release notes for path: node1, releaseType: node
...
+exports['NodeWorkspace plugin run includes headers for packages with configured strategies 4'] = `
+### Dependencies
+
+* The following workspace dependencies were updated
+  * dependencies
+    * @here/pkgA bumped from 3.3.3 to 3.3.4
+`
+
+exports['NodeWorkspace plugin run includes headers for packages with configured strategies 5'] = `
+## 1.1.2 (2023-12-15)
```
kinyoklion, line 208: This also would not have been present. This doesn't include a diff, because it didn't have the previous version information included.

test/plugins/node-workspace.ts
```diff
@@ -372,6 +376,171 @@ describe('NodeWorkspace plugin', () => {
...
+
+      const newCandidates = await plugin.run(candidates);
+      expect(newCandidates).lengthOf(1);
+      const nodeCandidate = newCandidates.find(
+        candidate => candidate.config.releaseType === 'node'
+      );
+      expect(nodeCandidate).to.not.be.undefined;
+      const updates = nodeCandidate!.pullRequest.updates;
+
+      assertHasUpdate(updates, 'node2/my-file', Generic);
```
kinyoklion, line 542: Without invoking the base strategy we would not have the update for extra-files.

kinyoklion:
I've added some additional doc comments, added tests, and made some comments in the PR for those tests.

In this PR I am only focusing on node, but I believe that these same problems likely affect the other workspace plugins.

kinyoklion:
I found some situation where the changelog is still not correct, so I am investigating that and will update.

TimothyJones:
I am certain this would fix #2109. 

In addition to the issues you mentioned, I think this _might_ also fix #1946, or at least address one of the issues that contributes to it.

chingor13 (approved):

GregKruz (approved):

PR #2117, opened 2023-11-06
feat(deps): refactor node-workspace plugin to drop lerna-lite/core

### LINKED ISSUE #2116: Can't build action with release-please v16
Looks like something in `@lerna-lite/core` can be interpreted by `@vercel/ncc` which is used to build the single-file javascript action:

```
~/code/release-please-action$ npm run build

> release-please-action@3.7.12 build
> ncc build index.js

ncc: Version 0.38.0
ncc: Compiling file index.js into CJS
Error: Module parse failed: Unexpected token (26:23)
File was processed with these loaders:
 * ./node_modules/@vercel/ncc/dist/ncc/loaders/empty-loader.js
 * ./node_modules/@vercel/ncc/dist/ncc/loaders/relocate-loader.js
 * ./node_modules/@vercel/ncc/dist/ncc/loaders/shebang-loader.js
You may need an additional loader to handle the result of these loaders.
|     npm_lifecycle_event: event,
|     npm_lifecycle_script: cmd,
>     __webpack_require__.ab + "node-gyp.js",
|   })
| 
    at /home/chingor/code/release-please-action/node_modules/@vercel/ncc/dist/ncc/index.js.cache.js:38:1896222
    at /home/chingor/code/release-please-action/node_modules/@vercel/ncc/dist/ncc/index.js.cache.js:38:396262
    at _done (eval at create (/home/chingor/code/release-please-action/node_modules/@vercel/ncc/dist/ncc/index.js.cache.js:21:75523), <anonymous>:9:1)
    at eval (eval at create (/home/chingor/code/release-please-action/node_modules/@vercel/ncc/dist/ncc/index.js.cache.js:21:75523), <anonymous>:34:22)
```

chingor13:
Refactors the node-workspace plugin to  use the internal workspace-plugin graph/dependency logic and built-in
package.json updaters to update the content. We do not need to rely on lerna internals.

Fixes #2116 

feat: PackageJson updater can update dependencies

### DIFF
--- a/src/plugins/node-workspace.ts
+++ b/src/plugins/node-workspace.ts
@@ -12,20 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {PackageGraph} from '@lerna-lite/core';
-import {
-  Package as LernaPackage,
-  RawManifest as PackageJson,
-} from '@lerna-lite/core';
 import {GitHub} from '../github';
 import {CandidateReleasePullRequest, RepositoryConfig} from '../manifest';
 import {Version, VersionsMap} from '../version';
-import {RawContent} from '../updaters/raw-content';
 import {PullRequestTitle} from '../util/pull-request-title';
 import {PullRequestBody} from '../util/pull-request-body';
 import {ReleasePullRequest} from '../release-pull-request';
 import {BranchName} from '../util/branch-name';
-import {jsonStringify} from '../util/json-stringify';
 import {Changelog} from '../updaters/changelog';
 import {
   WorkspacePlugin,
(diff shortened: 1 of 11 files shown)


src/plugins/node-workspace.ts
```diff
@@ -130,16 +133,23 @@ export class NodeWorkspace extends WorkspacePlugin<Package> {
           packagePath,
           this.targetBranch
         );
-        packagesByPath.set(path, new Package(contents.parsedContent, path));
+        const packageJson: ParsedPackageJson = JSON.parse(
+          contents.parsedContent
+        );
+        const pkg: Package = {
```
bcoe, line 139: If you made this a class rather than just an interface, it would make the refactor easier to read.

test/plugins/node-workspace.ts
```diff
@@ -157,7 +189,7 @@ describe('NodeWorkspace plugin', () => {
             ),
           ],
         }),
-        buildMockCandidatePullRequest('node1', 'node', '2.2.2', {
+        buildMockCandidatePullRequest('node1', 'node', '3.3.3', {
```
bcoe, line 192: Why do the mock values need to change? It would be good to see the same tests running with the same assertions against the refactored code.

test/plugins/node-workspace.ts
```diff
@@ -230,10 +255,10 @@ describe('NodeWorkspace plugin', () => {
       );
       expect(nodeCandidate).to.not.be.undefined;
       const updates = nodeCandidate!.pullRequest.updates;
-      assertHasUpdate(updates, 'package.json');
-      assertHasUpdate(updates, 'node1/package.json');
-      assertHasUpdate(updates, 'node4/package.json');
       snapshot(dateSafe(nodeCandidate!.pullRequest.body.toString()));
+      snapshotUpdate(updates, 'package.json');
```
bcoe, line 259: curious why this order changed.

test/updaters/fixtures/package-with-dependencies.json
```diff
@@ -0,0 +1,52 @@
...
+		"command",
+		"cli",
+		"parsing",
+		"option",
+		"args",
+		"argument"
+	],
+	"author": "Ben Coe <ben@npmjs.com>",
+	"license": "ISC",
+	"devDependencies": {
```
bcoe, line 28: I like that you tested very type of dependency here (dev, optional, peer).

bcoe (commented):
LGTM, left a few questions.

test/plugins/node-workspace.ts
chingor13: The node1 package (`@here/pkgA`) fixture is already at 3.3.3. The lerna updater only updated the dependency versions, so the output still had 3.3.3 as the `@here/pkgA` version. But switching to the PackageJson updater, it actually did change the value from 3.3.3 to 2.2.2 (which was invalid). To keep the test output the same before and after the refactor, I switched this value to be `3.3.3`.

test/plugins/node-workspace.ts
chingor13: The snapshot name is based on an incrementing count. To make the snapshots only additive, I put these ones after the existing snapshot.

src/plugins/node-workspace.ts
bcoe: Discussed in chat, no need for this change.

bcoe (approved):
Glad to eliminate a dependency, good work.

PR #2143, opened 2023-11-27
fix: pass pull request header and footer options to merge plugin

### LINKED ISSUE #2142: Merge plugin does not use `pull-request-header` and `pull-request-footer` config values
#### Environment details

  - OS: `Darwin pieholden-m1.local 23.1.0 Darwin Kernel Version 23.1.0: Mon Oct  9 21:28:12 PDT 2023; root:xnu-10002.41.9~6/RELEASE_ARM64_T8103 arm64 arm Darwin`
  - Node.js version: `20.7.0`
  - npm version: `10.2.3`
  - `release-please` version: `16.3.0`

#### Steps to reproduce

  1. Set `pull-request-header` and `pull-request-footer` to any string value
  2. Set `separate-pull-requests` to `false`
  3. Use `release-please` to create a PR with multiple components
  4. Observe that the default header and footer are part of the PR

Thanks!

lukekarrys:

Fixes #2142  🦕

### Details

This works by looping through `repositoryConfig` and getting the first set value for both `pullRequestHeader` and `pullRequestFooter`. This means they could possibly come from different components in a manifest release.

### Alternate Solutions

I thought about only using config values from `repositoryConfig[ROOT_PROJECT_PATH]`. I didn't have strong feelings either way so I went with the solution I felt would be less surprising. But I'd be open to reworking the PR to this alternate solution.

### DIFF
--- a/__snapshots__/manifest.js
+++ b/__snapshots__/manifest.js
@@ -117,7 +117,7 @@ exports['Manifest buildPullRequests should handle mixing componentless configs 1
 This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).
 `
 
-exports['Manifest buildPullRequests should handle multiple package repository 1'] = `
+exports['Manifest buildPullRequests with multiple packages should handle multiple package repository 1'] = `
 :robot: I have created a release *beep* *boop*
 ---
 
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -33,7 +33,7 @@ import {
 } from './factory';
 import {Release} from './release';
 import {Strategy} from './strategy';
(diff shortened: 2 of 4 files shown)


chingor13 (changes requested):
Thanks for this! The feature seems reasonable, but this one has a bit more complex logic and could use a unit test to demonstrate the example and to ensure we don't regress in the future.

lukekarrys:
Thanks @chingor13! I added a test case where it grabs the header from the first package and the footer from the second package another, while also ignoring the footer from the second package. Let me know what you think.

lukekarrys:
I didn't realize I left the tests in a failing state. Just pushed a fixup that should be passing now.

chingor13 (approved):
Thanks!

PR #2153, opened 2023-12-04
feat: update VERSION file in PHP repo if it exists

bshaffer:
This updates the `php` default strategy to update a `VERSION` file in the root of the repo if it exists. This will be useful for some PHP libraries to track which versions are being used.

### DIFF
--- a/src/strategies/php.ts
+++ b/src/strategies/php.ts
@@ -17,6 +17,7 @@ import {Changelog} from '../updaters/changelog';
 // PHP Specific.
 import {RootComposerUpdatePackages} from '../updaters/php/root-composer-update-packages';
 import {BaseStrategy, BuildUpdatesOptions, BaseStrategyOptions} from './base';
+import {DefaultUpdater} from '../updaters/default';
 import {Update} from '../update';
 import {VersionsMap} from '../version';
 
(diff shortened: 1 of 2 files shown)


src/strategies/php.ts
```diff
@@ -67,6 +68,21 @@ export class PHP extends BaseStrategy {
...
+    const contents = await this.github.getFileContentsOnBranch(
+      this.addPath('VERSION'),
+      this.targetBranch
+    );
+
+    // update VERSION file
+    updates.push({
+      path: this.addPath('VERSION'),
+      createIfMissing: false,
+      cachedFileContents: contents,
```
bshaffer, line 80: I am not sure what this does, or if we need it. Since we already know the version, we don't really care what's written in the file. however, I noticed this is being used in other strategies, so I left it in.

src/strategies/php.ts
```diff
@@ -67,6 +68,21 @@ export class PHP extends BaseStrategy {
       }),
     });
 
+    const contents = await this.github.getFileContentsOnBranch(
```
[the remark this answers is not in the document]
bshaffer: Oh yes, good catch! We definitely don't want that. I've removed it.

src/strategies/php.ts
bshaffer: Thanks for the explanation!

chingor13 (approved):

PR #2162, opened 2023-12-11
feat: add root package-lock.json support

### LINKED ISSUE #1842: Update versions of linked packages in package-lock.json to match package.json

### LINKED ISSUE #2161: Node Monorepo - Root package-lock.json
#### Environment details

  - OS: Ubunutu 22.04.3
  - Node.js version: 18.18.0
  - npm version: 9.8.1
  - `release-please` version:  16.4.0

#### Steps to reproduce

  1. create a nodejs/npm monorepo
  2. create release please manifest + config
  3. inspect pull request of release please bot

The problem is that release-please only tries to update the lock file in the local package folder (e.g. packages/package-a/package-lock.json), but not the root lock file.

In the following pull request https://github.com/authup/authup/pull/1566/files, for example, I would assume that the version of the respective package that is to be updated is also updated in the root package-lock.json file

tada5hi:

Fixes #2161 
Fixes #1842


chingor13 (commented):
Rather than explicitly adding a check for a root `package-lock.json`, would it work if the `node-workspace` plugin respected the `considerAllArtifacts` option that other `workspace` plugins can use?

This option tells the plugin to scan for all "artifacts" (e.g. paths that have a `package.json`) even if they are undeclared as paths in the `release-please-config.json`. When building the dependency graph of artifacts that affect each other, the plugin would find the root `package.json` and create updates for it and its related `package-lock.json`.

tada5hi:
@chingor13 I am unsure, as far as I can tell with my knowledge so far, I would say that this is not the case. 

Another problem is that the workspace packages are not (necessarily) part of the root package.json file, but only part of the package-lock.json file. For example, the [docs](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md#releasing-root-path-of-library-) describe a scenario in which the root package is also published and references parts of the workspace packages. However, in most cases this is not the case, as the root package should also be published.

blaine-arcjet:
@chingor13 I just ran into this same problem. What would it take to get this across the finish line?

tada5hi:
@chingor13 ?

chingor13 (approved):

Issue #2189, opened 2024-01-04
[PHP] Support tag to show up in release notes for miscellaneous 


bshaffer:
We'd like to configure a tag that will allow release notes to appear for things like `deprecations` (see https://github.com/googleapis/google-auth-library-php/commit/8cfb95c04e036e744393a49d6ea404f62f0e0ffc) or other such changes that do not fall under `fix` or `feat`, but are still something we'd like to highlight in our release notes.

The proposal here would be `misc`, where we could do something like `misc(deprecation):` and have it show up in the release notes like this:

```md
### Miscellaneous

**deprecation**: Removed support for lib v2
```

chingor13:
Which changelog sections appear is configurable on a per-repository basis. Please use a `release-please-config.json` configuration and set the `changelog-sections` configuration option: https://github.com/googleapis/release-please/blob/f78f87b3ec986344c9265a9bd225b92a41aeac23/schemas/config.json#L31-L52

This should override the default configuration: https://github.com/googleapis/release-please/blob/f78f87b3ec986344c9265a9bd225b92a41aeac23/src/strategies/php.ts#L24-L36

bshaffer:
@chingor13 this is super helpful, thank you.

I think we DO want to override the base `php` strategy, so that the behavior is consistent across all our repositories

PR #2190, opened 2024-01-05
feat: update default php strategy and php-yoshi to match

bshaffer:
address https://github.com/googleapis/release-please/issues/2189

 - Have the same default configuration for conventional commits
 - Ensure "yoshi-php" strategy will also set the "version" field in `composer.json` if it exists

Question for @chingor13  - what does "hidden" mean in `yoshi-php`? What is the difference between `hidden` and omitting the entry entirely from `CHANGELOG_SECTIONS` in that context? It seems like those items don't show up in the release notes at all (for `google-cloud-php` releases)


src/strategies/php.ts
```diff
@@ -26,8 +26,9 @@ const CHANGELOG_SECTIONS = [
   {type: 'fix', section: 'Bug Fixes'},
   {type: 'perf', section: 'Performance Improvements'},
   {type: 'revert', section: 'Reverts'},
-  {type: 'chore', section: 'Miscellaneous Chores'},
-  {type: 'docs', section: 'Documentation', hidden: true},
+  {type: 'docs', section: 'Documentation'},
+  {type: 'misc', section: 'Miscellaneous'},
+  {type: 'chore', section: 'Chores', hidden: true},
```
chingor13, line 31: This is a breaking behavior change for other folks using the `php` strategy and this could be considered for a new semver major version.

chingor13 (changes requested):
The `-yoshi` strategies are internal (no good way to separate right now), undocumented for the public (not meant for external users) and we can make behavior breaking changes whenever.

The non-`-yoshi` strategies are in use by other teams and we should not make behavior breaking changes.

chingor13:
> Question for @chingor13 - what does "hidden" mean in `yoshi-php`? What is the difference between `hidden` and omitting the entry entirely from `CHANGELOG_SECTIONS` in that context? It seems like those items don't show up in the release notes at all (for `google-cloud-php` releases)

I think this is an artifact of the initial implementation. We pass that config to the conventional changelog notes builder. I don't think there's a difference between `hidden: true` and omitted.

src/strategies/php.ts
bshaffer: Makes sense. I just added `misc`, is something that is considered breaking?

bshaffer: @chingor13 would it be possible to merge / release this before Friday? The second part of this (updating the `version` field in `composer.json` for the monorepo) is a blocked. I can split it into a separate PR if you'd like!

chingor13: Please separate the `version` field change into a separate PR.

Yes, adding `misc` will cause any `misc: ` commits to start forcing a patch bump when it is not part of the spec by default. These sections are already configurable in your repository without affecting every other user of that strategy.

bshaffer: Yes, I understand. We just have a LOT of repositories and it'd be nice to be able to configure this in one place.

bshaffer:
Opened https://github.com/googleapis/release-please/pull/2192 instead

PR #2192, opened 2024-01-09
feat: set "version" if it's in composer.json for php-yoshi config

bshaffer:


bshaffer:
@chingor13 can we get this looked at? It will be a blocker once the longrunning library is updated (which I'm worried we'll miss)

bshaffer:
@chingor13 another ping on this please 🙏

chingor13 (approved):

PR #2196, opened 2024-01-16
docs: include use of x-version-update annotation in abitrary file documentation

mpeddada1:
As seen in https://github.com/googleapis/sdk-platform-java/pull/2351, the `x-version-update` annotation can also be used to update versions in arbitrary files using release-please.


chingor13 (commented):
This is tricky, because it's only available to the `java-yoshi` strategy. In general, the community shouldn't be using this strategy as it's custom to our internal needs.

At some point if/when we get dynamic plugins implemented, then we can pull all of our custom strategies into their own sub-library and document these nuances there.

Issue #2198, opened 2024-01-26
PHP broken with last update


bshaffer:
Changes in https://github.com/googleapis/release-please/pull/2192 broke [our current release](https://github.com/googleapis/google-cloud-php/pull/6982#pullrequestreview-1844857221)

I copied [this line]() over from the generic `php.ts`, because I observed this behavior was working as expected for the external libraries. However, there's a

PR #2201, opened 2024-01-26
feat: set 'version' of composer.json in RootComposerUpdatePackages

bshaffer:
addresses https://github.com/googleapis/release-please/issues/2198

Marked as a `feat` since it changes the behavior of `RootComposerUpdatePackage` by always updating `version` in `composer.json`. 

**IMPORTANT**: The behavior of the php strategy (`src/strategies/php.ts`) will **not** change, as `version` was already being set in the `versionsMap`. The only difference in the php strategy is in the case where `composer.json` includes a `replace` section: the "version" key will no longer be added to it. This is the desired behavior, as adding such a string to `replace` results in a composer exception.

### DIFF
--- a/src/updaters/php/root-composer-update-packages.ts
+++ b/src/updaters/php/root-composer-update-packages.ts
@@ -31,6 +31,11 @@ export class RootComposerUpdatePackages extends DefaultUpdater {
       return content;
     }
     const parsed = JSON.parse(content);
+    if (parsed['version']) {
+      let fromVersion: string = parsed['version'];
+      parsed['version'] = this.version.toString();
+      logger.info(`updating "version" from ${fromVersion} to ${this.version.toString()}`);
+    }
     if (this.versionsMap) {
       for (const [key, version] of this.versionsMap.entries()) {
         const toVersion = version.toString() || '1.0.0';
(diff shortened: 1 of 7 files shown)


src/updaters/php/root-composer-update-packages.ts
```diff
@@ -26,11 +26,17 @@ export class RootComposerUpdatePackages extends DefaultUpdater {
...
       logger.info('no updates necessary');
       return content;
     }
     const parsed = JSON.parse(content);
+    if (parsed['version']) {
+      const fromVersion: string = parsed['version'];
+      const toVersion = this.version.toString() || '1.0.0';
+      parsed['version'] = toVersion;
+      logger.info(`updating "version" from ${fromVersion} to ${toVersion}`);
+    }
```
chingor13, line 39: Can you add tests for setting this root `version` entry (with the value set and unset)?

chingor13 (changes requested):

src/updaters/php/root-composer-update-packages.ts
bshaffer: I believe this is the kind of test you're looking for? https://github.com/googleapis/release-please/pull/2201/commits/5093438d6e8b111f4a8a75213109b641c58d93fd

chingor13 (approved):

PR #2203, opened 2024-01-26
Support multiple stable release branches / consider all branches when looking for previous release shas

### LINKED ISSUE #2202: Consider tags on separate release branches / allow multiple stable release branches
`release-please` currently expects all your release commits to be on the same branch you are releasing from. This doesn't work when you have multiple, long-lived release branches.

Our release process works as follows:
* We work off `main`
* When it's time to release a new version (major or minor) we create a branch off `main` at the feature-set we want to release
* Bug-fixes and security patches for release versions are later cherry-picked into these "release" branches (after being committed to main)

We are usually supporting at least the last 2 minors with patch releases. Using the `--target-branch` feature of `release-please`, we are able to release from these release branches, but since the release tags are on a commit on theses branches (which is desired), future releases on other branches can't find them. As a result, `release-please` doesn't find the correct previous release sha to compare to.

I'd like an option, like `--consider-all-branches` that when specified, does not limit the search of commits for released shas to the current branch. Of course, we don't want to look at every commit ever in the repo, so in this flow I suggest release please uses github's graphql api for comparing commits between two shas.

I tried writing a custom GitHub action that brought in a lot of release-please code as a library, but that was a much more code to maintain than the changes necessary in `release-please` to support them.
 
**Additional context**
I have an initial solution that I will PR after creating this issue. It's a draft at the moment, as it needs some testing, but I want to get buy in on the approach before investing more time.

trevorwhitney:

Fixes #2202 🦕

We are currently using this fork for our release process and it appears to be working as expected, but I'm sure there are edge cases that aren't being addressed yet.

This PR add a flag `--consider-all-branches` that, when specified, will use GitHub's compare commits API to compare the commits between head of the branch being released, and the found sha of the previous release, allowing the previous release's sha to be on a different branch. This allows, for example, to have a branch like `release-1.7.x`, with the latest release being, say, `1.7.4`, and then release a new minor version (`1.8.0`) from a branch like `release-1.8.x`, and have it correctly show the changelog of commits since `1.7.4`, even though that tag is on a different branch.

This PR isn't complete yet, and there's a lot of test coverage missing, but I wanted to get consensus on the approach before investing more time. Thanks!

### DIFF
--- a/src/bin/release-please.ts
+++ b/src/bin/release-please.ts
@@ -73,6 +73,8 @@ interface VersioningArgs {
   latestTagVersion?: string;
   latestTagSha?: string;
   latestTagName?: string;
+
+  includeAllReleases?: boolean;
 }
 
 interface ManifestConfigArgs {
@@ -345,6 +347,12 @@ function pullRequestStrategyOptions(yargs: yargs.Argv): yargs.Argv {
       describe: 'Override the detected latest tag name',
       type: 'string',
     })
+    .option('include-all-releases', {
+      describe:
+        'Include release shas from all branches when determining latest release',
+      default: false,
+      type: 'boolean',
+    })
     .middleware(_argv => {
       const argv = _argv as CreatePullRequestArgs;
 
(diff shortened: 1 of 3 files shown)


trevorwhitney:
We ended up forking to get this behavior. If there is interest in it please let me know and I can re-open.

Achllle:
Interested! Surprised this isn't supported (or documented) by default @trevorwhitney would also be interested in your fork if it's public.

PR #2222, opened 2024-02-12
feat: provide config `extends` feature

### LINKED ISSUE #1979: Could I extend a universal configuration?
Hi, I'm new to here, I want to `extend` some common configuration for projects, i.e. the `changelog-sections`, is there an option for this?
I've read the [config.json](https://github.com/googleapis/release-please/blob/main/schemas/config.json), but didn't find such option.

I wish to configure like the following.

`https://example.org/changelog-sections.json`:

```json
{
    "changelog-sections": [
        { "type": "feat", "section": "Features ✨" },
        { "type": "feature", "section": "Features ✨" },
        { "type": "fix", "section": "Bug Fixes 🐞" }
    ]
}
```

`release-please-config.json`:

```json
{
  "extends": [
    "https://example.org/changelog-sections.json"
  ]
}
```

### LINKED ISSUE #2048: FR: pull from external config
As a developer, I would like to have a single, centralized configuration for all projects in my GH org that are using release-please as a GitHub action. There should also be an option to extend this configuration on a per-repo basis, in case certain adjustments need to be made.

[Renovate](https://github.com/renovatebot/renovate/) has a good [example](https://github.com/renovatebot/renovate/blob/main/docs/usage/config-presets.md) of how this is implemented in its service. Ideally, I could use an `extends` field in my `release-please-config.json`:
```json
{
  "extends": "MyOrgName/.github",
  "packages": { ".": {} },
  "release-type": "node"
}
```

this would merge the individual repo's configuration with the configuration found in my organization's `.github` repo, overwriting the parent's configuration with the child's values.

antongolub:
This PR provides config extensions with github-hosted assets:

```json
{
  "extends": "owner/repo/main",
  "packages": {
    "packages/bot-config-utils": {
    }
  }
}
```

Fixes #2048 🦕
Fixes #1979


src/util/http-api.ts
```diff
@@ -0,0 +1,37 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as https from 'https';
+
+// Hmmm... no-experimental-fetch flag is used. Ok
```
chingor13, line 17: This flag is used in test because we've been using `nock` to fake http requests and it can't handle the built-in fetch :(

test/fixtures/manifest/config/config-with-remote-extends.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "owner/repo/main",
```
chingor13, line 2: Perhaps:
* `owner/repo@branchOrSha:path/to/file.json`
* `owner/repo/path/to/file.json@branchOrSha`
* `owner/repo:path/to/file.json@branchOrSha`

Or something similar to https://docs.renovatebot.com/config-presets/#github

chingor13 (changes requested):
Thanks for this pull request and this general feature seems useful!

However, we're aiming to be conservative with the number of external dependencies we take, and config management hasn't been enough of a pain point that we want to outsource this logic yet.

To me, it seems like we need a way to fetch the remote files (included in this PR) and a `deepMerge` implementation (~15 LoC, rather than the external dependency).

test/fixtures/manifest/config/config-with-remote-extends.json
antongolub: We can combine. Added support for _renovate-like_ format and for the last two options.

src/util/http-api.ts
antongolub: Ok. Is it necessary to mock the transport layer? Maybe `http.getJson` mock would suffice here?

antongolub:
> ~15 LoC, rather than the external dependency

Once the 15loc deepmerge meets the prod area, it turns into [115loc deepmerge](https://www.npmjs.com/package/deepmerge). And another 1115 lines of tests. I know this road; I have walked along it many many times. I'd even suggest using [yargs/helpers#applyExtends](https://github.com/yargs/yargs/blob/main/lib/utils/apply-extends.ts) here, but unfortunately there is no way to redefine its internal loader.

Dependencies are always a trade-off between complexity, risks and implementation costs. For example, someone's reinventing cosmiconfig in place ([ref1](https://github.com/cosmiconfig/cosmiconfig/blob/main/src/defaults.ts#L10), [ref2](https://github.com/TypeStrong/typedoc/blob/eb7510b26a2a4b18cfddb76bdd0fe0d1ed2702a2/src/lib/utils/options/readers/typedoc.ts#L165), [ref3](https://github.com/prettier/prettier/blob/main/src/config/prettier-config/config-searcher.js#L5)), while `release-please` uses [yaml@^2.2.2](https://github.com/googleapis/release-please/blob/main/package.json#L98) and [js-yaml@^4.0.0](https://github.com/googleapis/release-please/blob/main/package.json#L88) as prod deps as the same time :)

Issue #2125, opened 2023-11-09
Support `extra-files` globs for generic updater


t3mi:
**Is your feature request related to a problem? Please describe.**
Please add support for extra-files globes for all file types (.md in particular)

 **Describe the solution you'd like**
```
...
  extra-files:
    **/README.md
...
```

 **Describe alternatives you've considered**
Current way is to specify every file path which needs to be handled for version updates.

 **Additional context**
ref #1619

chingor13:
This seems pretty reasonable. In #1619, we already support globs for explicit extra file types (e.g. xml or json files). We could probably have a config type for `generic`:

```json
{
  "extra-files": [
    {
      "type": "generic",
      "glob": "**/README.md"
    }
  ]
}
```

OR 

```json
{
  "extra-files": [
    {
      "type": "generic",
      "path": "explicit/path/to/README.md"
    }
  ]
}
```

t3mi:
Fixed by #2227

PR #2227, opened 2024-02-17
fix: Support generic updater

### LINKED ISSUE #2205: Allow forcing the use of the generic updater in `extra-files`
Since #2072, any JSON, XML, YAML or TOML file is updated using their corresponding updater, meaning that the value of a top level `version` key, if existing, is bumped. It should be possible to override this default behavior.

My use case is to bump a docker image tag in a `docker-compose.yaml` file, that contains a top-level `version` key, like such:
```yaml
# docker-compose.yaml
version: "3.8"  # Docker version: https://docs.docker.com/compose/compose-file/04-version-and-name/

services:
  my-service:
    ...
    image: my-org/my-repo:v0.1.0 # x-release-please-version
```
I tried two things:

```yaml
"extra-files": ["docker-compose.yaml"]
```
When using this config, the `version: "3.8"` is replaced by `version: v0.1.1`, and the `image` value is left untouched.

```json
"extra-files": [
  {
    "type": "yaml",
    "path": "docker-compose.yaml",
    "jsonpath": "services.my-service.image"
  }
]
```
When using this second config, the `image` value is transformed to `v0.1.1`, deleting the `my-org/my-repo:` part.

Allow something like this, that won't attempt to modify the top-level `version`:
```json
"extra-files": [
  {
    "type": "generic",
    "path": "docker-compose.yaml"
  }
]
```

As specified [there](https://docs.docker.com/compose/compose-file/04-version-and-name/), the `version` key is not strictly necessary for a docker-compose file, so removing this key leads to the desired behavior. However, this is only a temporary workaround.

regseb:

Fixes #2205 🦕

### DIFF
--- a/docs/customizing.md
+++ b/docs/customizing.md
@@ -182,6 +182,21 @@ You can annotate a block by starting with a line containing:
 and close the block with a line containing `x-release-please-end`. Within
 the block, we will attempt to replace version values.
 
+Default updaters are applied depending on the file extension. If you want to
+force the [Generic](/src/updaters/generic.ts) updater, you must use type
+`"generic"`.
+
+```json
+{
+  "extra-files": [
+    {
+      "type": "generic",
+      "path": "path/to/file.yml"
+    }
+  ]
+}
+```
+
 ## Updating arbitrary JSON files
 
 For files with the `.xml` extension, the `version` property is updated.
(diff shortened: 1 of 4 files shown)


jyoder-deloitte:
Also addresses #2125.  It might be nice to have a `glob: true` example as well.

chingor13 (approved):
Apologies for the late review.

We can adjust the json schema definition in a follow-up PR.

PR #2270, opened 2024-04-16
fix: Disable updating peer dependencies when updatePeerDependencies is false.

### LINKED ISSUE #2269: Peer dependencies are updated regardless of "updatePeerDependencies" setting.
#### Environment details

  - OS: Linux
  - Node.js version: 18.19.0
  - npm version: 10.2.3
  - `release-please` version: 16.10.1

#### Steps to reproduce

  1. Use the node workspace plugin with at least 2 packages. For these instructions package A and package B.
  2. Package B has a peer and dev dependency on package A.
  4. Make a change to package A with a commit that results in a releasable unit.
  5. The package.json of B will have its peer dependency updated.

Example PR where this happened: https://github.com/launchdarkly/js-core/pull/418/files

Specific file:
https://github.com/launchdarkly/js-core/pull/418/files#diff-9838b4949cf31603774a130ab3f492f4afcb6f430996e05ac6c778692f45a10b

It is a common practice to have a peer dependency and a dev dependency on another package. This allows you to unit test with the package you use as a peer dependency. In the linked example the `node-server-dynamo` package should be tested with the latest package, but the peer dependency should be a version range.

kinyoklion:
The "updatePeerDependencies" option currently operates as a "considerPeerDependencies" option. What this means is that if a package only has a peer dependency on another, then that will not be considered and no update will be made (as if there was no dependency). If a package has a peer dependency as well as another dependency, for instance a dev dependency, then the peer dependency is updated regardless.

This propagates the "updatePeerDependencies" option to the PackageJson updater so that it can omit updates to the peer dependencies.

It isn't completely clear what the intended behavior here is. This update behavior matches my current use-case, but it is possible it could be a problem for others. In which case there could be two different options (but the current name does seem like it would impact peer dependency updates.)

- [na] Appropriate docs were updated (if necessary)

Fixes #2269
 🦕

### DIFF
--- a/__snapshots__/package-json.js
+++ b/__snapshots__/package-json.js
@@ -1,3 +1,59 @@
+exports['PackageJson updateContent does not update peer dependencies by default 1'] = `
+{
+\t"name": "yargs-parser",
+\t"version": "14.0.0",
+\t"description": "the mighty option parser used by yargs",
+\t"main": "index.js",
+\t"scripts": {
+\t\t"test": "nyc mocha test/*.js",
+\t\t"posttest": "standard",
+\t\t"coverage": "nyc report --reporter=text-lcov | coveralls",
+\t\t"release": "standard-version"
+\t},
+\t"repository": {
+\t\t"url": "git@github.com:yargs/yargs-parser.git"
+\t},
+\t"keywords": [
+\t\t"argument",
+\t\t"parser",
+\t\t"yargs",
+\t\t"command",
+\t\t"cli",
+\t\t"parsing",
+\t\t"option",
+\t\t"args",
+\t\t"argument"
+\t],
+\t"author": "Ben Coe <ben@npmjs.com>",
+\t"license": "ISC",
+\t"devDependencies": {
+\t\t"chai": "^4.2.1",
+\t\t"coveralls": "^3.0.2",
+\t\t"mocha": "^5.2.0",
+\t\t"nyc": "^13.0.1",
+\t\t"standard": "^12.0.1"
+\t},
+\t"dependencies": {
+\t\t"camelcase": "^6.0.0",
+\t\t"decamelize": "^1.2.0"
+\t},
+\t"optionalDependencies": {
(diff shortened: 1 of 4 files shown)


__snapshots__/package-json.js
```diff
@@ -1,3 +1,59 @@
...
+\t},
+\t"dependencies": {
+\t\t"camelcase": "^6.0.0",
+\t\t"decamelize": "^1.2.0"
+\t},
+\t"optionalDependencies": {
+\t\t"foo": "~0.1.0"
+\t},
+\t"peerDependencies": {
+\t\t"bar": ">= 1.0.0"
```
kinyoklion, line 44: Test that validates the peer dependency is not updated.

test/updaters/package-json.ts
```diff
@@ -36,6 +36,25 @@ describe('PackageJson', () => {
...
+      );
+      const versionsMap: VersionsMap = new Map();
+      versionsMap.set('camelcase', Version.parse('6.0.0'));
+      versionsMap.set('chai', Version.parse('4.2.1'));
+      versionsMap.set('foo', Version.parse('0.1.0'));
+      versionsMap.set('bar', Version.parse('2.3.4'));
+      const packageJson = new PackageJson({
+        version: Version.parse('14.0.0'),
+        versionsMap,
+        updatePeerDependencies: true,
```
kinyoklion, line 51: This test implicitly dependended on the updater always updating the peer dependencies. So it was updated and produces the same output.

test/updaters/package-json.ts
```diff
@@ -36,6 +36,25 @@ describe('PackageJson', () => {
...
+      });
+      const newContent = packageJson.updateContent(oldContent);
+      snapshot(newContent.replace(/\r\n/g, '\n'));
+    });
+
+    it('does not update peer dependencies by default', async () => {
```
kinyoklion, line 57: New test which does not update peer dependencies.

chingor13 (approved):
Thank you!

Issue #1858, opened 2023-02-20
separate-pull-requests + node-workspace plugin issues


monholm:
While trying to use `release-please` with `separate-pull-requests` and the `node-workspace` plugin enabled, I stumbled upon a couple of issues and some odd behaviour.

When pushing a commit for a single (and at this point in the repro, the only) package, in addition to the `chore(main): release <package-name> x.y.z` pr, `release-please` opens another pr (`chore: release main`) with 0 files changed. The `chore: release main` pr is open constantly and contains 0 files changed. 

Is this expected behaviour? If so, how is one supposed to treat this PR? Leave it as is, merge it before or after the pr containing actual changes, or? I was unable to find any documentation regarding this, so I'm guessing this might be a bug? :)

Moving on from there, using the same setup, if one updates a package that's referenced in another package, the `chore: release main` pr is updated with the patch bump change for the package that references the updated package. But the `changelog` change is invalid and if one tries to merge the PR, no tag is created for the patch bump, causing the versioning to fail. I've created a [repro](https://github.com/Simon-TechForm/rp-bug-triage), as this is quite difficult describing... 😅

#### Environment details

  - OS: Ubuntu 22.04
  - Node.js version: v18.14.1
  - npm version: v9.3.1
  - `release-please` version: v15.7.0

#### Steps to reproduce

  1. Setup basic manifest-based project
  2. Enable `separate-pull-request` and `node-workspace` plugin
  3. Add single package and push commit to trigger first release
  4. Observe two PR's being opened, one of which is empty
  5. Add another package and make it depend on the first package.
  6. Update the first package and observe the invalid changelog added to the previously empty PR
  7. If the PR mentioned in step 6 is merged, no tag is created for the bumped package

Again, I've created a [repro](https://github.com/Simon-TechForm/rp-bug-triage) that hopefully makes it easier to grasp. If not, let me know if I can help provide any other info :)

lorentzfb:
I have the same issues

shinebayar-g:
Having an exact same issue. release-please is creating additional empty PR besides the actual package PR.
<img width="1076" alt="image" src="https://github.com/googleapis/release-please/assets/3091558/ea61c83c-b6fb-45ee-8b3b-e60f8d029e2f">

release-please-config.json
```json
{
  "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json",
  "bootstrap-sha": "942d93a2423b28394b427eabdae1b2cc7a174742",
[...]

PR #2310, opened 2024-06-12
fix: make workspace plugin with `separate-pull-request` fine

Hajime-san:

Fixes https://github.com/googleapis/release-please/issues/1858 🦕

If the documentation needs updating, please let me know.


src/factories/plugin-factory.ts
```diff
@@ -63,6 +65,7 @@ const pluginFactories: Record<string, PluginBuilder> = {
       {
         ...options,
         ...(options.type as WorkspacePluginOptions),
+        merge: determineMerge(options),
```
chingor13, line 68: This will override the `merge` option provided by the user (`WorkspacePluginOptions` has an optional `merge`).

Also, it seems weird for `determineMerge` to be checking to see which type of plugin is being configured. `determineMerge` is already called in a block chosen based on the plugin type. We should just explicitly set the default in the plugin factory for each type.

src/factories/plugin-factory.ts
```diff
@@ -38,6 +39,7 @@ export interface PluginFactoryOptions {
   targetBranch: string;
   repositoryConfig: RepositoryConfig;
   manifestPath: string;
+  separatePullRequests: boolean;
```
chingor13, line 42: This should probably be optional and have a default value. The ability to register plugins and pluginFactories is part of the public API (`PluginFactoryOptions`). It's unlikely that someone is providing their own `PluginFactoryOptions` instance, but a default wouldn't hurt.

chingor13 (changes requested):

Hajime-san:
Thank you for checking @chingor13 !
Could you re-review the new change that is based on your pointing out?

chingor13 (approved):

PR #2321, opened 2024-06-25
docs: add guide for faster test debugging  with ts-node

bukowa:

Debugging tests can be much faster without compiling to `js`.


chingor13:
Thank you, but I don't think documentation belongs here. If we need this, then we should just install it as a dev dependency for everything

PR #2336, opened 2024-07-13
feat: add additional-paths property to manifest releaser

### LINKED ISSUE #1921: Ability to define multiple paths

I am missing option to define paths whićh may be considered to scan commits. E.G. i have project with structure
- core
- features
- shared
- demo
i want scant paths /core, /features, /shared and exclude /demo. 
Also with monorepo approach:
apps
* app1
* app2

libs
* lib1
* lib2
* shared

app1 uses code from lib1, shared, app2 uses code from lib2, shared. Theese libs are not standalone libs, it is just separation of code from app. So i do not need track changes for every separate lib, but i want include int with app. So in manifest i have only app1 and app2 

path should accept multiple paths separated by comma

I tried many trick like filtering repository using monorepo dependency tree and also another versoning plugins fox NX monorpo. But release-please fits best with PR approach, except this single feature missing

Another solution: let current attribute path as is and add another one: additionalPaths: string[] which accepts another paths from root of repository

 **Additional context**

jack-lewin:

Fixes #1921 🦕

---

This PR adds an `additional-paths` property to the manifest releaser.

Additional paths are useful for monorepos where a package might depend files from other directories.

For example, if we want to trigger a release in `apps/my-app-2` when commits are found in `libs/my-lib`:

```
apps/
    my-app-1
    my-app-2
libs/
    my-lib
```

```jsonc
{
  "packages": {
    "apps/my-app-1": {},
    "apps/my-app-2": {
      "additional-paths": ["libs/my-lib"] // additional-paths would be a new config option
    }
  }
}
```

Best reviewed commit-by-commit.

### DIFF
--- a/__snapshots__/cli.js
+++ b/__snapshots__/cli.js
@@ -190,7 +190,7 @@ Options:
                                     the first major release
                                                       [boolean] [default: false]
   --prerelease-type                 type of the prerelease, e.g., alpha [string]
-  --extra-files                     extra files for the strategy to consider
+  --extra-files                     extra files for the strategy to update
                                                                         [string]
   --version-file                    path to version file to update, e.g.,
                                     version.rb                          [string]
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -64,7 +64,7 @@ Extra options:
 | `--pull-request-title-pattern` | `string` | Override the pull request title pattern. Defaults to `chore${scope}: release${component} ${version}` |
 | `--pull-request-header` | `string` | Override the pull request header. Defaults to `:robot: I have created a release *beep* *boop*` |
 | `--pull-request-footer` | `string` | Override the pull request footer. Defaults to `This PR was generated with Release Please. See documentation.` |
-| `--extra-files` | `string[]` | Extra file paths for the release strategy to consider |
+| `--extra-files` | `string[]` | Extra file paths for the release strategy to update |
 | `--version-file` | `string` | Ruby only. Path to the `version.rb` file |
 
 ## Creating/updating release PRs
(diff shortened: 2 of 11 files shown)


jack-lewin:
Hi @chingor13, sorry for the nudge - are you still maintaining this project?
If not, is there anyone else I should reach out to?

chingor13 (changes requested):
Thanks for this and sorry for the delay in reviewing.

This mostly looks fine although we should explicitly define what the expected logic is if an additional path coincides with a different component's path or additional paths. Currently, we sort by the components' path length with the longest first as a way to prefer the deepest path in the case of overlapping paths. With this new logic, we are still sorting based on the main path, and preferring its additional paths over any other ones.

Instead, we may need to invert the mapping and have the keys be all the "paths" (main & additional) and the value is the main component path to associate to.

sylwit:
Hi @jack-lewin will you have time to update your PR do you need help ? Thank you for this PR

roivanov:
sync with the latest main: https://github.com/jack-lewin/release-please/pull/1

pkazi:
Any plans to merge this feature?

gtoph:
I've been searching for the ability to have multiple components share a common directory such that any updates to common code would trigger a PR request to all associated components. Looks like this change is exactly what is needed.

@chingor13 any hope of getting this in anytime soon?

stan-stately:
please merge this. we really need it. the main reason most people use a monorepo is to share dependencies between packages

stan-stately:
I pulled main and implemented a fix for @chingor13's suggestion here: https://github.com/googleapis/release-please/pull/2534

jack-lewin:
Thank you @stan-stately for picking this up!

@chingor13, we'd really appreciate a review on #2534 - I think it makes sense to focus on that branch since it's more up-to-date & addresses your previous suggestion.

PR #2337, opened 2024-07-14
feat: add "always-update" config option

### LINKED ISSUE #1459: Update release PR on every change (instead of only changes that modify the changelog)

We've been using release-please to manage our release and changelog. We've noticed that the release PR is only updated if the PR includes a commit with either `fix:` or `feat:`.

Sometimes we have PRs that add tests or fix problems with our release workflows. We'd like to include those commits in the release branch, but we don't want them to be included in the changelog (because our users aren't interested in those changes).

I wasn't able to find any option to make this happen, so I'm assuming it's not possible right now.

I'd like to be able to configure release-please so that all changes that go into main are included in the release PR, but only the categories that are "not hidden" are added to the changelog.

Either a top-level setting, or a new field in the `changelog-sections` would work fine for us.

Not sure about any alternatives.

Thank you for your consideration!

meyfa:
This patch adds a new option "always-update" that forces existing pull requests to be updated even when there are no changes to the release notes. This is useful, for example, when the repository enforces pull requests to be up-to-date with the main branch before they can be merged. Without this option set to `true`, that is, with the default behavior of release-please, if the last commit made to the main branch does not appear in the release notes, the existing pull request branch would not be rebased.

---

Fixes #1459 🦕

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -117,6 +117,7 @@ export interface ReleaserConfig {
   pullRequestFooter?: string;
   tagSeparator?: string;
   separatePullRequests?: boolean;
+  alwaysUpdate?: boolean;
   labels?: string[];
   releaseLabels?: string[];
   extraLabels?: string[];
@@ -176,6 +177,7 @@ interface ReleaserConfigJson {
   'pull-request-header'?: string;
   'pull-request-footer'?: string;
   'separate-pull-requests'?: boolean;
+  'always-update'?: boolean;
   'tag-separator'?: string;
   'extra-files'?: ExtraFile[];
   'version-file'?: string;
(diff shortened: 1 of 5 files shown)


GregKruz (approved):

src/manifest.ts
```diff
@@ -174,6 +175,7 @@ interface ReleaserConfigJson {
   'pull-request-header'?: string;
   'pull-request-footer'?: string;
   'separate-pull-requests'?: boolean;
+  'always-update'?: boolean;
```
chingor13, line 178: I think this one should be moved to `ManifestConfig` - `ReleaserConfigJson` are the per-component configurations and ` always-update` seems to be a repository-wide configuration.

chingor13 (changes requested):
Sorry for the delay in reviewing this.

Looks good, except we can make the `always-update` option only a top-level manifest configuration option (in the types).

zirkelc:
Hey @meyfa I'm really looking forward to this feature. Is there anything I can do to help, in case you're busy with other things at the moment?

PR #2350, opened 2024-08-07
fix: fix some links point to github.com when using github enterprise

### LINKED ISSUE #2351: When using release-please on GitHub enterprise, some links at changelog/release points to github

Please provide the following details.

When using release-please-action v4 on github enterprise (e.g https://self-host.github.example.com),  changelog-host needs to be configured in release-please-config. 

#### Environment details

  - OS:
  - Node.js version:
  - npm version:
  - `release-please` version: 16.12.0 (release-please-action)

#### Steps to reproduce

  1. ?
  2. ?

Thanks!

goatwu1993:

Fixes #2351 🦕

### DIFF
--- a/src/util/github-server.ts
+++ b/src/util/github-server.ts
@@ -1,0 +1,1 @@
+export const DEFAULT_GITHUB_SERVER_URL = process.env.GITHUB_SERVER_URL || 'https://github.com';
--- a/src/changelog-notes/default.ts
+++ b/src/changelog-notes/default.ts
@@ -18,12 +18,12 @@ import {
   BuildNotesOptions,
 } from '../changelog-notes';
 import {ConventionalCommit} from '../commit';
+import { DEFAULT_GITHUB_SERVER_URL } from '../util/github-server';
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const conventionalChangelogWriter = require('conventional-changelog-writer');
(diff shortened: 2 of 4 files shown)


goatwu1993:
@chingor13 please take a look?

src/util/github-server.ts
```diff
@@ -0,0 +1 @@
+export const DEFAULT_GITHUB_SERVER_URL = process.env.GITHUB_SERVER_URL || 'https://github.com';
```
chingor13, line 1: We would like any environment variable handling to happen at the entrypoint of running release-please and injecting the configuration into the classes rather than pulling from the environment in various spots.

In this case, can we configure/pull this from the `GitHub` class?

chingor13 (changes requested):
Thanks for this and sorry for the long delay

src/util/github-server.ts
goatwu1993: maybe this should not happend in release-please but only in release-please-action while merging default...

Pjrich1313 (approved):

goatwu1993:
@chingor13 updated.

please review again

NOTE: this has some duplication with the input changelog host and i am not sure if this two input should be aliased

ferrarimarco:
Hi @goatwu1993. Can you please have a look at updating tests in `test/cli.ts`?

PR #2358, opened 2024-08-15
feat: handle multiple commits in a single message

### LINKED ISSUE #2357: Bug: splitMessages function fails to handle multiple conventional commits
Thank you to the creators of `release-please` for your hard work and dedication to the project.

I wanted to bring to your attention an issue with the current implementation of the parsing logic in `src/commit.ts`. It appears that the logic does not handle multiple conventional commits within a single squash merge correctly. The implementation seems to rely heavily on specific formatting and structure that does not fully align with the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#specification).

For example, problems arise when commit messages include bullet points, empty lines, or varying indentations, as demonstrated in the experiment examples provided. Additionally, the documentation does not mention the requirement for body messages to be indented by two spaces, as seen in some examples in the repository's README. This discrepancy suggests that the current logic may not fully adhere to the specification, which the repository aims to support. 

Here is the conventional commit format which I believe the examples below all fit in.

```
<type>[optional scope]: <description>

[optional body]

[optional footer(s)]
```
Here are some examples from the experiments I made

<details>
<summary>Slightly modified example of the example in the <a href="https://github.com/googleapis/release-please?tab=readme-ov-file#what-if-my-pr-contains-multiple-fixes-or-features">README</a> - Parsed correctly, however there needs to be a new line between body and title</summary>

```
feat: adds v4 UUID to crypto

This adds support for v4 UUIDs to the library.

fix(utils): unicode no longer throws exception
- some more stuff

feat(utils): update encode to support unicode
- does stuff
- more stuff
```

</details>

<details>
<summary>Same example except this time with bullets are at the body (Parsed Incorrectly)</summary>

```
feat: adds v4 UUID to crypto

This adds support for v4 UUIDs to the library.

fix(utils): unicode no longer throws exception

- some more stuff

feat(utils): update encode to support unicode

- does stuff
- more stuff
```

</details>

<details>
<summary>Same example with an empty line between the description and the body (Parsed Incorrectly - only the first commit is treated as a conventional commit)</summary>

```
feat: adds v4 UUID to crypto

This adds support for v4 UUIDs to the library.

fix(utils): unicode no longer throws exception

- some more stuff

feat(utils): update encode to support unicode

- does stuff
- more stuff
```

</details>

<details>
<summary>Multiple commits with a simple body (Parsed Incorrectly - only the first commit is treated as a conventional commit)</summary>

```
feat: adds v4 UUID to crypto

This adds support for v4 UUIDs to the library.

fix(utils): unicode no longer throws exception

some more stuff

feat(utils): update encode to support unicode

does stuff
more stuff
```

</details>

#### Additional Details

- I found out that the `BEGIN_NESTED_COMMIT` and `END_NESTED_COMMIT` could be used to customize the nested commits, however I think it would be awesome if release please can handle these automatically.
- I feel like the `splitMessages` function could be modified to support the aforementioned commits and improve the parsing. A simple change like below supports the commit types I mentioned above.(It brakes only 5 tests)

```typescript
function splitMessages(message: string): string[] {
  const parts = message.split('BEGIN_NESTED_COMMIT');
  const messages = [parts.shift()!];
  for (const part of parts) {
    const [newMessage, ...rest] = part.split('END_NESTED_COMMIT');
    messages.push(newMessage);
    messages[0] = messages[0] + rest.join('END_NESTED_COMMIT');
  }

  // Split the first message into conventional commits
  const conventionalCommits = messages[0].split(/\n(?=(?:feat|fix|docs|style|refactor|perf|test|build|ci|chore|revert)(?:\(.*?\))?: )/);

  return [...conventionalCommits, ...messages.slice(1)];
}

```

#### Environment details

  - OS: macos sonoma 14.5
  - Node.js version: v22.2.0
  - npm version: 10.7.0
  - `release-please` version: 16.12.0

#### Steps to reproduce

  1. Create a new txt file from the examples above in the `test/fixtures/commit-messages/` directory
  2. Create a new test in `test/commits.ts` with the following content

  ```typescript
    it('test custom commits', async () => {
    const commits = [buildCommitFromFixture('name-of-the-fixture')];
    const conventionalCommits = parseConventionalCommits(commits);
    console.log('Conventional Commits:');
    console.log(JSON.stringify(conventionalCommits, null, 2));
    expect(conventionalCommits).lengthOf(1);
    expect(conventionalCommits[0].breaking).to.be.true;
    expect(conventionalCommits[0].message).not.include('I should be removed');
  });
  
  3. Run `npm test` > out.txt
  5. Search the `out.txt` file for "Conventional Commits:" to inspect the conventionalCommits array.

  ```

dgokcin:
## Summary
This PR enhances the breaking change detection mechanism and improves the parsing of commit messages to handle multiple conventional commits and nested commit blocks mentioned in #2357 

## Changes
- Added additional breaking change detection from the commit body in `toConventionalChangelogFormat` and `postProcessCommits` functions.
- Enhanced `splitMessages` function to:
  - Separate conventional commits within the main message.
  - Extract nested commits enclosed in `BEGIN_NESTED_COMMIT`/`END_NESTED_COMMIT` blocks.
  - Preserve the original message structure outside of nested commit blocks.
  - Handle multiple nested commits and conventional commits in a single message.
- Updated tests in `test/commits.ts` to reflect the changes and added new test cases for multiple commits in a single message.
- Added a new fixture `multiple-commits-single-message.txt` for testing multiple commits in a single message.

## Additional Notes
- The changes ensure that breaking changes mentioned in the commit body are correctly appended to the breaking change notes.
- The `splitMessages` function now correctly handles and separates multiple conventional commits and nested commit blocks, improving the accuracy of commit parsing.
- Updated test cases ensure the robustness of the new functionality and maintain backward compatibility.
- First time contributing to release-please so please review carefully as the splitMessages function seems to be a critical function 🙃 

Fixes #2357

### DIFF
--- a/src/commit.ts
+++ b/src/commit.ts
@@ -169,6 +169,19 @@ function toConventionalChangelogFormat(
       }
     }
   );
+
+  // Add additional breaking change detection from commit body
+  if (body) {
+    const bodyString = String(body);
+    const breakingChangeMatch = bodyString.match(/BREAKING-CHANGE:\s*(.*)/);
+    if (breakingChangeMatch && breakingChangeMatch[1]) {
+      if (breaking.text) {
+        breaking.text += '\n';
+      }
+      breaking.text += breakingChangeMatch[1].trim();
+    }
+  }
+
   if (breaking.text !== '') headerCommit.notes.push(breaking);
 
   // Populates references array from footers:
@@ -313,6 +326,22 @@ function postProcessCommits(commit: parser.ConventionalChangelogCommit) {
     }
     note.text = text.trim();
   });
+
+  const breakingChangeMatch = commit.body?.match(/BREAKING-CHANGE:\s*(.*)/);
+  if (breakingChangeMatch && breakingChangeMatch[1]) {
+    const existingNote = commit.notes.find(
+      note => note.title === 'BREAKING CHANGE'
+    );
+    if (existingNote) {
+      existingNote.text += `\n${breakingChangeMatch[1].trim()}`;
+    } else {
+      commit.notes.push({
+        title: 'BREAKING CHANGE',
+        text: breakingChangeMatch[1].trim(),
+      });
+    }
+  }
+
   return commit;
 }
 
@@ -338,21 +367,32 @@ function parseCommits(message: string): parser.ConventionalChangelogCommit[] {
   ).map(postProcessCommits);
 }
 
-// If someone wishes to aggregate multiple, complex commit messages into a
-// single commit, they can include one or more `BEGIN_NESTED_COMMIT`/`END_NESTED_COMMIT`
-// blocks into the body of the commit
+/**
+ * Splits a commit message into multiple messages based on conventional commit format and nested commit blocks.
+ * This function is capable of:
+ * 1. Separating conventional commits (feat, fix, docs, etc.) within the main message.
+ * 2. Extracting nested commits enclosed in BEGIN_NESTED_COMMIT/END_NESTED_COMMIT blocks.
+ * 3. Preserving the original message structure outside of nested commit blocks.
+ * 4. Handling multiple nested commits and conventional commits in a single message.
+ *
+ * @param message The input commit message string
+ * @returns An array of individual commit messages
+ */
 function splitMessages(message: string): string[] {
   const parts = message.split('BEGIN_NESTED_COMMIT');
   const messages = [parts.shift()!];
   for (const part of parts) {
     const [newMessage, ...rest] = part.split('END_NESTED_COMMIT');
     messages.push(newMessage);
-    // anthing outside of the BEGIN/END annotations are added to the original
-    // commit
-    messages[0] = messages[0] + rest.join();
+    messages[0] = messages[0] + rest.join('END_NESTED_COMMIT');
   }
 
-  return messages;
+  const conventionalCommits = messages[0]
+    .split(
+      /\n(?=(?:feat|fix|docs|style|refactor|perf|test|build|ci|chore|revert)(?:\(.*?\))?: )/
+    )
+    .filter(Boolean);
+  return [...conventionalCommits, ...messages.slice(1)];
 }
 
 /**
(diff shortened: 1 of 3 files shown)


chingor13 (changes requested):
Thanks for this and sorry for the delay in reviewing.

For the splitting commit messages, this seems ok to me -- we're splitting on a newline and then the next line looks like a conventional commit message. The scenario we're intentionally trying to avoid is GitHub's default squash/merge message where incremental commit messages might be parsed which we don't want (example: https://github.com/googleapis/release-please/blob/main/test/fixtures/commit-messages/1257-breaking-change.txt). It seems this feature still ignores the subcommit messages because the next line doesn't start with the conventional commit looking message.

For the additional breaking change detection, please split it into a separate PR and add a test for it. Additionally, the current breaking change detection is handled by the conventional-commits/parser library which uses a grammar to actually conform to the conventionalcommits.org spec. I'm not sure we want to add this additional way of marking a breaking change.

dgokcin:
Hi @chingor13, thanks for the review.

I want to clarify the necessity of including the breaking change detection lines in `src/commit.ts` (Lines 172 to 184 and 329 to 344) from this branch. Without these lines, the following two tests fail, which is why I felt it was essential to add the additional checks.

1. **"DefaultChangelogNotes buildNotes with commit parsing should handle meta commits"**
   - This test fails because the generated changelog is missing the "BREAKING CHANGES" section. The expected output includes a breaking change for the `recaptchaenterprise` component, but the actual output does not contain this section, leading to a mismatch.

2. **"parseConventionalCommits parses meta commits"**
   - This test fails because it expects a commit to be recognized as a breaking change (true), but it is incorrectly identified as false. This discrepancy arises from the lack of additional breaking change detection logic, which is crucial for accurately parsing and recognizing breaking changes in commit messages.

From my understanding, the library you mentioned is not capable of detecting breaking changes in the body of the commit out of the box. If it were, I would expect the tests to pass without the additional checks. By including these lines, I ensure that the tests pass and that the changelog accurately reflects the intended breaking changes, thus maintaining the integrity of the commit parsing process and stay backward compatible.

I would be happy to try out any ideas you might have to keep the scope of this PR limited to the splitting of the commit messages, without including any breaking change detection. If you have any suggestions on how to accomplish this without affecting the breaking change logic, I would appreciate your input.

chingor13 (approved):

PR #2367, opened 2024-08-30
fix: don't escape html tags inside of inline code in release notes

### LINKED ISSUE #2366: HTML tags are incorrectly escaped within code blocks resulting in improper markdown
Ever since #1661 all `<` and `>` are replaced with `&lt;` and `&gt;`. This was to generate correct markdown, but this is incorrect markdown if the `<` and `>` are within back-ticks to be in-line code

#### Steps to reproduce

  1. Commit something in a repo that uses manifest releases with text between `<>` in a code block
    For example: "test: `<picture>` tags are now tested"
  2. See that the the generated changelog is invalid markdown
    The example generates: "test: `&lt;picture&gt;` tags are now tested"

Thanks so much for your help!

mehalter:
This makes sure to only escape HTML tags that are not within inline code blocks as these are handled explicitly and not rendered the same way they are outside of code blocks.

Fixes #2366  🦕

Example, a commit such as the following:

```
feat: render all imagesets as <picture> `<picture>` `` `<picture>` ``
```

would render as the following markdown:

render all imagesets as &lt;picture&gt; `<picture>` `` `<picture>` ``

### DIFF
--- a/__snapshots__/default-changelog-notes.js
+++ b/__snapshots__/default-changelog-notes.js
@@ -153,6 +153,15 @@ exports['DefaultChangelogNotes buildNotes with commit parsing should handle html
 * render all imagesets as &lt;picture&gt; ([383fb14](https://github.com/googleapis/java-asset/commit/383fb14708ae91f7bb7e64bf0bacab38))
 `
 
+exports['DefaultChangelogNotes buildNotes with commit parsing should handle html tags as inline code 1'] = `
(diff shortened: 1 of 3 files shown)


mehalter:
I updated the commits to follow the markdown syntax exactly for inline code including inline code blocks that contain `` ` `` characters. I also made sure all tests still pass along with adding a new test to fully validate the rest of the possible HTML escaping situations. This should be ready for review 🥳 

Thanks so much for such a great project and I'm excited to be able to help contribute!

chingor13 (approved):

PR #2440, opened 2024-11-20
feat: allow date updating in generic strategy

### LINKED ISSUE #1798: Allow for date replacement when running release-please
When releasing a version my team finds it useful to provide the date of release within our readme.md files (and extra-files as well). 

It would be great to provide the date within replacements for use within files (being updated) in addition to also doing the other version bumps.

Additional information
**Example:** Our README.md has the following

```
<!-- x-release-please-lastest-version -->
### Latest Release [1.2.3] 09-02-2022
<!-- x-replease-please-end -->
```

It would be great to bump that date to the date of the latest merge within this action instead of doing another action for this step.

jolars:
Allow updating of dates using the generic updater, by adding a x-release-please-date or x-release-please-version-date (to update both version and date) to a file anywhere.

The implementation is still very basic. The date is retrieved by just checking the current date, but a better approach would be to look at the timestamp of the previous feat/fix or breaking change for a conventional commit, I think.

I've added an option --date-format to specify the date format. There are no assertions with respect to this right now, and the regex matching can be improved in several ways.

One option I considered was to try to auto-detect the date format, but I think this is bound to be problematic because there are ambiguities in date formatting.

Fixes #1798 🦕

### DIFF
--- a/src/bin/release-please.ts
+++ b/src/bin/release-please.ts
@@ -346,6 +346,10 @@ function pullRequestStrategyOptions(yargs: yargs.Argv): yargs.Argv {
       describe: 'Override the detected latest tag name',
       type: 'string',
     })
+    .option('date-format', {
+      describe: 'format in strftime format for updating dates',
+      default: 'string',
+    })
     .middleware(_argv => {
       const argv = _argv as CreatePullRequestArgs;
 
--- a/src/strategies/base.ts
+++ b/src/strategies/base.ts
@@ -84,6 +84,7 @@ export interface BaseStrategyOptions {
   logger?: Logger;
   initialVersion?: string;
   extraLabels?: string[];
+  dateFormat?: string;
 }
 
 /**
@@ -113,6 +114,7 @@ export abstract class BaseStrategy implements Strategy {
   readonly componentNoSpace?: boolean;
   readonly extraFiles: ExtraFile[];
   readonly extraLabels: string[];
+  protected dateFormat: string;
 
   readonly changelogNotes: ChangelogNotes;
 
@@ -148,6 +150,7 @@ export abstract class BaseStrategy implements Strategy {
     this.extraFiles = options.extraFiles || [];
     this.initialVersion = options.initialVersion;
     this.extraLabels = options.extraLabels || [];
+    this.dateFormat = options.dateFormat || '%Y-%m-%d';
   }
 
   /**
@@ -330,7 +333,13 @@ export abstract class BaseStrategy implements Strategy {
       commits: conventionalCommits,
     });
     const updatesWithExtras = mergeUpdates(
-      updates.concat(...(await this.extraFileUpdates(newVersion, versionsMap)))
+      updates.concat(
+        ...(await this.extraFileUpdates(
+          newVersion,
+          versionsMap,
+          this.dateFormat
+        ))
+      )
     );
     const pullRequestBody = await this.buildPullRequestBody(
       component,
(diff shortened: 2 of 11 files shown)


src/updaters/generic.ts
```diff
@@ -52,24 +61,32 @@ export interface GenericUpdateOptions extends UpdateOptions {
...
+  private readonly dateFormat: string;
 
   constructor(options: GenericUpdateOptions) {
     super(options);
 
     this.inlineUpdateRegex = options.inlineUpdateRegex ?? INLINE_UPDATE_REGEX;
     this.blockStartRegex = options.blockStartRegex ?? BLOCK_START_REGEX;
     this.blockEndRegex = options.blockEndRegex ?? BLOCK_END_REGEX;
+    this.date = options.date ?? new Date();
+    this.dateFormat = options.dateFormat ?? '%Y-%m-%d';
```
chingor13, line 89: nit: can we define this as a constant?

src/strategies/base.ts
```diff
@@ -148,6 +150,7 @@ export abstract class BaseStrategy implements Strategy {
     this.extraFiles = options.extraFiles || [];
     this.initialVersion = options.initialVersion;
     this.extraLabels = options.extraLabels || [];
+    this.dateFormat = options.dateFormat || '%Y-%m-%d';
```
chingor13, line 153: nit: define as a constant like `DEFAULT_DATE_FORMAT` in this file

chingor13 (commented):
Thank you for this, and apologies for the delayed review.

This all seems reasonable (with a few minor nits), except that the special `version-date` replacement seems a bit arbitrary. In the example test, the java code could have separate constants for the version and the date and format it separately in code.

Is there a use case where this is generally needed?

jolars:
> This all seems reasonable (with a few minor nits), except that the special version-date replacement seems a bit arbitrary.

Hm, I'm not sure I understand what you mean by arbitrary here?

> Is there a use case where this is generally needed?

Do you mean the functionality in general? If so, then my use case is latex packages, that have this:

```latex
\ProvidesPackage{beamercolorthememoloch-highcontrast}[2025-01-17 v0.5.0 Moloch color theme]
```

But there is also https://github.com/googleapis/release-please/issues/1798, where @aaronware needed it for another purpose.

Thanks for the review! I'll fix the issues momentarily.

src/strategies/base.ts
jolars: I set this as a constant in `src/updaters/generic.ts` and imported it from there.

src/bin/release-please.ts
```diff
@@ -346,6 +346,10 @@ function pullRequestStrategyOptions(yargs: yargs.Argv): yargs.Argv {
       describe: 'Override the detected latest tag name',
       type: 'string',
     })
+    .option('date-format', {
+      describe: 'format in strftime format for updating dates',
+      default: 'string',
```
chingor13, line 351: Sorry, I missed this on the first review. This was probably meant to be `type` and/or the default value should be `%Y-%m-%d` (or left as nullable and rely on the default value set in code).

chingor13 (changes requested):
Small nit on the CLI options, but otherwise LGTM

jolars:
Ah, of course, yes, thank you for catching that. I'll set it to `type: 'string'`. Not sure what's best in terms of setting the default or not. I'll leave it as nullable for now.

chingor13 (approved):

PR #2450, opened 2024-12-10
fix: versioning-strategy field incorrect

reggi:
👋 hi (one of a couple different incoming prs) this is an easy one:

https://github.com/googleapis/release-please/blob/main/src/bin/release-please.ts#L304

The field is "versioning-strategy" not "versioning", also missing "prerelease-type".

PR #2451, opened 2024-12-11
feat: adds "semver" versioning-strategy 

reggi:
First off 🫶 thank you for woking on release-please it makes our lives much easier.

I work on the `npm` cli which uses `release-please`. I think its safe to say we're power users and rely on the `prerelease` functionality.

This PR adds a proper "semver" strategy. There's a lot of issues related to `prereleases` when it comes to the existing  `PrereleaseVersioningStrategy` and `DefaultVersioningStrategy` (more below), these can be fixed but I think still will not do "proper" / "intended" semver incrementation.

There are also issues with semver itself which we aim to fix in https://www.npmjs.com/package/semver in a future release, this pr shims that functionality for now with a custom `semver.inc`  impementation.

I'll copy some of my findings from https://github.com/npm/template-oss/pull/495 but keep in mind the context:

* I was unable to get a repo *into* prerelease using the latest version of release-please on my personal project `reggi/packages`.
* And the `npm/cli` has been stuck in prerelease mode. It was able to get itself into prerelease because of the wrapper around release-please we have in this repo. (trail: [npm/cli workflow](https://github.com/npm/cli/blob/latest/.github/workflows/release.yml#L54), [Strategy Turnery](https://github.com/npm/template-oss/blob/abdc9398f47d4761d589617235bee0fdb4827128/lib/release/release-please.js#L56))

I do not believe release please has this `options.prerelease ternary anymore` here's the [mapping of what versioning-strategy is uses](https://github.com/googleapis/release-please/blob/3e807972fc8d1a348cd45e32d933ada4d25d4a28/src/factories/versioning-strategy-factory.ts#L40-L47).

And even if it did choose to use `PrereleaseVersioningStrategy` or `DefaultVersioningStrategy` neither of these strategies have the ability to leave a prerelease. Because they always [add the current `prereleaseType` or `prereleaseType` from the config from the existing version](https://github.com/googleapis/release-please/blob/3e807972fc8d1a348cd45e32d933ada4d25d4a28/src/versioning-strategies/prerelease.ts#L116-L135). Neither of these classes even have access to the `options.prerelease` boolean. And at the point in which both have it whats the point in differentiating between them anymore?

This is added as a new versioning-strategy `semver`. To note the property currently documented as `versioning-strategy` in the docs is not working, and is currently `versioning`. [I opened a PR to fix the json schema for the config](https://github.com/googleapis/release-please/pull/2450), but now we know it's a bug i'll leave that discussion open there.

So `"versioning-strategy": "semver"` would be ignored and you'd need `"versioning": "semver"`, again not relevant to *this* PR but another quirk to note.

I'm going to keep this in draft for now, just want to open up for discussion, and next steps. My main questions are 

* Is the `release-please` team open to new "semver" versioning strategy?
* How could we fix the existing strategies? I think they're design overall is broken (unless I'm missing something) but perhaps outside the context of semver / node people rely on that functionality? Is there a world in which we can deprecate `PrereleaseVersioningStrategy` or `DefaultVersioningStrategy` and just use `SemverVersioningStrategy`?

I also have a published fork if anyone wants to try it https://www.npmjs.com/package/@reggi/release-please

### DIFF
--- a/__snapshots__/cli.js
+++ b/__snapshots__/cli.js
@@ -201,7 +201,7 @@ Options:
                                     generated?        [boolean] [default: false]
   --versioning-strategy             strategy used for bumping versions
         [choices: "always-bump-major", "always-bump-minor", "always-bump-patch",
-                   "default", "prerelease", "service-pack"] [default: "default"]
+         "default", "prerelease", "semver", "service-pack"] [default: "default"]
   --changelog-path                  where can the CHANGELOG be found in the
                                     project?  [string] [default: "CHANGELOG.md"]
   --changelog-type                  type of changelog to build
--- a/src/factories/versioning-strategy-factory.ts
+++ b/src/factories/versioning-strategy-factory.ts
@@ -21,6 +21,7 @@ import {ServicePackVersioningStrategy} from '../versioning-strategies/service-pa
 import {GitHub} from '../github';
 import {ConfigurationError} from '../errors';
 import {PrereleaseVersioningStrategy} from '../versioning-strategies/prerelease';
+import {SemverVersioningStrategy} from '../versioning-strategies/semver';
 
 export type VersioningStrategyType = string;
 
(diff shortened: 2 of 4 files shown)


reggi:
Ping, would love to start a convo about this one!

PR #2505, opened 2025-03-12
fix: workspace - always update manifest versions

### LINKED ISSUE #2172: Not updating `.release-please-manifest.json`
We use `release-please` to create releases for the Puppeteer project. The current structure is a mono repo so we use a [configuration file ](https://github.com/puppeteer/puppeteer/blob/main/release-please-config.json) to express this.
This works well for the 3 packages that depend on each other  `puppeteer`, `puppeteer-core`, and `@puppeteer/browsers`.
But for the last one `@puppeteer/ng-schematics` we are seeing issue that the release does not update [`.release-please-manifest.json`](https://github.com/puppeteer/puppeteer/blob/main/.release-please-manifest.json).

This can be observed on the latest PR - https://github.com/puppeteer/puppeteer/pull/11543.
If we do merge this PR without updating the manifest manually `release-please` will open a new one as the version differs with the manifest file. Observed:
Initial - https://github.com/puppeteer/puppeteer/pull/11496 
A second one - https://github.com/puppeteer/puppeteer/pull/11508

The issue is not consistent. It doesn't fail all the time as you can see from https://github.com/puppeteer/puppeteer/pull/11488

#### Environment details

  - GitHub App

#### Steps to reproduce

  1.  Have not reproduced locally, but happy to follow up if instruction are provided.

raphael-trzpit:
- [x ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/release-please/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ x] Ensure the tests and linter pass
- [ x] Code coverage does not decrease (if any source code was changed)
- [ x] Appropriate docs were updated (if necessary)

Fixes #2172 🦕

For the NodeWorkspace and CargoWorkspace plugin, manifest updates are generated with an empty versionMap.

From what I understand, the [buildUpdatedVersions](https://github.com/googleapis/release-please/blob/main/src/plugins/workspace.ts#L185C22-L185C41) method is only adding candidates in the `updatedPathVersions` return (but is correctly adding everything in the `updatedVersions` return).
This feels like an oversight, as I don't know why the `updatedVersions` and the `updatedPathVersions` could have different versions.

I managed to make tests fail with my first commit, where manifest updates are not generated for cargo and node in the current version. (I also added the assertions for the maven plugin, but it works correctly).

The second commit introduce my proposition to fix this: always adding new versions to both outputs of the buildUpdatedVersions method.

What do you think about this change ?
Could this modification have side effects that I am not aware of ?

### DIFF
--- a/src/plugins/workspace.ts
+++ b/src/plugins/workspace.ts
@@ -262,6 +262,9 @@ export abstract class WorkspacePlugin<T> extends ManifestPlugin {
         const version = existingCandidate.pullRequest.version!;
         this.logger.debug(`version: ${version} from release-please`);
         updatedVersions.set(packageName, version);
+        if (this.isReleaseVersion(version)) {
+          updatedPathVersions.set(this.pathFromPackage(pkg), version);
+        }
       } else {
         const version = this.bumpVersion(pkg);
         this.logger.debug(`version: ${version} forced bump`);
--- a/test/plugins/cargo-workspace.ts
+++ b/test/plugins/cargo-workspace.ts
@@ -36,6 +36,7 @@ import {CargoToml} from '../../src/updaters/rust/cargo-toml';
 import {parseCargoManifest} from '../../src/updaters/rust/common';
 import {ConfigurationError} from '../../src/errors';
 import assert = require('assert');
+import {ReleasePleaseManifest} from '../../src/updaters/release-please-manifest';
 
 const sandbox = sinon.createSandbox();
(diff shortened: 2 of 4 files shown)


raphael-trzpit:
Hi @chingor13 , I hope you're doing well! I just wanted to ping you for a review when you have a moment. Let me know if you need any changes or more context.
Thanks in advance!

PR #2555, opened 2025-06-08
feat: add support for Julia projects

### LINKED ISSUE #2152: Add release type for Julia packages
It would be nice to have a release type for julia packages.

As a minimum it should update `version` in the `Project.toml` file. Here an example: https://github.com/JuliaMusic/MIDI.jl/blob/main/Project.toml

Most packages in Julia do not keep a changelog, but I think it would make sense to at least have an optional `CHANGELOG.md` (like the package above does, https://github.com/JuliaMusic/MIDI.jl/blob/main/CHANGELOG.md)

cadojo:
Julia projects have their versions specified in their dependency files, not their lock files. Their dependency files can be named `JuliaProject.toml` or (more commonly) `Project.toml`. Each package version is held by the `'version'` key in the project TOML file. This makes `release-please` parsing simple!

---

Fixes #2152


chingor13 (changes requested):
Thanks for this! It seems reasonable, but the tests are failing.

jdc-pub:
> Thanks for this! It seems reasonable, but the tests are failing.

@chingor13 I've updated the listing rules! Would it be possible for you to enable CI for this PR?

jdc-pub:
This is ready for review! Tests should be passing, and I've opened a draft [PR](https://github.com/JuliaAstro/GeneralAstrodynamics.jl/pull/275) in a Julia monorepo I maintain as an example.

jdc-pub:
@chingor13 could you please enable CI once again? I've done some more testing locally, and I'm confident tests will pass (for non-Windows tests at least, where I'm able to test locally).

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "./build/src/index.js",
   "bin": "./build/src/bin/release-please.js",
   "scripts": {
-    "test": "cross-env ENVIRONMENT=test LC_ALL=en c8 mocha --node-option no-experimental-fetch --recursive --timeout=5000 build/test",
+    "test": "cross-env ENVIRONMENT=test LC_ALL=en c8 mocha --recursive --timeout=5000 build/test",
```
cadojo: I'm unsure about this change, but I needed it to run tests locally.

jdc-pub:
@chingor13 could you please enable CI? I think tests should pass now.

jdc-pub:
@ferrarimarco @chingor13 gentle ping, I think this PR is good to go!

PR #2567, opened 2025-07-21
fix: Correctly handle proxies with undici

### LINKED ISSUE #2259: Proxy fails when using HTTP CONNECT (E.G. HTTPS via a HTTP only proxy)
release-please fails when trying to connect to a HTTP only proxy via HTTP CONNECT
See:
https://github.com/googleapis/release-please/blob/08331fa555ed482f675fb3e9dd5b0a5296e200a3/src/github.ts#L226-L231

It would probably make sense just to take the whole proxy URL (E.G. http://my-proxy:80) and apply it to the HTTP agent instead of making assumptions about the users environment?

Happy to raise a PR for this 🙂 

Note: This would also require an update in the GitHub action to stop it splitting the proxy input and would probably therefore be a breaking change for both release-please and the action.

#### Environment details

  - OS: Ubuntu 22.04 x64 and Windows 10 x64
  - Node.js version: 20.12.1
  - npm version: 10.5.1
  - `release-please` version: 16.10.1

#### Steps to reproduce

  1. Use a proxy value such as `my-proxy:80` which only accepts `http` connections
  2. Get an error when release-please incorrectly tries to use `https` connection to the proxy.

PeterC89:

Fixes #2259 🦕

This PR does a number of things:

- Removes redundant `request` options
- Removes old `http-proxy-agent` based proxying as this isn't supported by native node fetch
- Adds undici's `EnvHttpProxyAgent` to support proxies using standard environment vars
- ~~Bumps minimum node version to 20 as 18 is EOL~~

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [18, 20]
+        node: ['>=20.18.1 <21', 22]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v3
@@ -33,7 +33,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: ">=20.18.1 <21"
       - run: npm install
       - run: npm test
         env:
@@ -44,7 +44,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: ">=20.18.1 <21"
       - run: npm install
       - run: npm run lint
   docs:
@@ -53,7 +53,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: ">=20.18.1 <21"
       - run: npm install
       - run: npm run docs
       - uses: JustinBeckwith/linkinator-action@v1
@@ -68,6 +68,6 @@ jobs:
           ref: main
       - uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: ">=20.18.1 <21"
       - run: npm install --save googleapis/release-please#${{ github.ref }}
       - run: npm run build
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,6 +35,7 @@
         "semver": "^7.5.3",
         "type-fest": "^3.0.0",
         "typescript": "^4.6.4",
+        "undici": "^7.18.2",
         "unist-util-visit": "^2.0.3",
         "unist-util-visit-parents": "^3.1.1",
         "xpath": "^0.0.34",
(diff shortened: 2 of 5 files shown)


meiswjn:
Nice PR!
@chingor13 any ETA on this already? Thanks!

kauppfbi:
Hey @chingor13, is there any update on this PR? Any plans to introduce and release it soon, or do you plan a different implementation for #2259?

schulmak:
Any updates on this PR? We would really appreciate a merge if nothing blocks it!

hmrc87:
@chingor13  heavily needed here

ferrarimarco:
Hi @PeterC89. Can you please rebase against the latest `main`, and elaborate on the changes you're proposing? thanks.

meiswjn:
@ferrarimarco Thanks for the response. I think Peter explained it already, especially in https://github.com/googleapis/release-please/issues/2259 - what part is unclear? Perhaps I can also jump in.
@PeterC89 can you rebase this? Alternatively, @ferrarimarco you could also rebase it as the maintainer?

ferrarimarco:
There are a bunch of changes in this PR, but there's no explanation about why they are needed. 😄

PeterC89:
@ferrarimarco 
The current implementation doesn't work due to:

1. A move to node native `fetch` / `undici` in the underlying `ocktokit` [libraries](https://github.com/octokit/request.js/blob/9ba6ae0fd2eff35a6c29a9f032eca6e9ba002976/package.json#L45), which doesn't use the same agent method and instead uses dispatchers for handling proxies.

2. The original implementation is incorrect anyway for proxy connections using `HTTP CONNECT`, as it wrongly assumes that to connect to a `https` endpoint, you have to use a `https` proxy, which is not the case.

Additionally Node18 was removed as it was EOL, at this time GitHub are also [deprecating](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/) Node20 on their runners as it goes EOL in April.

Given that is has been nearly 2 years since the original issue was raised we have moved on as a team from trying to use `release-please` as it seemed like there was no support forthcoming to solve this issue.
I will look to rebase this PR or I can close if preferred.

ferrarimarco:
Thanks for updating the PR description.

I suggest limiting changes in this PR to what's absolutely necessary to fix the issue, so review is easier.

Also, I'd suggest to add tests where applicable. I see that this PR removes tests, and doesn't add any, so might be useful to add some. Note that we don't look at testing the proxy feature end-to-end, but rather baseline testing to check that we have everything in place as needed.

PeterC89:
@ferrarimarco The existing tests are wrong anyway, hence why they were removed.
It's difficult to test that a proxy is working without end to end testing but I'll take a look.

PeterC89:
@ferrarimarco I remember the other reason for bumping to node 18 now, Uncidi needs `File` from `Buffer` which isn't available until Node [18.13](https://nodejs.org/docs/latest-v18.x/api/buffer.html#class-file)
Additionally the latest version of `undici`, which I assume would be desired to ensure there are minimal security vulnerabilities, requires Node 20.
Therefore I will continue with requiring Node 20 as part of this PR.

meiswjn:
@ferrarimarco Would it be possible to review this before another rebase is necessary? :) I guess Peter provided the changes requested. Thanks!

meiswjn:
@ferrarimarco @chingor13  Any updates? :) Considering that Peter was specifically asked to rebase etc. it would be nice to have this reviewed soon

webben-de:
yeah i agree with that

stevehaskew:
@ferrarimarco @chingor13 +1 for needing/depending on this!

hmrc87:
@ferrarimarco Please approve this

PR #2627, opened 2025-11-23
feat: add --force-tag-creation option to explicitly create git tags for releases.

### LINKED ISSUE #1650: Using `draft: true` causes `release-pr` to fail to find the previous release.

Product issue

Not that I'm aware.

No

#### Environment details

  - OS: MacOS Monterey
  - Node.js version: 14
  - npm version:
  - `release-please` version:

#### Steps to reproduce

1. use `draft` GH release
2. make a release using `github-release` command
3. immediately issue a `release-pr` command

The `release-pr` will fail to find the drafted release and will immediately create a new "release PR".

From my investigation, this is due to the fact that "draft" GH releases are able to lazily create tags.
Which is to say that when you create a draft GH release, you can specify the tag name and `ref`, but those are not yet materialized in `git`; instead, they are only created when the release is published.
That said, "draft" GH releases can also point at existing tags and will "look" like any other GH release except with `isDraft` set to `true`.

For example, in my fork for this repo, I created 2 releases. One pointing at an existing tag (`v14.6.0-exists`) and another that will lazily create the tag (`v14.6.1-lazy`). 

Now, if we issue the following GraphQL query (which is used by `release-please` [here](https://github.com/googleapis/release-please/blob/7d4abfbc097c6841c1c5194e531417a7e80daa8e/src/github.ts#L825-L846), except for the additional `tagName` field):

```graphql
query ($name: String!, $owner: String!) {
  repository(owner: $owner, name: $name) {
    releases(orderBy: {field: CREATED_AT, direction: DESC}, first: 10) {
      nodes {
        name
        isDraft
        url
        tag {
          name
        }
        tagCommit {
          oid
        }
        tagName
      }
    }
  }
}
```

We obtain this:

```json
{
  "data": {
    "repository": {
      "releases": {
        "nodes": [
          {
            "name": "v14.6.0-exists",
            "isDraft": true,
            "url": "https://github.com/plaflamme/release-please/releases/tag/untagged-1343daf7aae1ed4c3da5",
            "tag": {
              "name": "v14.6.0"
            },
            "tagCommit": {
              "oid": "faaf56cc557cbcfdbb881efe81e1da8e22441d31"
            },
            "tagName": "v14.6.0"
          },
          {
            "name": "v14.6.1-lazy",
            "isDraft": true,
            "url": "https://github.com/plaflamme/release-please/releases/tag/untagged-dd0aaa72f95f1a467bbe",
            "tag": null,
            "tagCommit": null,
            "tagName": "v14.6.1"
          }
        ]
      }
    }
  }
}
```

As you can see the `tag` and `tagCommit` values are missing for the "lazy" draft release. NOTE: you will not be able to reproduce this on my fork since "draft" releases are private.

This causes the iterator defined [here](https://github.com/googleapis/release-please/blob/7d4abfbc097c6841c1c5194e531417a7e80daa8e/src/github.ts#L860) to skip this release and makes `release-pr` fail to find the "previous" release.

Now the question is what should be the fix? Initially, I thought about removing the filter and properly handle the missing `tagCommit` and `tag` fields; but an alternate approach would be to make the `github-release` command actually create the `tag` in `git` and avoid these "lazy" releases altogether. This second approach seems more correct since the `github-release` command did indeed create the release and the repository should be tagged accordingly.

Any thoughts on what the fix should be here?

Fanzzzd:

Fixes #1650

### DIFF
--- a/__snapshots__/cli.js
+++ b/__snapshots__/cli.js
@@ -58,8 +58,8 @@ Options:
                                 with the release for future tag creation upon
                                 "un-drafting" the release.
                                                       [boolean] [default: false]
-  --force-tag-creation          Force the creation of a Git tag for the
-                                release.              [boolean] [default: false]
+  --force-tag-creation          Force the creation of a Git tag for the release.
+                                                      [boolean] [default: false]
   --prerelease                  mark release that have prerelease versions as as
                                 a prerelease on Github[boolean] [default: false]
   --label                       comma-separated list of labels to remove to from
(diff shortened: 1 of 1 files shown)


Fanzzzd:
Use this PR to add "force-tag" to "release-please-config.json" to force the creation of tags. With this option, the draft releases can be traced by the force-created tags.

This option is set to false by default, so it won't affect other implementations.

Here is an example of how to use this option:
```release-please-config.json
 {
    "packages": {
        "apps/desktop": {
            "release-type": "node",
            "package-name": "@acme/desktop",
            "draft": true,
            "force-tag": true, <---Set "force-tag" to true
            "extra-files": [
                "src-tauri/Cargo.toml",
                "src-tauri/Cargo.lock"
            ]
        },
        "apps/web": {
            "release-type": "node",
            "package-name": "@acme/web",
            "draft": true,
            "force-tag": true, <---Set "force-tag" to true
        }
    },
    "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json"
}
```

Fanzzzd:
And I also forked the release-please action to test this branch more easily.

You can replace:

```
-  uses: googleapis/release-please-action@v4
```

in your `release-please.yml` with:

```
-  uses: Fanzzzd/release-please-action@release-with-force-tag
```

to try this new feature.

Here is the link for your interest: [[Fanzzzd/release-please-action](https://github.com/Fanzzzd/release-please-action)]

__snapshots__/cli.js
```diff
@@ -58,6 +58,9 @@ Options:
                                 with the release for future tag creation upon
                                 "un-drafting" the release.
                                                       [boolean] [default: false]
+  --force-tag                   explicitly create a git tag for the release.
```
ferrarimarco: Refactor to:

> Force the creation of a Git tag for the release

Check the rest of the PR to update similar strings, thanks

ferrarimarco (changes requested):
Thanks for this PR. A few things to check, then I'll have a look at testing this.

Fanzzzd:
@ferrarimarco Thank you for the thorough review! I've addressed all your feedback:

1. Renamed --force-tag to --force-tag-creation to better convey what the option does.
2. Updated the description to "Force the creation of a Git tag for the release."
3. Extended the documentation to explain why this is useful when draft is enabled - GitHub does not create tags for draft releases until they are published, which causes release-please to fail to find the previous release.

Please let me know if there's anything else you'd like me to adjust.

src/github.ts
```diff
@@ -1391,6 +1392,21 @@ export class GitHub {
...
+        try {
+          await this.octokit.git.createRef({
+            owner: this.repository.owner,
+            repo: this.repository.repo,
+            ref: `refs/tags/${release.tag.toString()}`,
+            sha: release.sha,
+          });
+        } catch (err: any) {
+          // ignore if tag already exists
+          if (err.status !== 422) {
```
ferrarimarco: Can you add a debug log line when this happens so we have some feedback?

erka:
Hey @Fanzzzd and @ferrarimarco. Any progress with this? How could we help to push it forward?

Fanzzzd:
@ferrarimarco , thank you for the review. I have addressed the two issues you mentioned.

ferrarimarco (approved):

erka:
Thank you all so much!

bendasgfyug:
> (compileroptioned":tsbuildinfofile":node modules

PR #2633, opened 2025-12-08
feat: add include-v-in-release-name config option

mldangelo:
## Summary

This PR adds a new configuration option `include-v-in-release-name` that controls whether the `v` prefix is included in GitHub release names.

### Problem

Currently, the GitHub release name is hardcoded to always include the `v` prefix (e.g., `v1.2.3`). Users who prefer releases without the `v` prefix have no way to configure this. While `include-v-in-tag` exists to control the tag name, there's no equivalent for the release name.

### Solution

Added a new `include-v-in-release-name` boolean configuration option that:

- Defaults to `true` (maintains backwards compatibility)
- When set to `false`, creates releases without the `v` prefix (e.g., `1.2.3`)
- Works independently of `include-v-in-tag` to allow different formats for tags vs release names

### Changes

- **schemas/config.json**: Added `include-v-in-release-name` property to the schema
- **src/manifest.ts**: Added `includeVInReleaseName` to `ReleaserConfig` interface and config parsing
- **src/strategies/base.ts**: Modified release name generation to use the new option
- **src/updaters/release-please-config.ts**: Added serialization support
- **test/strategies/base.ts**: Added tests for the new functionality

### Example Config

```json
{
  "include-v-in-tag": false,
  "include-v-in-release-name": false,
  "packages": {
    ".": {
      "release-type": "node"
    }
  }
}
```

With this config:
- Tag: `1.2.3`
- Release name: `1.2.3`

### Use Case

Users who:
1. Want consistent naming between tags and release names without `v`
2. Have downstream tooling that expects release names without the `v` prefix
3. Prefer semantic version numbers without the `v` prefix for releases

This follows the pattern established by the existing `include-v-in-tag` option and provides users with complete control over version prefix formatting.

### DIFF
--- a/schemas/config.json
+++ b/schemas/config.json
@@ -86,6 +86,10 @@
           "description": "When tagging a release, include `v` in the tag. Defaults to `false`.",
           "type": "boolean"
         },
+        "include-v-in-release-name": {
+          "description": "Include `v` in the GitHub release name. Defaults to `true`.",
+          "type": "boolean"
+        },
         "changelog-type": {
           "description": "The type of changelog to use. Defaults to `default`.",
           "type": "string",
@@ -471,6 +475,7 @@
     "extra-label": true,
     "include-component-in-tag": true,
     "include-v-in-tag": true,
+    "include-v-in-release-name": true,
     "changelog-type": true,
     "changelog-host": true,
     "changelog-path": true,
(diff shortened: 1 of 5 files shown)


unique-dominik:
I was looking for that! Could this be reviewed? ❤️

PR #2646, opened 2026-01-12
docs: fix incorrect default value for include-v-in-tag option

### LINKED ISSUE #2645: Documentation: Incorrect default value for include-v-in-tag option
## Description

The schema documentation states that the include-v-in-tag option defaults to false, but the actual implementation defaults to true.

## Location of the Discrepancy

Schema documentation ([schemas/config.json:85-87)](https://github.com/googleapis/release-please/blob/main/schemas/config.json#L85-L87):
```json
"include-v-in-tag": {
    "description": "When tagging a release, include `v` in the tag. Defaults to `false`.",
    "type": "boolean"
},
```

Actual implementation ([src/strategies/base.ts:151](https://github.com/googleapis/release-please/blob/main/src/strategies/base.ts#L151)):
```javascript
this.includeVInTag = options.includeVInTag ?? true;
```

## Expected Behavior

The documentation and the implementation should be consistent. Either:

1. Update the schema documentation to state Defaults to `true`
2. Or update the implementation to default to false

Given that changing the implementation default could be a breaking change for existing users, updating the documentation to reflect the actual behavior (true) seems like the safer approach.

## Additional Context

This inconsistency can lead to confusion for users who configure their releases based on the documented default value, only to find that the actual behavior differs.

gimsesu:
The schema documentation incorrectly stated that include-v-in-tag defaults to false, but the actual implementation in src/strategies/base.ts defaults to true.

Fixes #2645 🦕

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,7 +55,6 @@
         "@types/mocha": "^10.0.0",
         "@types/node": "^18.0.0",
         "@types/npmlog": "^7.0.0",
-        "@types/pino": "^7.0.0",
         "@types/semver": "^7.0.0",
         "@types/sinon": "^17.0.0",
         "@types/xmldom": "^0.1.31",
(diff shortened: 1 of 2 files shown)


ferrarimarco:
Thanks for this PR!

PR #2679, opened 2026-02-18
feat: add ability to configure how many commits we retrieve when iterating through repo history

ldetmer:
Release please has been consistently failing on google-cloud-node repo due to 502 errors when trying to [run this query](https://github.com/googleapis/release-please/blob/main/src/github.ts#L402).  The only way to actually have it succeed was to update the number of commits per request to be 1.  We don't want to make this change for all clients as it will slow down  release please run time, so lets make it configurable so node can set it to a low value.

### DIFF
--- a/schemas/config.json
+++ b/schemas/config.json
@@ -431,6 +431,10 @@
           "description": "When considering commit history, only look this many commits deep.",
           "type": "number"
         },
+        "commit-batch-size": {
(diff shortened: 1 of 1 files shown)


chingor13:
> Can you add this to the manifest config json schema in `schemas/`?

Sorry, `config.json`, not manifest.

chingor13 (approved):

colebanderson8392-arch (approved):

PR #2693, opened 2026-03-02
feat: support uv.lock files in the Python strategy

### LINKED ISSUE #2561: Support `uv.lock` files out-of-the-box.
`release-please` already understands and correctly manages the `pyproject.toml` file used by [`uv`](https://docs.astral.sh/uv/), but not the`uv.lock` lockfile, which also contains the package version. The trick is to add a special `extra-files` entry to `release-please-config.json`:

```json
      "extra-files": [
        {
          "path": "uv.lock",
          "type": "toml",
          "jsonpath": "$.package[?(@.name.value=='my-package')].version"
        },
```
as described in [this extremely helpful issue comment](https://github.com/googleapis/release-please/issues/2455#issuecomment-2643132770) from a user.

It would be great to not need this trick!

jlopez:
## Summary

Closes #2561

The Python strategy now automatically updates the project's entry in `uv.lock` when cutting a release, without requiring any `extra-files` configuration.

### What changed

- New `UvLock` updater (`src/updaters/python/uv-lock.ts`) — mirrors the existing `CargoLock` pattern: parses the lockfile, finds the matching `[[package]]` entry by name, and uses `replaceTomlValue()` to update the version while preserving all formatting and comments.
- `Python` strategy (`src/strategies/python.ts`) — adds `uv.lock` to `buildUpdates()` alongside the other Python files. Gated on `projectName` being known (same condition as `__init__.py` updates). `createIfMissing: false` makes it a safe no-op for projects that don't use uv.
- Package name matching applies PEP 503 normalization (`My_Package` → `my-package`) so names are compared case-insensitively with `-`, `_`, and `.` treated as equivalent.
- Virtual workspace members (entries without a `version` field) are skipped with a specific warning distinguishing them from packages that are genuinely absent from the lockfile.

## Test plan

- [ ] `test/updaters/uv-lock.ts` — unit tests covering: version update with formatting preserved (snapshot), PEP 503 name normalization (case, underscores, dots), virtual package warning, not touching unrelated dependencies, empty lockfile, missing package
- [ ] `test/strategies/python.ts` — asserts `uv.lock` appears in the update list for both the common-files and skip-changelog paths, in addition to when `pyproject.toml` is present
- [ ] `npm test` — all passing, 0 failing

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/updaters/uv-lock.ts
+++ b/test/updaters/uv-lock.ts
@@ -1,4 +1,4 @@
-// Copyright 2025 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/test/updaters/uv-lock.ts
+++ b/test/updaters/uv-lock.ts
@@ -1,0 +1,90 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {readFileSync} from 'fs';
+import {resolve} from 'path';
+import * as snapshot from 'snap-shot-it';
+import {describe, it} from 'mocha';
+import {expect} from 'chai';
+import * as TOML from '@iarna/toml';
+import {UvLock} from '../../src/updaters/python/uv-lock';
+import {Version} from '../../src/version';
+
+const fixturesPath = './test/updaters/fixtures';
+
+function loadFixture(): string {
+  return readFileSync(resolve(fixturesPath, 'uv.lock'), 'utf8').replace(
+    /\r\n/g,
+    '\n'
+  );
+}
+
+function parsedPackages(content: string): {name?: string; version?: string}[] {
+  const parsed = TOML.parse(content) as {
+    package?: {name?: string; version?: string}[];
+  };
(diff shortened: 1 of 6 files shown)


westonplatter:
I'm interested in using this. Anything I can do to help + support?

jlopez:
> I'm interested in using this. Anything I can do to help + support?

It seems to be stuck on Google's reviewing side. Not sure if there's anything that can be done to move that along.

StayPirate:
I'm also interested in using this. Is there any update?

kojiromike (commented):
Thanks for this — we hit exactly the stale-`uv.lock`-at-release problem this fixes, and I've tested the branch against real lockfiles (including crafted edge cases run directly through the compiled updater). The core mechanics check out: the `replaceTomlValue` splice only touches the matched entry's `version` value (version strings embedded in dependency URLs/hashes are untouched), PEP 503 normalization handles uppercase/underscore/dot names on both sides, and missing lockfile / no-`[[package]]` / virtual-root cases all degrade gracefully. The warn-instead-of-throw on an empty lockfile is a nice deliberate improvement over the `CargoLock` precedent.

A few observations from that testing, inline — one documented-limitation-worthy (uv workspaces), the rest small.

----

Uh, sorry about the above AI-generated comment. Please disregard. I had not intended to post it. It was a pending review comment I used for guidance and intended to delete. I learned a new implicit behavior of GitHub I'll need to watch out for.

test/updaters/uv-lock.ts
```diff
@@ -0,0 +1,152 @@
...
+        packageName: 'socketry',
+        version: Version.parse('0.3.0'),
+      });
+      const newContent = uvLock.updateContent(oldContent);
+      const pkg = parsedPackages(newContent).find(p => p.name === 'socketry');
+      expect(pkg).to.deep.include({name: 'socketry', version: '0.3.0'});
+      snapshot(newContent);
+    });
+
+    it('normalizes package name per PEP 508 when matching', () => {
```
kojiromike, line 55: Did you mean PEP 503?

```suggestion
    it('normalizes package name per PEP 503 when matching', () => {
```

PR #2709, opened 2026-03-18
feat: resolve Dependabot security alerts

codyoss:
Addresses multiple high and moderate vulnerabilities found via `npm audit` in devDependencies:
- Moderate: ajv (ReDoS) - Resolved via automatic fixes
- High: flatted (DoS) - Resolved via automatic fixes
- High: serialize-javascript (RCE) - Resolved via manual override

Changes:
1. Run `npm audit fix` for automatic resolution of patchable transitive dependencies.
2. Update `snap-shot-it` to `^7.9.10` in `package.json`.
3. Add `overrides` in `package.json` to force a safe version of `serialize-javascript` to `^7.0.4`.
4. Drop node 18 support to enable updating some deps.

Verified with `npm audit` (0 critical/high vulnerabilities).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -73,7 +73,7 @@
         "snap-shot-it": "^7.9.10"
       },
       "engines": {
-        "node": ">=18.0.0"
+        "node": ">=20.0.0"
       }
     },
     "node_modules/@babel/code-frame": {


codyoss:
Bah, node 18 strikes again. I am going to remove it in this PR and make this a feat. I am not sure that is the node way but we need to update these deps.

codyoss:
Interesting, so I need to update synthtool as it is overwriting my package json change.

codyoss:
synthtool running here actually did an okay thing, but I have unregistered owlbot from touching this repo in the future as it was not intended to run.

codyoss:
If I get approval I will remove the required 18 check.

codyoss:
@chingor13 bump

PR #2713, opened 2026-03-22
feat: julia release type with registration included

### LINKED ISSUE #2152: Add release type for Julia packages

MilesCranmer:
This adds a `julia` release strategy to release-please to support both `Project.toml` and `JuliaProject.toml`.

As with other release please workflows, this:
- update Julia project versions while preserving formatting/comments
- infer the package/component name from the Julia project file

This also tries to match other workflows in that after creating a release, it will post a Registrator commit comment with release notes so the Julia registry registration can be triggered automatically. (My fave part of using the rust workflow)

It also supports subdir packages by using `subdir=...` in the Registrator trigger

This builds directly on the earlier Julia work in #2555, and extends it with the automatic Julia registry registration step after release creation.

I also tested the full workflow end-to-end against a fake Julia package and a fake clone of General:

- fake package repo: https://github.com/MilesCranmerBot/FakeJuliaReleasePkg
- release PR created by release-please: https://github.com/MilesCranmerBot/FakeJuliaReleasePkg/pull/1
- post-merge release action run: https://github.com/MilesCranmerBot/FakeJuliaReleasePkg/actions/runs/23392761804
- fake General clone: https://github.com/MilesCranmerBot/GeneralTestRegistry
- registry PR opened there by the workflow: https://github.com/MilesCranmerBot/GeneralTestRegistry/pull/1

---

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/release-please/issues/new/choose) before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea

Fixes #2152 🦕

---

cc @chingor13 @ferrarimarco - would you be willing to review? This builds on #2555 and adds the Julia Registrator auto-registration step.

@joeycarpinelli this builds on your nice earlier PR.


MilesCranmer:
Confirming it works in production too, check it out:

https://github.com/MilesCranmer/BorrowChecker.jl/commit/d5752615ed3d722d9451b592729416d695aef730#commitcomment-180265743

Which creates

https://github.com/JuliaRegistries/General/pull/151076

src/manifest.ts
```diff
@@ -1317,6 +1319,30 @@ export class Manifest {
...
+      const config = this.repositoryConfig[release.path];
+      if (config.releaseType !== 'julia') continue;
+
+      await this.github.commentOnCommit(
+        this.buildJuliaRegistratorComment(release),
+        release.sha
+      );
+    }
+  }
+
```
chingor13, line 1345: Can we separate out this extra logic from the Julia strategy and file updates?

Everything in manifest.ts should be strategy/language-agnostic so it doesn't feel right having special logic for Julia here

chingor13 (changes requested):
Let's separate the commenting logic out of this PR and separate it as it's an architectural thing that we have the correct abstraction for

src/manifest.ts
MilesCranmer: How do you recommend we carry out this part? Julia's registration is GitHub bot-based so might deserve the special casing. Not sure though!

PR #2775, opened 2026-05-12
feat: update librarian.yaml for Node

jskeet:
This introduces a new Node strategy specifically for https://github.com/googleapis/google-cloud-node, with an updater which is aware of the expected Librarian format for Node.

Once Librarian has moved off release-please, the strategy and updater can be removed.

Towards https://github.com/googleapis/librarian/issues/5461

### DIFF
--- a/__snapshots__/cli.js
+++ b/__snapshots__/cli.js
@@ -51,9 +51,9 @@ Options:
                                 for?
   [choices: "bazel", "dart", "dotnet-yoshi", "elixir", "expo", "go", "go-yoshi",
           "helm", "java", "java-backport", "java-bom", "java-lts", "java-yoshi",
-       "java-yoshi-mono-repo", "krm-blueprint", "maven", "node", "ocaml", "php",
-         "php-yoshi", "python", "r", "ruby", "ruby-yoshi", "rust", "salesforce",
-                                           "sfdx", "simple", "terraform-module"]
+     "java-yoshi-mono-repo", "krm-blueprint", "maven", "node", "node-librarian",
+       "ocaml", "php", "php-yoshi", "python", "r", "ruby", "ruby-yoshi", "rust",
+                             "salesforce", "sfdx", "simple", "terraform-module"]
   --config-file                 where can the config file be found in the
                                 project? [default: "release-please-config.json"]
   --manifest-file               where can the manifest file be found in the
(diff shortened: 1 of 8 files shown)


jskeet:
(This was largely copied from #2750...)

chingor13 (changes requested):
Does this need to be a new strategy? If the librarian.yaml file doesn't exist on other yoshi-node repos, then it's a no-op.

Can we add it to the existing strategy?

chingor13:
> Does this need to be a new strategy? If the librarian.yaml file doesn't exist on other yoshi-node repos, then it's a no-op.
> 
> Can we add it to the existing strategy?

Oh, I see, there is no yoshi-node strategy -- only the default node one.

chingor13 (dismissed):
You'll need to add to the factory so that it can build the strategy you've written. https://github.com/googleapis/release-please/blob/main/src/factory.ts

jskeet:
Yup - no Yoshi one, and you never know whether another Node repo may have a librarian.yaml file of an entirely different nature. Also, once google-cloud-node moves off release-please, it would be good to just revert this change :)

All that said, I have little experience with release-please, so I'm happy to go in whatever direction you suggest.

jskeet:
> You'll need to add to the factory so that it can build the strategy you've written. https://github.com/googleapis/release-please/blob/main/src/factory.ts

Thanks, will do that on Monday.

jskeet:
> You'll need to add to the factory so that it can build the strategy you've written. https://github.com/googleapis/release-please/blob/main/src/factory.ts

Done. I looked at the factory tests, but it didn't *look* like I needed to do anything there.

jskeet:
> > You'll need to add to the factory so that it can build the strategy you've written. https://github.com/googleapis/release-please/blob/main/src/factory.ts
> 
> Done. I looked at the factory tests, but it didn't _look_ like I needed to do anything there.

Turns out I needed to regenerate `__snapshots__/cli.js`; now done so tests are passing.

jskeet:
@chingor13 It turns out the code was completely broken, due to expecting a version map instead of a single version. I've now changed it a lot (all in the last commit) - https://github.com/jskeet/google-cloud-node/pull/9 is a sample release PR.

codyoss (approved):

PR #2790, opened 2026-05-28
fix(commit-parser): fall back to header-only parsing when body is unparseable

### LINKED ISSUE #2564: commit could not be parsed: ... unexpected token '('
Release PR is not updated because of:
```
❯ commit could not be parsed: 5343558be9ed3911861ddcacdd51e4e3d0b581e7 feat: Add support for async queries (#238)
❯ error message: Error: unexpected token '(' at 25:18, valid tokens [)]
```

for [this commit message](https://github.com/tekumara/fakesnow/commit/5343558be9ed3911861ddcacdd51e4e3d0b581e7) on this line:

```
TABLE(RESULT_SCAN('sfqid'))` queries
```

narhen:
Fixes #2564 🦕

## Problem

`parseConventionalCommits` silently drops commits whose body trips the `@conventional-commits/parser` grammar. The grammar requires that any `(` on a body line have its matching `)` on the same line. GitHub's PR-body line-wrap routinely produces commits that violate this — e.g. a function call whose closing paren ends up on the next line:

```
Body wraps a call:
computeValue(longArgumentName
= true)
```

This throws `unexpected token '\n' at LINE:COL, valid tokens [)]` from inside the parser's `scope()` rule. The header itself parses fine; only the body is broken.

The existing `catch` in `parseConventionalCommits` (`src/commit.ts`) logs at debug level and skips the commit, so the entry never reaches the release PR's changelog. #2564 is the canonical bug report for this symptom.

## Root cause

Upstream in `@conventional-commits/parser`, `scope()` throws on a newline before `)` rather than returning a recoverable error. An open PR ([conventional-commits/parser#48](https://github.com/conventional-commits/parser/pull/48)) targets this exact case but has been stalled since March 2024; the parser has not had an npm release since January 2021. We can't realistically wait for an upstream fix.

## Fix

When full-message parsing throws, retry with just the first line of the commit message. If the header parses, push it as a `ConventionalCommit` and log a `warn` so users know the body was discarded. If the header also fails to parse, fall through to the existing debug log and skip the commit (unchanged behavior).

Preserved on the fallback path: `type`, `scope`, `bareMessage`, `breaking` (via `!` in the header), `references` from the header, `pullRequest`, `sha`, `files`.

Lost on the fallback path: body-derived `BREAKING CHANGE` notes, body-derived references, footer trailers. The user-visible improvement is "commit appears in the changelog" rather than "commit silently vanishes."

## Test

Added a `parseConventionalCommits` test with a body that wraps a function call across lines. Without the fix it returns `[]`; with the fix it returns one parsed commit with the correct type/scope/subject.

## Checklist

- [x] Tests and linter pass locally (`npm test`: 1101 passing, 0 failing; `npm run lint`: no new errors)
- [x] Code coverage does not decrease (added test exercises the new branch)
- [x] No doc changes required (behavior change is internal recovery path)
- [x] CLA signed — pending; will follow up

### DIFF
--- a/src/commit.ts
+++ b/src/commit.ts
@@ -421,31 +421,49 @@ export function parseConventionalCommits(
     for (const commitMessage of splitMessages(
       preprocessCommitMessage(commit)
     )) {
+      const pushParsedCommit = (
+        parsedCommit: parser.ConventionalChangelogCommit
+      ) => {
+        const breaking =
+          parsedCommit.notes.filter(note => note.title === 'BREAKING CHANGE')
+            .length > 0;
+        conventionalCommits.push({
+          sha: commit.sha,
+          message: parsedCommit.header,
+          files: commit.files,
+          pullRequest: commit.pullRequest,
+          type: parsedCommit.type,
+          scope: parsedCommit.scope,
+          bareMessage: parsedCommit.subject,
+          notes: parsedCommit.notes,
+          references: parsedCommit.references,
+          breaking,
+        });
+      };
       try {
         for (const parsedCommit of parseCommits(commitMessage)) {
-          const breaking =
-            parsedCommit.notes.filter(note => note.title === 'BREAKING CHANGE')
-              .length > 0;
-          conventionalCommits.push({
-            sha: commit.sha,
-            message: parsedCommit.header,
-            files: commit.files,
-            pullRequest: commit.pullRequest,
-            type: parsedCommit.type,
-            scope: parsedCommit.scope,
-            bareMessage: parsedCommit.subject,
-            notes: parsedCommit.notes,
-            references: parsedCommit.references,
-            breaking,
-          });
+          pushParsedCommit(parsedCommit);
         }
       } catch (_err) {
-        logger.debug(
-          `commit could not be parsed: ${commit.sha} ${
-            commit.message.split('\n')[0]
-          }`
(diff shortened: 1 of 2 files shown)


narhen:
@codyoss / @jskeet / @suztomo, any interest in getting this reviewed and merged?
(Just tagging people based on who I noticed have been active in reviewing and merging PRs lately, sorry for the noise)

svyatov:
I worked through the root cause of this parse failure on #2564, with a minimal reproduction and the exact trigger.

The trigger is narrower than it looks. A body line throws only when its first token is immediately followed by `(`, and the parentheses do not close as a simple scope. `TABLE(RESULT_SCAN('x'))` and `x(` throw. `TABLE(x)` parses, and so does the same text anywhere other than the start of a line.

That may let this fallback keep more than the header. A body whose only defect is one such line could be repaired before the second parse, rather than dropped whole.

Two measurements for the tradeoff you already document in the comment. A `!` in the header survives the header-only path, so `feat!:` still cuts a major. A `BREAKING CHANGE:` note that appears only in the body does not, and that commit cuts a minor with the new warning as the only sign.

#2564 also covers why the upstream fix is unavailable, and lists two other options.

narhen:
Closing this due to lack of interest by maintainers

PR #2799, opened 2026-06-01
feat: add go-librarian release strategy

codyoss:
Introduces the `go-librarian` release strategy and the associated
`LibrarianYamlUpdater` to support automated releases for Go packages
defined in `librarian.yaml`.This new strategy will be used by google-cloud-go in the future. 

This was tested using
https://github.com/codyoss/google-cloud-go/tree/rp-migration-test. I made a commit to secretmanager and then ran these local changes to create the following PR: https://github.com/codyoss/google-cloud-go/pull/8

Internal Bug: b/518756701

### DIFF
--- a/__snapshots__/cli.js
+++ b/__snapshots__/cli.js
@@ -49,11 +49,11 @@ Options:
                                                                         [string]
   --release-type                what type of repo is a release being created
                                 for?
-  [choices: "bazel", "dart", "dotnet-yoshi", "elixir", "expo", "go", "go-yoshi",
-          "helm", "java", "java-backport", "java-bom", "java-lts", "java-yoshi",
-     "java-yoshi-mono-repo", "krm-blueprint", "maven", "node", "node-librarian",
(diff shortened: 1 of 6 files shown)


suztomo:
https://github.com/codyoss/google-cloud-go/pull/6/changes does not touch Librarian files. Shouldn't it update Librarian.yaml file?

codyoss:
@suztomo This was my bad. When I made my last change before opening the PR I regenerated the config files with the strategy "go" and not "go-librarian". After I updated the config in google-cloud-go it worked with no code changes here. Sorry about that! I created an updated PR(7 not 6 now)

suztomo:
That pull request 7 has CHANGELOG.md while the repository uses CHANGES.md.
https://github.com/googleapis/google-cloud-go/blob/main/secretmanager/CHANGES.md

Would you demonstrate this change will not create the discrepancy?

codyoss:
@suztomo Good catch, I forgot Go is the only language that uses a different file. I have updated the code and PR.

suztomo (approved):

PR #2803, opened 2026-06-02
fix(go): support semver dot-separated prerelease suffixes in version-go regex

noahdietz:
If the version value in `internal/verison.go` had a SemVer prerelease suffix with dot separated segments (https://semver.org/#spec-item-9) e.g. `1.2.3-rc.1`, the version updater wouldn't match the string and wouldn't update it with the new version.

Discovered this while testing release-please for previews (https://github.com/googleapis/librarian/issues/6115#issuecomment-4597146301), where we expect to use dot-separated prerelease segments.

### DIFF
--- a/src/updaters/go/version-go.ts
+++ b/src/updaters/go/version-go.ts
@@ -17,7 +17,7 @@ import {DefaultUpdater} from '../default';
 export class VersionGo extends DefaultUpdater {
   updateContent(content: string): string {
     return content.replace(
-      /const Version = "[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z-.]+)?"/,
+      /const Version = "[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?"/,
       `const Version = "${this.version.toString()}"`
     );
   }
--- a/src/updaters/go/version-go.ts
+++ b/src/updaters/go/version-go.ts
@@ -17,7 +17,7 @@ import {DefaultUpdater} from '../default';
 export class VersionGo extends DefaultUpdater {
   updateContent(content: string): string {
     return content.replace(
-      /const Version = "[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z-.]+)??"/,
+      /const Version = "[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z-.]+)?"/,
       `const Version = "${this.version.toString()}"`
     );
   }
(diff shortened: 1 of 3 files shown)


noahdietz:
Apologies @chingor13 didn't realize automation would assign you. Feel free to review as well if you'd like!

src/updaters/go/version-go.ts
```diff
@@ -17,7 +17,7 @@ import {DefaultUpdater} from '../default';
 export class VersionGo extends DefaultUpdater {
   updateContent(content: string): string {
     return content.replace(
-      /const Version = "[0-9]+\.[0-9]+\.[0-9](-\w+)?"/,
+      /const Version = "[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z-.]+)??"/,
```
jskeet: Is the second question mark at the end required?

jskeet (approved):
One nit, but basically fine - assuming I actually understand it...

src/updaters/go/version-go.ts
codyoss: Agreed on the `??`. 

Also, the suffix is would allow values like `0.58.0-rc..1`, which maybe is just fine as we have some level of trust? If that is the case though maybe just make it even more permissive of a grouping for the suffix, `[^"]*`. If we truly want detailed regexs we should likely reference the spec: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string (that is a wild regex)

codyoss (approved):

src/updaters/go/version-go.ts
noahdietz: The extra `?` is not necessary - its the difference between lazy matching (??) and greedy (?). Just removed it.

> Also, the suffix is would allow values like `0.58.0-rc..1`, which maybe is just fine as we have some level of trust?

I'd rather not match that, so I've fixed it to not match with consecutive dots, thanks for catching that.

> If that is the case though maybe just make it even more permissive of a grouping for the suffix, [^"]*...

Kind of want to strike a balance I think - I like the clear readability/intent of "we match prerelease suffixes that are dot separated alphanumeric segments" with a little bit of "we won't match just anything".

> If we truly want detailed regexs we should likely reference the spec...(that is a wild regex)

Yeah I looked at that and was like "....wut", ran it through jetski and it indicated some of the things in there aren't supported by JS regex so we'd need to further tweak it, and honestly I think it's way too hard to grok. So if it's all the same to you (I think it is given you've stamped this) I'm going to stick with a middle ground, semi-explicit, semi-permissive matcher.

PR #2818, opened 2026-06-17
fix(go-librarian): support Go v2 subdirectory releases without branch collisions

### LINKED ISSUE #2817: support Go's v2 subdirectories
Right now release-please has some issue when opening multiple PRs when there are two Go major versions being released with the folder strategy. For example:

```json
"pubsub": {
  "component": "pubsub",
  "exclude-paths": [
    "pubsub/v2"
  ]
},
"pubsub/v2": {
  "component": "pubsub"
},
```

The component for both needs to be pubsub so the tag prefix is correct for Go. But the way release-please opens PRs is including the component in the name of the branch. This means only one of these can be open at a time. We need a way for something like this config to be valid or a way for Go to control its tag for major versions.

Tags should look like pubsub/v1.X.X and pubsub/v2.X.X respectively.

codyoss:
Automatically handles Go v2+ subdirectory modules (e.g. `pubsub/v2`) by stripping the `/v2` suffix for tag names (generating `pubsub/v2.X.X` tags) but preserving a sanitized version of it for branch names (`pubsub-v2`) to avoid Git collisions.

Also updates Manifest lookup mappings to support multiple paths sharing a tag component name, and adds strict validation during release generation to prevent wrong cross-component releases.

Example config and setup:
- Config: https://github.com/codyoss/google-cloud-go/blob/rp-patch/release-please-individual-config.json
- Example PR1(pubsub): https://github.com/codyoss/google-cloud-go/pull/11
- Example PR2(pubsub/v2): https://github.com/codyoss/google-cloud-go/pull/10 

Fixes: #2817

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -323,7 +323,7 @@ export class Manifest {
   private snapshotLabels: string[];
   readonly plugins: ManifestPlugin[];
   private _strategiesByPath?: Record<string, Strategy>;
-  private _pathsByComponent?: Record<string, string>;
+  private _pathsByComponent?: Record<string, string[]>;
   private manifestPath: string;
   private bootstrapSha?: string;
   private lastReleaseSha?: string;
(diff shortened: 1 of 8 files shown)


codyoss:
cc @shollyman after this lands I will revert https://github.com/googleapis/google-cloud-go/pull/20004 and things should just work™️

suztomo:
Can you attach example pull requests in your forked google-cloud-go, created by this change?

suztomo:
Also, would you add a planned change in the Release Please configuration in google-cloud-go, that will work with this pull request in the pull request description? https://github.com/googleapis/release-please/issues/2817#issuecomment-4735533613 ?

codyoss:
@suztomo I have added the example config and the two pull requests that opened from running with these changes.

suztomo:
Can it find the latest release for each component? Checking ...

suztomo:
As long as pull request has a correct pull request merged, it should work.

suztomo (approved):

PR #2830, opened 2026-06-29
fix: avoid duplicate updates in librarian.yaml for Go v2+

### LINKED ISSUE #2826: go-librarian strategy erroneously updating ibrarian.yaml twice

codyoss:
Restricts the component fallback match in LibrarianYamlUpdater to root packages or when packagePath is empty. This prevents v1 libraries from being updated when releasing v2+ versions.

Fixes: #2826

### DIFF
--- a/test/updaters/librarian-yaml.ts
+++ b/test/updaters/librarian-yaml.ts
@@ -45,6 +45,30 @@ libraries:
     version: 1.10.0
 `;
 
+const goContentMultipleVersions = `language: go
+libraries:
+  - name: accessapproval
+    version: 1.13.0
+    apis:
+      - path: google/cloud/accessapproval/v1
+  - name: accessapproval/v2
+    version: 2.0.0
+    apis:
+      - path: google/cloud/accessapproval/v2
+`;
+
+const goContentMultipleVersionsUpdatedV2 = `language: go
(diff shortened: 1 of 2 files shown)


test/updaters/librarian-yaml.ts
```diff
@@ -375,6 +399,16 @@ libraries:
       expect(newContent).to.eq(goContentUpdated);
     });
 
+    it('updates ONLY exact match when multiple versions exist (Go style)', () => {
+      const updater = new LibrarianYamlUpdater({
+        version: Version.parse('2.1.0'),
+        packagePath: 'accessapproval/v2',
+        component: 'accessapproval',
```
suztomo, line 406: Note for myself:

In the case of PubSub in google-cloud-go, the package contains "/v2" and component is "pubsub"

https://github.com/googleapis/google-cloud-go/blob/9968adfbea5b6bba137101ac446132fac8616734/release-please-individual-config.json#L43-L44

```
    "pubsub/v2": {
      "component": "pubsub"
```

suztomo (approved):

PR #2249, opened 2024-03-19
fix: support-node-workspace-plugin-prerelease

Animalmix55:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/release-please/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes #2248 🦕

PR #2836, opened 2026-07-03
fix(cargo-workspace): delegate to versioning strategy for cascade bumps in prerelease

### LINKED ISSUE #2828: cargo-workspace: cascaded dependent bumps ignore the active prerelease window
## What happens

When a cargo workspace is in a prerelease window (a `PrereleaseVersioningStrategy`, e.g. `prerelease: true` with `prereleaseType: "beta"`), crates that are released directly get the correct prerelease bump (`1.0.0-beta.0` → `1.0.0-beta.1`).

But a crate that is only **cascade-bumped** - i.e. it has no release commits of its own and is force-bumped solely because it depends on a crate that *is* being released - gets the wrong version. Instead of incrementing the prerelease counter, it bumps the base patch and keeps the stale prerelease suffix:

```
1.0.0-beta.0  →  1.0.1-beta.0   ❌  (expected: 1.0.0-beta.1)
```

So in a single release PR you end up with two inconsistent paths:

- `core/*` (directly released) → `1.0.0-beta.1` ✅
- `bindings/*` (only a dependent of `core`) → `1.0.1-beta.0` ❌

## Root cause

`CargoWorkspace.bumpVersion` always uses a raw `PatchVersionUpdate`, which is unaware of the configured versioning strategy:

src/plugins/cargo-workspace.ts

```ts
protected bumpVersion(pkg: CrateInfo): Version {
  const version = Version.parse(pkg.version);
  return new PatchVersionUpdate().bump(version);
}
```

`PatchVersionUpdate.bump` increments `patch` and carries the existing `preRelease` string through verbatim, which is exactly the `1.0.0-beta.0` → `1.0.1-beta.0` behavior above.

## This was already fixed for node-workspace

`node-workspace` had the identical bug and it was fixed in **#2249** (`fix: support-node-workspace-plugin-prerelease`, commit `88dc416`) by delegating to the configured strategy instead of hard-coding a patch bump:

```ts
protected bumpVersion(pkg: Package): Version {
  const version = Version.parse(pkg.version);
  const strategy = this.strategiesByPath[pkg.path];

  if (strategy) return strategy.versioningStrategy.bump(version, []);
  return new PatchVersionUpdate().bump(version);
}
```

`cargo-workspace` already stores `strategiesByPath` in `preconfigure()`, just like `node-workspace`, so the same one-line delegation applies directly.

## References

- node-workspace fix: #2249 (`88dc416`)

anishesg:
The `CargoWorkspace.bumpVersion()` method was unconditionally using `PatchVersionUpdate().bump()` for cascade-bumped packages, which increments the patch version and carries prerelease suffixes verbatim. This caused packages bumped only as dependencies to get incorrect versions during prerelease windows (e.g., `1.0.0-beta.0` → `1.0.1-beta.0` instead of `1.0.0-beta.1`).

This fix ports the solution from the `node-workspace` plugin (PR #2249), which delegates to the configured `versioningStrategy` when available. The `strategiesByPath` map is already populated in `preconfigure()` and is now used in `bumpVersion()` to apply the correct prerelease strategy to cascade-bumped packages.

Added a regression test that verifies prerelease cascade bumps increment the prerelease counter rather than the patch version.

Fixes #2828

### DIFF
--- a/src/plugins/cargo-workspace.ts
+++ b/src/plugins/cargo-workspace.ts
@@ -173,6 +173,9 @@ export class CargoWorkspace extends WorkspacePlugin<CrateInfo> {
 
   protected bumpVersion(pkg: CrateInfo): Version {
     const version = Version.parse(pkg.version);
+    const strategy = this.strategiesByPath[pkg.path];
+
+    if (strategy) return strategy.versioningStrategy.bump(version, []);
     return new PatchVersionUpdate().bump(version);
   }
 
--- a/test/plugins/cargo-workspace.ts
+++ b/test/plugins/cargo-workspace.ts
@@ -36,6 +36,9 @@ import {CargoToml} from '../../src/updaters/rust/cargo-toml';
 import {parseCargoManifest} from '../../src/updaters/rust/common';
 import {ConfigurationError} from '../../src/errors';
 import assert = require('assert');
+import {Rust} from '../../src/strategies/rust';
+import {PrereleaseVersioningStrategy} from '../../src/versioning-strategies/prerelease';
+import {TagName} from '../../src/util/tag-name';
 
 const sandbox = sinon.createSandbox();
 const fixturesPath = './test/fixtures/plugins/cargo-workspace';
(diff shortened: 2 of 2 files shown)


anishesg:
closing - no activity in a while, happy to reopen if it's still useful.

PR #2846, opened 2026-07-11
fix(github-api): restore proxy-server support with undici dispatcher

### LINKED ISSUE #2838: proxy-server input is silently ignored (request.agent dropped by @octokit/request v8 native fetch)

Please provide the following details.

#### Environment details

  - OS: Linux (self-hosted CI runner, behind a corporate HTTP proxy)
  - Node.js version: 24.x (also reproduces on 22.x)
  - npm version: 10.x
  - `release-please` version: 17.10.1 (via googleapis/release-please-action@v5), also present on earlier v17 lines

#### Steps to reproduce

  1. Run release-please on a machine whose only egress to `api.github.com` is through an HTTP proxy, and set the `proxy-server` input (action) or `--proxy-server`-equivalent option (library), e.g. `proxy-server: proxy.example.com:912`.
  2. Observe that requests do NOT go through the proxy. On a network where direct egress is blocked, the run fails to reach the API (in our case the org IP allow list rejects the runner's raw egress IP); on a network where direct egress is allowed, requests silently bypass the proxy entirely.
  3. Set `NODE_USE_ENV_PROXY=1` together with a standard `HTTPS_PROXY` env var, and requests now go through the proxy correctly, even though `proxy-server` is unchanged. This is what makes the bug easy to miss: the env var does the work while `proxy-server` appears to be responsible.

#### What's happening (root cause)

`GitHub.create()` builds an `HttpsProxyAgent`/`HttpProxyAgent` from the `proxy` option and passes it as `request.agent` to Octokit and to the GraphQL client:

```ts
// src/github-api.ts
octokit: new Octokit({
  baseUrl: apiUrl,
  auth: options.token,
  request: {
    agent: this.createDefaultAgent(apiUrl, options.proxy),  // <- dropped downstream
    fetch: options.fetch,
  },
}),
// ...and the same request.agent on graphql.defaults({ ... })
```

But the library depends on `@octokit/request` `^8.3.1`. In `@octokit/request` v8, the fetch-wrapper calls native `fetch` and forwards only a fixed set of options:

```js
// @octokit/request@8.4.1 dist-node/index.js (fetch-wrapper)
return fetch(requestOptions.url, {
  method: requestOptions.method,
  body: requestOptions.body,
  headers: requestOptions.headers,
  signal: requestOptions.request?.signal,
  ...(requestOptions.body && { duplex: "half" })
});
```

`request.agent` is not among them, so the proxy agent is silently discarded. This is consistent with the `@octokit/request` v8.0.0 breaking changes, which state: "Replace support for Node.js http(s) Agents with documentation on using fetch dispatchers instead" and "Remove ability to pass custom request options, except method, headers, body, signal, data."

Net effect: the documented `proxy-server` input (README: "Configure a proxy server in the form of `<host>:<port>`"; action.yml: "set proxy server when you run this action behind a proxy") is dead code. It is accepted, produces no error, and has no effect. Worse, if a user also sets `NODE_USE_ENV_PROXY=1` + `HTTPS_PROXY` (Node's own env-proxy support), proxying starts working and `proxy-server` looks like it is doing the job when it is not.

#### Suggested fix

Follow the same migration Octokit documents for v8: instead of passing `request.agent`, pass a custom `fetch` that uses an undici `ProxyAgent` as its `dispatcher`, e.g.

```ts
import { ProxyAgent } from 'undici';

const dispatcher = proxy ? new ProxyAgent(`http://${proxy.host}:${proxy.port}`) : undefined;
const proxyFetch = dispatcher
  ? (url, opts) => fetch(url, { ...opts, dispatcher })
  : undefined;

new Octokit({ baseUrl: apiUrl, auth: token, request: { fetch: proxyFetch ?? options.fetch } });
```

(and the equivalent for the GraphQL client). That restores the `proxy-server` contract on native fetch. Alternatively, if the intent is to rely on Node's env-proxy support, the docs for `proxy-server` should say so and point at `NODE_USE_ENV_PROXY`.

I could not find an existing issue for this specific case (#1654 was a related but different graphql-proxy bug predating the octokit v8 bump; #2259 is about HTTP CONNECT tunneling). Happy to open a PR if a fix along these lines is welcome.

Thank you!

anishesg:
The `proxy-server` input was silently ignored because `@octokit/request` v8 dropped support for `request.agent` in favor of native fetch with dispatchers. The proxy configuration was being passed as `request.agent` to Octokit and GraphQL clients, but this parameter is filtered out before reaching the fetch call.

This change replaces the `createDefaultAgent` method with `createProxyFetch` that creates a custom fetch function using undici's `ProxyAgent` as a dispatcher. The proxy agent is now properly passed through the fetch options, which is compatible with `@octokit/request` v8's architecture. The fix adds undici as a dependency and updates tests to validate the new proxy fetch implementation.

Users who relied on the `NODE_USE_ENV_PROXY` + `HTTPS_PROXY` workaround will continue to work as expected, and the documented `proxy-server` input now functions correctly.

Fixes #2838

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,6 +35,7 @@
         "semver": "^7.5.3",
         "type-fest": "^3.0.0",
         "typescript": "^4.6.4",
+        "undici": "^7.28.0",
         "unist-util-visit": "^2.0.3",
         "unist-util-visit-parents": "^3.1.1",
         "xpath": "^0.0.34",
@@ -5296,6 +5297,15 @@
       },
       "engines": {
         "node": ">=0.8.0"
+      }
+    },
+    "node_modules/undici": {
+      "version": "7.28.0",
+      "resolved": "https://registry.npmjs.org/undici/-/undici-7.28.0.tgz",
+      "integrity": "sha512-cRZYrTDwWznlnRiPjggAGxZXanty6M8RV1ff8Wm4LWXBp7/IG8v5DnOm74DtUBp9OONpK75YlPnIjQqX0dBDtA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=20.18.1"
       }
     },
     "node_modules/undici-types": {
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "semver": "^7.5.3",
     "type-fest": "^3.0.0",
     "typescript": "^4.6.4",
+    "undici": "^7.28.0",
     "unist-util-visit": "^2.0.3",
     "unist-util-visit-parents": "^3.1.1",
     "xpath": "^0.0.34",
--- a/src/github-api.ts
+++ b/src/github-api.ts
@@ -34,8 +34,7 @@ import {
 import {logger as defaultLogger} from './util/logger';
 
 import {graphql} from '@octokit/graphql';
-import {HttpsProxyAgent} from 'https-proxy-agent';
-import {HttpProxyAgent} from 'http-proxy-agent';
+import {ProxyAgent} from 'undici';
 
 export const GH_API_URL = 'https://api.github.com';
 export const GH_GRAPHQL_URL = 'https://api.github.com';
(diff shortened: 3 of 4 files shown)


LunaticMuch:
+1 I have similar issues. For what it matters...I checked and LGTM. Thanks @anishesg <3

anishesg:
Friendly ping — this restores proxy-server support by using undici's dispatcher option. CI is green except the multi-approvers gate (which requires maintainer approval). Would appreciate a maintainer review when you get a chance.

PR #2862, opened 2026-07-28
feat: add PHPLibrarian strategy

JoeWang1127:
Add `PHPLibrarian` strategy extending `PHPYoshi` to automatically update `librarian.yaml` during release PR creation, along with unit tests verifying update generation and content modifications.

For https://github.com/googleapis/librarian/issues/6670

### DIFF
--- a/src/strategies/php-librarian.ts
+++ b/src/strategies/php-librarian.ts
@@ -1,0 +1,38 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


src/strategies/php-librarian.ts
```diff
@@ -0,0 +1,38 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {PHPYoshi} from './php-yoshi';
+import {BuildUpdatesOptions} from './base';
+import {Update} from '../update';
+import {LibrarianYamlUpdater} from '../updaters/librarian-yaml';
+
+export class PHPLibrarian extends PHPYoshi {
```
codyoss, line 20: In order to actually use this new strategy, you need to register it in `factory.ts`. It should be as simple as an import and adding to an existing map, just follow the conventions in the file.

codyoss (changes requested):

src/strategies/php-librarian.ts
JoeWang1127: Done.

codyoss (approved):

PR #2886, opened 2026-08-24
fix: ignore git submodules during tree traversal to prevent 422 errors

codyoss:
## Description
Fixes an issue where GitHub's Git Trees API returns HTTP 422 (`Invalid object requested. SHA must identify a commit or a tree`) when traversing repositories containing Git submodules (mode `160000` / type `commit`).

### Changes
* **Remote Tree Traversal (`src/util/file-cache.ts`)**: Fall back to non-recursive tree traversal when recursive requests return 422, explicitly skipping submodule entries while discovering all valid files.
* **Local Git Traversal (`src/local-github.ts`)**: Filter out mode `160000` gitlink entries during file searches and file reads.
* **Testing**: Added unit tests covering repositories containing Git submodules across both remote and local execution paths.

### DIFF
--- a/src/github.ts
+++ b/src/github.ts
@@ -34,8 +34,8 @@ import {Release} from './release';
 import {ROOT_PROJECT_PATH} from './manifest';
 import {GitHubApi, GitHubCreateOptions} from './github-api';
 import {signoffCommitMessage} from './util/signoff-commit-message';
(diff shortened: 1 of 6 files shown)


codyoss:
I am going to fix this upsteam.

alberto4003 (approved):

PR #2874, opened 2026-08-09
fix(commit): preserve author metadata in parseConventionalCommits

anishesg:
The `parseConventionalCommits()` function in `src/commit.ts` was dropping the `author` field when transforming `Commit` objects into `ConventionalCommit` objects. This caused author metadata to be lost before changelog rendering, even though the `include-commit-authors` config option was enabled and the GitHub API provided author data.

The fix adds `author: commit.author` to the object literal at line 434 where other fields like `sha`, `message`, `files`, and `pullRequest` are already being preserved. This ensures that author information flows through the entire pipeline from commit fetch to changelog generation.

A regression test was added to verify that author metadata (name, email, username) is correctly preserved when parsing conventional commits.

Fixes #2761

PR #2892, opened 2026-08-27
fix(commit): preserve author metadata in parseConventionalCommits

### LINKED ISSUE #2761: include-commit-authors has no effect because author metadata is dropped in parseConventionalCommits()
### What happened

I enabled `include-commit-authors: true` in a manifest-based setup, but generated changelog entries still do not include `(@username)` or author name.

This was reproduced via `release-please-action@v5.0.0` with manifest config, and the release PR was refreshed after the config change. The PR content updated, but changelog entries still had no author suffixes.

### Expected

When `include-commit-authors` is enabled, changelog entries should include:

- `(@username)` when a GitHub username is available
- author name otherwise

### Actual

The changelog renders normally, but author attribution is never appended.

### Reproduction

Minimal shape:

```json
{
  "include-commit-authors": true,
  "packages": {
    ".": {
      "release-type": "elixir",
      "changelog-type": "default"
    }
  }
}
```

Then run release-please on a repo with normal releasable commits. The release PR is generated or updated, but changelog entries do not show author info.

### Suspected cause

I traced the released source and it looks like the config and renderer are wired correctly:

- config parsing reads `include-commit-authors`
- strategy passes `includeCommitAuthors` into changelog note generation
- GitHub commit fetching populates raw `Commit.author`
- `DefaultChangelogNotes` appends author info when `options.includeCommitAuthors && commit.author`

But `parseConventionalCommits()` appears to drop the `author` field when converting `Commit` into `ConventionalCommit`, so by the time changelog notes are rendered, `commit.author` is gone.

The problematic shape seems to be here in `src/commit.ts`:

```ts
conventionalCommits.push({
  sha: commit.sha,
  message: parsedCommit.header,
  files: commit.files,
  pullRequest: commit.pullRequest,
  type: parsedCommit.type,
  scope: parsedCommit.scope,
  bareMessage: parsedCommit.subject,
  notes: parsedCommit.notes,
  references: parsedCommit.references,
  breaking,
});
```

It looks like `author: commit.author` needs to be preserved there.

### Additional note

The existing tests for this feature seem to cover `DefaultChangelogNotes` in isolation with handcrafted commits that already include `author`, but not the full path through `parseConventionalCommits()`. That may be why this slipped through.

If useful, I can open a PR with the minimal fix plus a regression test.

piotrek-sblok:
Recreates #2874, which was closed unmerged (["closing for now to keep the PR list tidy"](https://github.com/googleapis/release-please/pull/2874#issuecomment-3555424672)) with the source branch deleted. Credit for the original fix goes to @anishesg; #2874 has since had people asking for the fix, and #2761 is still open, so I've re-opened it from a live branch and dealt with the test fallout.

### The bug

`parseConventionalCommits()` in `src/commit.ts` drops the `author` field when transforming `Commit` objects into `ConventionalCommit` objects. Everything else on the `include-commit-authors` path is wired correctly — config parsing, the strategy passing `includeCommitAuthors` into changelog note generation, GitHub commit fetching populating `Commit.author`, and `DefaultChangelogNotes` appending author info when `options.includeCommitAuthors && commit.author` — but by the time notes are rendered, `commit.author` is `undefined`, so author attribution never appears in the changelog.

### The fix

`author: commit.author` in the object literal in `parseConventionalCommits()`, alongside the other preserved fields (`sha`, `message`, `files`, `pullRequest`).

### Tests

- `test/commits.ts`: a regression test asserting `name`/`email`/`username` survive `parseConventionalCommits()`. The existing coverage tests `DefaultChangelogNotes` in isolation with handcrafted commits that already carry `author`, so nothing exercised this path — which is why the bug slipped through. I confirmed the new test fails on `main` (`expected undefined not to be undefined`) and passes with the fix.
- `test/manifest.ts`: the `processCommits` plugin-hook `sinon.assert.calledWith` compares against full commit literals, so it needed `author: undefined` added to the two expected commits. This was **not** part of #2874 and is why the full suite would have gone red without it.
- 
Fixes #2761


blafourcade:
We need it too ! 

@piotrek-sblok thank you ☕

